PR #149, opened 2020-03-08
Fix window is undefined

pbxothi:
Moving global window access to onMount hook for to fix "window is undefined" error when using with sapper, etc.

### DIFF
--- a/src/components/UIShell/UIShell.svelte
+++ b/src/components/UIShell/UIShell.svelte
@@ -14,20 +14,24 @@
   import UIShellSideNavWrapper from './UIShellSideNav/UIShellSideNavWrapper.svelte';
   import UIShellSideNavItem from './UIShellSideNav/UIShellSideNavItem.svelte';
   import HamburgerMenu from './UIShellSideNav/HamburgerMenu.svelte';
-
+  import { onMount } from 'svelte' 
+  
   let isSideNavOpen = undefined;
-  let winWidth = window.innerWidth;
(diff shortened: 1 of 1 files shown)


metonym:
@pbxothi Thank you for the PR.

Could you use the [svelte:window API](https://svelte.dev/docs#svelte_window) so that the event listener is also removed when the component is destroyed?

src/components/UIShell/UIShell.svelte
```diff
@@ -14,20 +14,24 @@
...
-      isSideNavOpen = true;
-    } else {
-      isSideNavOpen = false;
-    }
-  });
+  let winWidth = undefined;
+  
+  onMount(() => {
+    winWidth = window.innerWidth
+    window.addEventListener('resize', () => {
```
metonym: Using the `svelte:window` API will automatically remove the event listener when destroyed.

```
<svelte:window on:resize="..." />
```

pbxothi: Thank you, I changed the code to use <svelte:window />; see [revision](https://github.com/IBM/carbon-components-svelte/pull/149/commits/ee27d7b8cd9d12688a173f2c4378bf75bf841a90)

Appreciate the help

metonym:
Thank you @pbxothi – patched in [v0.3.3](https://github.com/IBM/carbon-components-svelte/blob/master/CHANGELOG.md#033---2020-03-09)

PR #168, opened 2020-04-18
Refactor

mfeitoza:
@metonym from official Carbon Docs:

> The UI shell is made up of three components: The header, the left panel, and the right panel. All three can be used independently, but the components were designed to work together.

So I refactored to use Header and SideNav independently.
Improve the StoryBook and organize the exports.
```js
<Header bind:isSideNavOpen>
</Header>
<SideNav bind:isOpen={isSideNavOpen}>
</SideNav>
``

And also remove <footer> from SideNav, because I didn't found nothing in Carbon Docs.


metonym:
Released in [v0.4.1](https://github.com/IBM/carbon-components-svelte/blob/master/CHANGELOG.md#041---2020-04-18)

PR #169, opened 2020-04-20
added on:blur for select

httpsOmkar:
Added event for select when it gets blur


metonym:
Patched in [v0.4.2](https://github.com/IBM/carbon-components-svelte/blob/master/CHANGELOG.md#042---2020-04-20)

PR #185, opened 2020-05-28
Added Event Forwarding To Pagination Component

mabentley85:
Added Event Forwarding to the Pagination component by adding handlers to the Next & Previous Page buttons and the Page Size and Page select fields.

These handlers will dispatch one event containing `page` and `pageSize` details.  This was done to prevent duplicate dispatches when the `pageSize` changes, as the `page` is always set to 1 when `pageSize` is changed.

Question: Is there a better way to handle select value changes than `onBlur`? Using `onChange` results in duplicate triggers since they are binded to the `page` and `pageSize` variables.

### DIFF
--- a/src/components/Pagination/Pagination.svelte
+++ b/src/components/Pagination/Pagination.svelte
@@ -22,6 +22,19 @@
   import CaretRight24 from 'carbon-icons-svelte/lib/CaretRight24';
   import { cx, fillArray } from '../../lib';
   import Select, { SelectItem } from '../Select';
+  import { createEventDispatcher } from 'svelte';
+
+  const dispatch = createEventDispatcher();
+  const handlePaginationChange = (type, value) => {
+      if (type === 'pageSize') {
+        page = 1;
+        pageSize = value;
+      } else {
+        page = value;
+      }
+
+      dispatch('update', {pageSize, page});
+  };  
 
   $: totalPages = Math.max(Math.ceil(totalItems / pageSize), 1);
   $: selectItems = fillArray(totalPages);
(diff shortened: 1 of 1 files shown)


metonym:
Hi @mabentley85 – thank you for the PR. I think it would be useful to dispatch `page` and `pageSize` values.

Using `on:blur` is not the expected behavior because it will not fire immediately like `on:change`. To address the "double firing" from on:change, try using the `afterUpdate` to dispatch the values.

mabentley85:
@metonym Thanks for the hint, I refactored my changes to make use of the `afterUpdate` function.

src/components/Pagination/Pagination.svelte
```diff
@@ -22,6 +22,13 @@
   import CaretRight24 from 'carbon-icons-svelte/lib/CaretRight24';
   import { cx, fillArray } from '../../lib';
   import Select, { SelectItem } from '../Select';
+  import { afterUpdate, createEventDispatcher } from 'svelte';  
+
+  const dispatch = createEventDispatcher();
+
+  afterUpdate(() => {
+    dispatch('update', {pageSize, page});
```
metonym, line 30: After an initial dispatch, the values for `pageSize` and `page` are converted to a string by the Select component. See the console log screenshot below.

Could you convert the value type to a number/integer when dispatching? (i.e. `Number(pageSize)`)

Screen Shot 2020-05-29 at 4 44 23 PM

mabentley85: OK, I changed it to this:
```js
dispatch('update', {pageSize: parseInt(pageSize), page: parseInt(page)});
```
since they should both always be integers

mabentley85:
@metonym Made the requested change

metonym:
Thanks again @mabentley85  for the submission!

Released in v0.7.0 – [changelog notes](https://github.com/IBM/carbon-components-svelte/blob/master/CHANGELOG.md#070---2020-05-29)

Issue #186, opened 2020-06-01
Add header key and sort direction to the TableHeader on:click event in the DataTable component


mabentley85:
I think it's important that the Header Key and Sort direction are included in the `click:header` dispatch event.

I would think that in most cases changing the column that a table is sorted by or the sort direction would necessitate an API request.  Also the ability to disable the current behavior of sorting only the rows currently in the DataTable would need to be added.

Do any plans exist to enable this? If not I can volunteer to add this functionality.

metonym:
This would be a nice enhancement, looking forward to the PR.

mabentley85:
Pull Request #188

metonym:
Released in v0.7.2 – [changelog notes](https://github.com/IBM/carbon-components-svelte/blob/master/CHANGELOG.md#072---2020-06-04)

PR #188, opened 2020-06-04
Added the Sort Direction to the header click dispatch

mabentley85:
Also added `sortCurrentData` to control whether or not the current table data is sorted.

This features will aid in the integrating outside data sources, as you'll most likely want to make a new API call when the sorting column or direction is changed.

Issue #186


mabentley85:
@metonym OK, I removed the `sortCurrentData` boolean.

I will create a new issue for how to address preventing a DataTable from sorting it's current row.

PR #199, opened 2020-07-10
add name to form components

softartisan:
Hi, I add name prop for components that could be used by a form


softartisan:
> IMO, the name prop should not have a default value. We use a default value for the `id` prop only because we need it for setting context.
> 
> Can you initialize `name` as undefined?

Yup, I fix it

PR #201, opened 2020-07-16
Option to set text for HeaderAction component

danielboven:
On the IBM _Enterprise Design Thinking_ website I noticed that the `HeaderAction` component has a text next to the icon, saying **Log in** (or **Log out**). For a printscreen of what I mean, please see: https://prnt.sc/tiulyd
Or the source: https://www.ibm.com/design/thinking/

By adding these labels next to the icons, the UX of an application can be improved. However, I saw that adding text to the `HeaderAction` component was not part of the component (nor for `HeaderActionLink` & `HeaderActionSearch`).
To make this possible I modified the `HeaderAction` & `HeaderActionLink` component to export the `text` variable and add this when it has been defined. To make sure the text does not say `undefined` when users do not use the `text` var (which is intended behaviour of Svelte, see https://github.com/sveltejs/svelte/issues/3905), I added a small function called `optional` which takes care of that.
Also the CSS needs to be modified in order to display the icon and text / label next to each other. I mostly copied the styling for that from the IBM _Enterprise Design Thinking_ website. Because I have no idea where to add the CSS, I added it to a global styling file of my Svelte project. This is the styling I used:

```
.bx--header__action {
    display: flex !important;
    justify-content: center;
    align-items: center;
}

.bx--header__action--text {
    font-size: 16px;
    line-height: 20px;
    text-decoration: none;
    margin-bottom: 16px;
    min-width: 120px;
    color: #fff;
    -webkit-font-smoothing: antialiased;
    -moz-osx-font-smoothing: grayscale;
    text-rendering: optimizeLegibility;
    text-shadow: 1px 1px 1px rgba(0, 0, 0, 0.004);
}

.bx--header__action--text > svg {
    margin-right: 1rem;
}
```

Perhaps this could be modified or added in a better way, I'm not sure what the usual approach is to do that. Also, I removed the `action-link` class and the `action-link` styling inside the `HeaderActionLink` component. This class becomes unnecessary when you center the icon of the `HeaderActionLink` component with `display: flex` (please see the CSS above for `.bx--header__action` for what I mean).

Note that I have not changed the `HeaderActionSearch` component. If my solution is good, same changes could be made to that component (probably).

### DIFF
--- a/src/components/UIShell/GlobalHeader/HeaderAction.svelte
+++ b/src/components/UIShell/GlobalHeader/HeaderAction.svelte
@@ -2,10 +2,11 @@
   export let type = undefined;
   export let icon = undefined;
   export let isOpen = undefined;
+  export let text = undefined;
 
   let elRigthPanel = undefined;
 
-  import { cx } from '../../../lib';
+  import { cx, optional } from '../../../lib';
   import Icon from '../../Icon/Icon.svelte';
   import { slide } from 'svelte/transition';
 
@@ -23,12 +24,13 @@
 <div>
   <button
     aria-label={type}
-    class={cx('--header__action', isOpen && '--header__action--active')}
+    class={cx('--header__action', isOpen && '--header__action--active', text && '--header__action--text')}
     type="button"
     on:click={() => {
       isOpen = true;
     }}>
     <Icon {...icon} />
+    {optional(text)}
   </button>
   {#if isOpen}
     <div
--- a/src/components/UIShell/GlobalHeader/HeaderAction.svelte
+++ b/src/components/UIShell/GlobalHeader/HeaderAction.svelte
@@ -6,7 +6,7 @@
 
   let elRigthPanel = undefined;
 
-  import { cx, optional } from '../../../lib';
+  import { cx } from '../../../lib';
   import Icon from '../../Icon/Icon.svelte';
   import { slide } from 'svelte/transition';
 
@@ -47,7 +47,7 @@
       isOpen = true;
     }}>
     <Icon {...icon} />
-    <slot name="custom-el">
+    <slot name="text">
       {#if text}
         <span>{text}</span>
       {/if}
--- a/src/components/UIShell/GlobalHeader/HeaderAction.svelte
+++ b/src/components/UIShell/GlobalHeader/HeaderAction.svelte
@@ -19,18 +19,39 @@
   }
 </script>
 
+<style>
+    .action-text {
+        font-size: 16px;
+        line-height: 20px;
+        text-decoration: none;
+        color: #fff;
+        width: 100%;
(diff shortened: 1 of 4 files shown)


danielboven:
Thank you for your suggestions! Using a `slot` is a great idea.
I can also understand that custom CSS is difficult to maintain, considering the use of a flagship Carbon CSS file. In order to display the text correctly for the `HeaderAction` component custom CSS will still be required, though, since without extra styling it looks like this:

loginNoCSS

A solution scoped to only a `HeaderAction` component which uses the exported `text` variable could look like this:

```
<script>
    export let text = undefined;
</script>

<style>
    .action-text {
        font-size: 16px;
        line-height: 20px;
        text-decoration: none;
        color: #fff;
        width: 100%;
        padding: 0 1rem;
    }

    .action-text > span {
        margin-left: 0.75rem;
        vertical-align: top;
    }
</style>

<button
    .........
    class:action-text={text}
    ......... >

    <slot name="custom-el">
      {#if text}
        <span>{text}</span>
      {/if}
    </slot>
</button>
```

I have also added the code as seen above to my latest commit :+1:

src/components/UIShell/GlobalHeader/HeaderAction.svelte
```diff
@@ -2,10 +2,11 @@
   export let type = undefined;
   export let icon = undefined;
   export let isOpen = undefined;
+  export let text = undefined;
 
   let elRigthPanel = undefined;
 
-  import { cx } from '../../../lib';
+  import { cx, optional } from '../../../lib';
```
metonym, line 9: Remove the "optional" import

src/components/UIShell/GlobalHeader/HeaderAction.svelte
```diff
@@ -18,17 +19,39 @@
...
   <button
     aria-label={type}
     class={cx('--header__action', isOpen && '--header__action--active')}
+    class:action-text={text}
     type="button"
     on:click={() => {
       isOpen = true;
     }}>
     <Icon {...icon} />
+    <slot name="custom-el">
```
metonym, line 50: Let's use a different name for the slot. Maybe "text"?

src/components/UIShell/GlobalHeader/HeaderAction.svelte
danielboven: I must have forgotten to remove that. Done!

src/components/UIShell/GlobalHeader/HeaderAction.svelte
danielboven: Sure, I changed it to "text" :+1:

danielboven:
Thanks for your comments, the changes can be seen in my latest commit.

PR #237, opened 2020-09-02
Fix missing required attribute on TextInput.

### LINKED ISSUE #236: TextInput - Required attribute is not working in Form component
## Description
Required attribute is not working in Form component.

## Expected Outcome
When value of the field is empty and I submit form it should show an error that the field is empty and shouldnt allow to submit my form.

## Actual Outcome
When I submit the form with an empty value it does not give any error and it submits the form.

## Code Snippet

<Form>
  <div class:child-div={childDiv}>
    <TextInput
      placeholder="Please Enter your name"
      labelText="Donor Name"
      id="name"
      name="name"
      required="true"
      bind:value={donor} />
  </div>  
</Form>

Im happy to submit a PR.

kamil-murtaza:
Fixes #236 .

### DIFF
--- a/src/TextInput/TextInput.svelte
+++ b/src/TextInput/TextInput.svelte
@@ -83,6 +83,12 @@
    */
   export let ref = null;
 
+  /**
+   * Set to `true` to mark the field as required
+   * @type {boolean} [required="false"]
+   */
+   export let required = false;
+
   import WarningFilled16 from "carbon-icons-svelte/lib/WarningFilled16";
 
   $: errorId = `error-${id}`;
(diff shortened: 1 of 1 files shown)


src/TextInput/TextInput.svelte
```diff
@@ -83,6 +83,12 @@
    */
   export let ref = null;
 
+  /**
+   * Set to `true` to mark the field as required
+   * @type {boolean} [required="false"]
```
metonym, line 88: `"false"` should be `false`

metonym (changes requested):
Minor typo

src/TextInput/TextInput.svelte
kamil-murtaza: Okay I will complete this change. 

Thank you

kamil-murtaza:
> Minor typo

The issue is fixed.

PR #261, opened 2020-09-18
fix(TimePickerSelect): deprecate hideLabel to match spec

### LINKED ISSUE #260: [10.11]: fix(TimePickerSelect): deprecate hideLabel to match spec
https://github.com/carbon-design-system/carbon/commit/229c2364f

josefaidt:
fixes #260 

## Description

Adds `@deprecated` JSDoc syntax with message from referenced commit. Also removes `hideLabel` option from TimePickerSelect story


metonym (approved):
Thanks for keeping this up to spec.

PR #271, opened 2020-09-19
fix(data-table): change DataTable headers prop type

### LINKED ISSUE #270: fix: DataTable header prop type
`DataTable` component’s header prop has a type defined as follows:
https://github.com/IBM/carbon-components-svelte/blob/be7bbeda033674dfc68fe07ffacd3d8da3c9df99/types/index.d.ts#L724
Where it should be defined as an array of that object type:
```ts
headers?: { key: string; value: string }[];
```

albertms10:
Fixes #270

---

- fix(data-table): change DataTable headers prop type

### DIFF
--- a/src/DataTable/DataTable.svelte
+++ b/src/DataTable/DataTable.svelte
@@ -1,7 +1,7 @@
 <script>
   /**
    * Specify the data table headers
-   * @type {{key: string; value: string;}} [headers=[]]
+   * @type {{key: string; value: string;}[]} [headers=[]]
    */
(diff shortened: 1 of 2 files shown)


metonym (changes requested):
Nice catch.

The TypeScript definitions are generated from the JSDocs in the Svelte components.

Could you apply this fix to `DataTable.svelte`?

albertms10:
Sure!

metonym (approved):
👍

PR #277, opened 2020-09-23
feat(text-input): add inline option for TextInput

### LINKED ISSUE #276: [10.16]: feat(text-input): add inline option for TextInput
feat(text-input): add inline option for TextInput

josefaidt:
fixes #276 

## Description

- adds `inline` prop and appropriate classes following [Carbon React text-input](https://github.com/carbon-design-system/carbon/blob/a5e27b74c9e4659759806db0facf7455987d5f48/packages/react/src/components/TextInput/TextInput.js)
- adds necessary input outer wrapper

@metonym I noticed there are inline variants of the React password input fields but I'm not sure if it's functional there. Should we aim to include the inline variants on these other text inputs as well? Are the input fields similar enough to extract out to the `.Skeleton.svelte` file?
image

### DIFF
--- a/src/TextInput/TextInput.stories.js
+++ b/src/TextInput/TextInput.stories.js
@@ -29,6 +29,7 @@ export const Default = () => ({
       'This will overwrite your current settings'
     ),
     placeholder: text("Placeholder text (placeholder)", "Placeholder text."),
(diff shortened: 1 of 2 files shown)


metonym (approved):
Wonderful work

PR #279, opened 2020-09-24
feat(datatable-skeleton): change to size prop

### LINKED ISSUE #278: Use size prop in DataTableSkeleton as in DataTable
`DataTableSkeleton` should use the same size logic as its relative `DataTable` component, just using the proper checks in class names.

https://github.com/IBM/carbon-components-svelte/blob/13ca3bb4c64e9495e4d3ea6b9ee20eb3273dec5b/src/DataTable/DataTable.svelte#L15-L19

https://github.com/IBM/carbon-components-svelte/blob/13ca3bb4c64e9495e4d3ea6b9ee20eb3273dec5b/src/DataTableSkeleton/DataTableSkeleton.svelte#L14-L18

albertms10:
fixes #278

---

- feat(datatable-skeleton): change to size prop

### DIFF
--- a/src/DataTableSkeleton/DataTable.stories.js
+++ b/src/DataTableSkeleton/DataTable.stories.js
@@ -1,4 +1,10 @@
-import { withKnobs, array, boolean, number } from "@storybook/addon-knobs";
+import {
+  withKnobs,
+  array,
+  boolean,
+  number,
+  select,
+} from "@storybook/addon-knobs";
 import Component from "./DataTableSkeleton.Story.svelte";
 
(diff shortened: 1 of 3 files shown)


metonym (changes requested):
Great work.

In the [DataTableSkeleton story](https://github.com/IBM/carbon-components-svelte/blob/master/src/DataTableSkeleton/DataTable.stories.js):

- document the `size` prop
- remove the `compact` prop

metonym (approved):
Nice work – this feature makes the sizing consistent with that of the datatable

PR #291, opened 2020-09-29
feat(pagination): add pageSizeInputDisabled

weaseldotro:
useful for programs that have fixed pagination size, which cannot / must not be changed by the user


weaseldotro:
Sorry, I accidentally the wrong component name, navigation instead of pagination.

PR #292, opened 2020-09-29
feat(pagination): add pageSizeInputDisabled

weaseldotro:
useful for programs that have fixed pagination size, which cannot / must not be changed by the user

### DIFF
--- a/src/Pagination/Pagination.stories.js
+++ b/src/Pagination/Pagination.stories.js
@@ -20,6 +20,7 @@ export const Default = () => ({
       false
     ),
     pageInputDisabled: boolean("Disable page input (pageInputDisabled)", false),
+    pageSizeInputDisabled: boolean("Disable page size input (pageSizeInputDisabled)", false),
     backwardText: text(
       "The description for the backward icon (backwardText)",
       "Previous page"
@@ -49,6 +50,7 @@ export const Multiple = () => ({
       false
     ),
     pageInputDisabled: boolean("Disable page input (pageInputDisabled)", false),
+    pageSizeInputDisabled: boolean("Disable page size input (pageSizeInputDisabled)", false),
     backwardText: text(
       "The description for the backward icon (backwardText)",
       "Previous page"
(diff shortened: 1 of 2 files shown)


metonym (changes requested):
Makes sense to me – it could be useful to disable/hide the select options for both items per page and current page.

Could you document the `pageSizeInputDisabled` in the Story as well?

weaseldotro:
After thinking about this a little bit more, I think it makes more sense that the page size input and label should be opt-in, so disabled by default and enabled using the prop, should the user need it. This is because from a programming point of view, when building an API or UI, the standard is that the page size is a fixed number, and the user-changeable page size is usually an additional feature.

So `pageSizeInputDisabled` should be changed to something like `pageSizeInputEnabled`, which defaults to false and thus the input and label are hidden.

Yes, I know this is contrary to the existing naming of the `pageInputDisabled` prop, but I still think it's a better way to deal with this.

What do you think?

metonym:
@weaseldotro I appreciate the proposal but I disagree that inputs should be opt-in. By design, the [Carbon Pagination component](https://www.carbondesignsystem.com/components/pagination/usage) is meant to afford navigation between pages. I don't think that only the back/forward buttons should be the default navigation.

weaseldotro:
The page number select box should definitely be enabled by default and disabled by request. Obviously the user must be able to select a specific page. Moreover, a first page & last page navigation buttons should also be added.

However, the page size is a different issue, because of the reasons I stated above.

metonym:
> Moreover, a first page & last page navigation buttons should also be added.

This is a great suggestion which also applies to the `PaginationNav` component. I'd recommend raising this as an issue in the [flagship Carbon monorepo](https://github.com/carbon-design-system/carbon/issues). I think you'll get a more decisive answer from the Carbon team designers.

metonym (approved):

metonym:
We treat the API of `carbon-components-react` as a canonical reference because it is the most developed/consumed/maintained implementation. That library (and other implementations like for Vue/Angular) use the CSS from `carbon-components`. As a result, we must align as best as we can with that API, otherwise components will appear broken (e.g. helper text, fluid forms, tabs).

Based on that, certain component behaviors must be "opt in" by default.

PR #301, opened 2020-10-04
feat(link): add href prop

### LINKED ISSUE #300: Missing Link href prop
TypeScript does not recognize the `href` prop from the `Link` component.

```
Type '{ href: string; }' is not assignable to type 'IntrinsicAttributes & { inline?: boolean; disabled?: boolean; visited?: boolean; ref?: HTMLParagraphElement | HTMLAnchorElement; }'.
  Property 'href' does not exist on type 'IntrinsicAttributes & { inline?: boolean; disabled?: boolean; visited?: boolean; ref?: HTMLParagraphElement | HTMLAnchorElement; }'. Did you mean 'ref'? ts(2322)
```

albertms10:
fixes #300

---

- feat(link): add href prop

### DIFF
--- a/src/Link/Link.svelte
+++ b/src/Link/Link.svelte
@@ -1,5 +1,11 @@
 <script>
   /**
+   * Specify href value
+   * @type {string} [href]
+   */
+  export let href;
+
+  /**
    * Set to `true` to use the inline variant
    * @type {boolean} [inline=false]
    */
@@ -48,6 +54,7 @@
     class:bx--link--inline="{inline}"
     class:bx--link--visited="{visited}"
     rel="{$$restProps.target === '_blank' ? 'noopener noreferrer' : undefined}"
+    {href}
     {...$$restProps}
     on:click
     on:mouseover
--- a/src/Link/Link.svelte
+++ b/src/Link/Link.svelte
@@ -1,9 +1,9 @@
 <script>
   /**
-   * Specify href value
+   * Specify the href value
    * @type {string} [href]
    */
-  export let href;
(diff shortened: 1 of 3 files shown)


src/Link/Link.svelte
```diff
@@ -1,4 +1,10 @@
 <script>
+  /**
+   * Specify href value
+   * @type {string} [href]
+   */
+  export let href;
```
metonym, line 6: I know it's the same, but to stay consistent initialize `href` as `undefined`

src/Link/Link.svelte
```diff
@@ -1,4 +1,10 @@
 <script>
+  /**
+   * Specify href value
```
metonym, line 3: Add a "the" after "Specify"

types/index.d.ts
```diff
@@ -1884,6 +1884,11 @@ export class InlineNotification extends CarbonSvelteComponent {
 
 export class Link extends CarbonSvelteComponent {
   $$prop_def: {
+    /**
```
metonym, line 1887: Did you manually add this? It should be generated as part of the build process from running `yarn prepack`.

metonym (changes requested):
See comments

src/Link/Link.svelte
albertms10: Well, I (badly?) supposed that not initializing a prop made it kind of mandatory. Nevertheless, I will change this!

metonym (approved):
Thanks for fixing this

PR #303, opened 2020-10-05
add ref to ListBoxMenu

weaseldotro:


src/ListBox/ListBoxMenu.svelte
```diff
@@ -4,9 +4,16 @@
    * @type {string} [id]
    */
   export let id = "ccs-" + Math.random().toString(36);
+
+  /**
+   * Obtain a reference to the input HTML element
+   * @type {null | HTMLInputElement} [ref=null]
```
metonym, line 10: This should be typed as an `HTMLDivElement`

src/RadioButtonGroup/RadioButtonGroup.svelte
```diff
@@ -23,6 +23,12 @@
    */
   export let orientation = "horizontal";
 
+  /**
+   * Set an id for the container div element
+   * @type {string} [id]
+   */
+   export let id = "ccs-" + Math.random().toString(36);
```
metonym, line 30: The RadioButtonGroup does not need a default `id`

metonym (changes requested):

src/RadioButtonGroup/RadioButtonGroup.svelte
```diff
@@ -23,6 +23,12 @@
    */
   export let orientation = "horizontal";
 
+  /**
+   * Set an id for the container div element
+   * @type {string} [id]
+   */
+   export let id
```
metonym, line 30: I should have been more clear.

The `id` prop isn't necessary for this component, because it can be spread through `$$restProps` anyways.

metonym (changes requested):

src/RadioButtonGroup/RadioButtonGroup.svelte
weaseldotro: I agree, but I'm using typescript and I get an error for props that are not exported, including this one:
```
(JSX attribute) id: string
Type '{ id: string; selected: string; }' is not assignable to type 'IntrinsicAttributes & { selected?: string; disabled?: boolean; labelPosition?: "right" | "left"; orientation?: "horizontal" | "vertical"; }'.
  Property 'id' does not exist on type 'IntrinsicAttributes & { selected?: string; disabled?: boolean; labelPosition?: "right" | "left"; orientation?: "horizontal" | "vertical"; }'.ts(2322)
Peek Problem (Alt+F8)
No quick fixes available
```
So that's why I added this prop, so I don't get an error. However, even with the error, the `id` prop is correctly forwarded to the div and shows up in the rendered HTML.

So what do you think? Keep the code as is because it already works, even though there's this stupid error, or add this (somewhat useless) prop?

metonym: I see. This error has come up multiple times. Let's include until the type definitions are fixed to include intrinsic element attributes.

metonym (approved):

PR #316, opened 2020-10-08
remove locales from DatePicker / flatpickr

### LINKED ISSUE #313: Remove locales from flatpickr (DatePicker component)
Locales for the DatePicker components shouldn't be preinstalled, developers can add them depending on their app requirements.
Removing locales from DatePicker saves 41972 bytes from the total **build** bundle size.

It's the second line in this file: https://github.com/IBM/carbon-components-svelte/blob/master/src/DatePicker/createCalendar.js
```
import l10n from "flatpickr/dist/l10n/index.js";
```

weaseldotro:
Fixes #313

### DIFF
--- a/src/DatePicker/createCalendar.js
+++ b/src/DatePicker/createCalendar.js
@@ -1,6 +1,8 @@
 import flatpickr from "flatpickr";
 import rangePlugin from "flatpickr/dist/plugins/rangePlugin";
 
+var l10n
+
 function updateClasses(instance) {
   const {
     calendarContainer,
--- a/src/DatePicker/createCalendar.js
+++ b/src/DatePicker/createCalendar.js
@@ -1,7 +1,7 @@
 import flatpickr from "flatpickr";
 import rangePlugin from "flatpickr/dist/plugins/rangePlugin";
 
-var l10n
+let l10n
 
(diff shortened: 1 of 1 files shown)


metonym (changes requested):
In the [deploy preview](https://carbon-components-svelte-git-patch-1.carbon-svelte.vercel.app/components/DatePicker), I'm unable to open the calendar for the "single" type DatePicker.

There are still references to `l10n` in the file.

Error in the DevTools console:

Screen Shot 2020-10-08 at 9 19 27 AM

weaseldotro:
Fixed, I added the missing variable.

src/DatePicker/createCalendar.js
```diff
@@ -1,7 +1,8 @@
 import flatpickr from "flatpickr";
-import l10n from "flatpickr/dist/l10n/index.js";
 import rangePlugin from "flatpickr/dist/plugins/rangePlugin";
 
+var l10n
```
albertms10, line 4: Maybe we could declare it following ES2015:

```js
let l10n
```

weaseldotro: Correct, sorry about that. It is done.

metonym (approved):

PR #318, opened 2020-10-08
feat(Tabs): add light variant

### LINKED ISSUE #317: Tabs light variant
https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Tabs/Tabs.js#L40

josefaidt:
fixes #317


metonym:
Unable to see the light variant in the Storybook. I think the upstream CSS behavior has changed.

For 10.21, it seems that `light` only applies to the scrollable Tabs variant: https://github.com/carbon-design-system/carbon/blame/master/packages/react/src/components/Tabs/Tabs.js#L401

PR #321, opened 2020-10-08
chore(links): add rel="" to all links

### LINKED ISSUE #320: add rel="" to all links (ClickableTile, SideLink etc.)
Not all components with the `href` prop would qualify:

- SkipToContent (href expected to be an anchor)
- BreadcrumbItem (assume that `href` should not direct to a new domain)

josefaidt:
fixes #320 

## Description

Adds ```rel="{$$restProps.target === '_blank' ? 'noopener noreferrer' : undefined}"``` to the following components:

- Button.Skeleton
- ClickableTile
- SideNavLink
- HeaderActionLink
- HeaderNavItem
- HeaderPanelLink

### DIFF
--- a/src/Tile/ClickableTile.svelte
+++ b/src/Tile/ClickableTile.svelte
@@ -26,6 +26,7 @@
   class:bx--tile--light="{light}"
   {...$$restProps}
   href="{href}"
+  rel=""
   on:click
   on:click="{() => {
     clicked = !clicked;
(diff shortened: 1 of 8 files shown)


src/Tile/ClickableTile.svelte
```diff
@@ -26,6 +26,7 @@
   class:bx--tile--light="{light}"
   {...$$restProps}
   href="{href}"
+  rel=""
```
metonym, line 29: This should be updated and moved before restProps

metonym (changes requested):

metonym (approved):

PR #331, opened 2020-10-14
Retire storybook

metonym:
**Documentation**

- remove all traces of Storybook (dependencies, set-up, stories)

**Housekeeping**

- transfer `.prettierrc` config to `package.json`
- split commands to build pre-compiled CSS and rollup into `build:css` and `build:lib`
- add `npm-run-all` to run `build:css` and `build:lib` in parallel


AGoblinKing:
Hi, what was the motive behind removing storybook if I may ask?

metonym:
@AGoblinKing Off the top of my mind:

- Documenting Svelte components was arduous. Each component had to be wrapped in a ".Story.svelte" and there was no example code that could be copied. With our new set-up, sample code is easily generated alongside each example.

- We don't want to use React for the documentation website.

PR #348, opened 2020-10-20
fix(ComposedModal): remove bx--body--with-modal-open from body classes after destroy

weaseldotro:
When onMount is async, the returned function is not executed: https://github.com/sveltejs/svelte/issues/4927
Because the class is not remove from the body, the code should be moved to a onDestroy function.


metonym (approved):
thank you for fixing this

PR #352, opened 2020-10-21
feat: add DataTable display and sort methods

### LINKED ISSUE #282: Add custom DataTable sorting method
The current DataTable sorting just works fine. It would also be great to be able to set a custom sorting method to manage situations where the displayed data is treated differently than the ‘raw’ data, e.g. currency, should be sorted as a number (`3217.58`) and displayed as a formatted string (`$3,217.58`).

albertms10:
fixes #282

---

- feat: add display and sort methods in header object
- feat: update types and Component Index
- feat: update Public API with datatable headers methods

### DIFF
--- a/src/DataTable/DataTable.svelte
+++ b/src/DataTable/DataTable.svelte
@@ -1,7 +1,7 @@
 <script>
   /**
    * Specify the data table headers
-   * @type {{key: string; value: string;}[]} [headers=[]]
+   * @type {{key: string; value: string; display?: (item) => string; sort?: (a, b) => number}[]} [headers=[]]
    */
   export let headers = [];
 
@@ -85,7 +85,12 @@
   };
   const dispatch = createEventDispatcher();
   const tableSortable = writable(sortable);
-  const sortHeader = writable({ id: null, key: null, sortDirection: "none" });
+  const sortHeader = writable({
+    id: null,
+    key: null,
+    sort: undefined,
+    sortDirection: "none",
+  });
   const headerItems = writable([]);
   const thKeys = derived(headerItems, () =>
     headers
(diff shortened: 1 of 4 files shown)


src/DataTable/DataTable.svelte
```diff
@@ -231,7 +241,9 @@
                 dispatch('click:cell', cell);
               }}"
             >
-              <slot name="cell" row="{row}" cell="{cell}">{cell.value}</slot>
+              <slot name="cell" row="{row}" cell="{cell}">
+                {headers[j].display ? headers[j].display(cell.value) : cell.value}
```
albertms10, line 245: Should the `display` method be accessed this way—referencing the original `headers` array—or by passing it to the cells object?

https://github.com/IBM/carbon-components-svelte/blob/ead9634ab641ce50b13046604687164ee86be8cd/src/DataTable/DataTable.svelte#L119-L122

src/DataTable/DataTable.svelte
```diff
@@ -85,7 +85,12 @@
   };
   const dispatch = createEventDispatcher();
   const tableSortable = writable(sortable);
-  const sortHeader = writable({ id: null, key: null, sortDirection: "none" });
+  const sortHeader = writable({
+    id: null,
+    key: null,
+    sort: undefined,
```
albertms10, line 91: Should `sort` be initialized with `null` as the other `sortHeader` properties?

src/DataTable/DataTable.svelte
metonym: What you have is good

metonym (approved):
Nice work! This feature makes the `DataTable` very flexible.

Now, you'd be able to:

- customize display/sorting per header as needed
- customize cell value with markup/components using named slots

Can't wait to start using this.

PR #356, opened 2020-10-23
fix(datatable): fix sort parameters

### LINKED ISSUE #355: Fix DataTable sort method direction and key values
While the default DataTable `sort` method manages sort direction and header key values properly, the custom method is defined without these variables in scope.

https://github.com/IBM/carbon-components-svelte/blob/8a65e463a16a7dcc48166471d6dab6362cfe2b3e/src/DataTable/DataTable.svelte#L135-L136

We may directly pass the `a` and `b` params as a result of `itemA` and `itemB` values, respectively.

albertms10:
fixes #355

---

- fix(datatable): fix sort parameters

### DIFF
--- a/src/DataTable/DataTable.svelte
+++ b/src/DataTable/DataTable.svelte
@@ -128,22 +128,19 @@
     if ($sortHeader.sortDirection === "none") {
       sortedRows = rows;
     } else {
-      sortedRows = [...rows].sort(
-        $sortHeader.sort
-          ? $sortHeader.sort
-          : (a, b) => {
-              const itemA = ascending ? a[sortKey] : b[sortKey];
-              const itemB = ascending ? b[sortKey] : a[sortKey];
+      sortedRows = [...rows].sort((a, b) => {
+        const itemA = ascending ? a[sortKey] : b[sortKey];
+        const itemB = ascending ? b[sortKey] : a[sortKey];
 
-              if (typeof itemA === "number" && typeof itemB === "number") {
-                return itemA - itemB;
-              }
+        if ($sortHeader.sort) return $sortHeader.sort(itemA, itemB);
 
-              return itemA
-                .toString()
-                .localeCompare(itemB.toString(), "en", { numeric: true });
-            }
-      );
+        if (typeof itemA === "number" && typeof itemB === "number")
+          return itemA - itemB;
+
+        return itemA
+          .toString()
+          .localeCompare(itemB.toString(), "en", { numeric: true });
+      });
     }
   }
 </script>


metonym:
@albertms10  Could you add an example that adapts the existing ["Sortable" example](https://carbon-svelte.vercel.app/components/DataTable#sortable) that demonstrates the custom sort method?

albertms10:
Something like using `dayjs` and `i18n` related dependencies:

```svelte
<script lang="ts">
  import dayjs from "dayjs";
  import relativeTime from "dayjs/plugin/relativeTime";
  import { number } from "svelte-i18n";

  dayjs.extend(relativeTime); // for `dayjs().fromNow()` method

  const headers = [
    {
      key: "name",
      value: "Name",
    },
    {
      key: "quantity",
      value: "Quantity"
      display: (quantity) => $number(quantity, { format: "EUR" }),
    },
    {
      key: "date",
      value: "Date",
      display: (date) => dayjs(date).fromNow(),
      sort: (a, b) => (dayjs(a.date).isAfter(dayjs(b.date)) ? 1 : -1),
    },
  ];

  const rows = [
    {
      id: 1,
      name: "John",
      quantity: 157.89,
      date: "2020-10-21",
    }
  ];
</script>

<DataTable {headers} {rows} />
```

metonym:
To avoid adding additional dependencies, can you simplify the example even further to omit dayjs, svelte-i18n?

albertms10:
Ok, what about:

```svelte
<script lang="ts">

  const headers = [
    {
      key: "name",
      value: "Name",
    },
    {
      key: "quantity",
      value: "Quantity",
      display: (quantity: number) => `${quantity} €`,
    },
    {
      key: "date",
      value: "Date",
      display: (date: string) => new Date(date).toLocaleString(),
      sort: (a: string, b: string) => (new Date(a) <= new Date(b) ? -1 : 1),
    },
  ];

  const rows = [
    {
      id: 1,
      name: "John",
      quantity: 157.89,
      date: "2020-10-21",
    }
  ];
</script>

<DataTable {headers} {rows} />
```

metonym:
That works

albertms10:
What do you think, @metonym, about the types that involve `rows` and `headers`? https://github.com/IBM/carbon-components-svelte/pull/356#issuecomment-715884784

metonym:
@albertms10 For the TypeScript comment, I'm not sure how it would be automatically inferred without typing it. Would typing `headers` as a tuple work?

Something like:

```ts
const headers: [{key: "quantity", value: "Quantity", display: (quantity: number) => string;}] = [{
  key: "quantity",
  value: "Quantity",
  display: (quantity: number) => `${quantity} €`,
}]
```

metonym (approved):

albertms10:
To get things properly organized, I opened a new issue (#364) to talk about TypeScript types definitions and let this issue close the way it should.

PR #366, opened 2020-10-25
Forward scroll event for ComboBox and add a reference to the list HTML element

weaseldotro:
This prepares the ComboBox for a "infinite" loading scenario: we watch the scroll position of the list, so when the user scrolls and reaches the end of the list then we can append some new items to the list.

### DIFF
--- a/src/ComboBox/ComboBox.svelte
+++ b/src/ComboBox/ComboBox.svelte
@@ -106,6 +106,12 @@
    * @type {null | HTMLInputElement} [ref=null]
    */
   export let ref = null;
+  
+	/**
+	 * Obtain a reference to the list HTML element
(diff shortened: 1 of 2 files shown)


src/ComboBox/ComboBox.svelte
```diff
@@ -106,6 +106,12 @@
    * @type {null | HTMLInputElement} [ref=null]
    */
   export let ref = null;
+  
+	/**
+	 * Obtain a reference to the list HTML element
+	 * @type {null | HTMLElement} [ref=null]
```
metonym, line 112:
```suggestion
	 * @type {null | HTMLDivElement} [ref=null]
```

metonym (approved):

Issue #364, opened 2020-10-24
DataTable: headers and rows types


albertms10:
It would be great if the `display` and `sort` method’s parameter types were automatically inferred, somehow.
It seems tricky, as having both rows and headers types related would create a circular dependency.

```ts
// Typing `headers` with `rows` objects data
const headers = [
  ...
  {
    key: "quantity",
    value: "Quantity",
    display: (quantity: rows[number]["quantity"]) => quantity + " €",
  },
  {
    key: "date",
    value: "Date",
    display: (date: rows[number]["date"]) => new Date(date).toLocaleString(),
    sort: (a: rows[number]["date"], b: rows[number]["date"]) => (new Date(a) <= new Date(b) ? -1 : 1),
  },
];
```

But, in order to force `rows` array to have objects with a set of keys:

```ts
// Typing `rows` with a `headersList` array
const headersList = ["name", "quantity", "date"] as const;

type DataTableRows = {
  [P in typeof headersList[number] | "id"]: string;
};

const rows: DataTableRows = [
  {
    id: 1,
    nam: "John", // Will trigger a TS error
    quantity: 157.89,
    date: "2020-10-21",
  },
];
```

albertms10:
Answering https://github.com/IBM/carbon-components-svelte/pull/356#issuecomment-716056350, having an interface like the following would allow typing the headers array this way:

```ts
interface Header<T = unknown> {
  key: string;
  value: string;
  display?: (value: T) => string;
  sort?: (a: T, b: T) => number;
}

const headers: [Header<string>, Header<number>, Header<string>] = [
  {
    key: "name",
    value: "Name",
  },
  {
    key: "quantity",
    value: "Quantity",
    display: (quantity) => quantity + " €",
  },
  {
    key: "date",
    value: "Date",
    display: (date) => new Date(date).toLocaleString(),
    sort: (a, b) => (new Date(a) <= new Date(b) ? -1 : 1),
  },
];
```

metonym:
Addressed by #1954

Issue #304, opened 2020-10-06
Fix IntrinsicAttributes error in TypeScript definitions


metonym:
This error has appeared several times now: #253, #300, #303

---

Component props should be typed as the intersection of the semantic element props and props exported from the component.

Furthermore, potentially adding `{ [key: string]: any; }` can serve as a catch-all for custom component attributes. A common use case would be `data-{X}` selectors used for testing/analytics.

For example, the following should not throw a TS error:

```jsx
<Link data-test style="color: red;" href="#">Link</Link>
```

The `data-test` and `style` attributes are not explicitly defined in the Link component, but they should still be valid.

metonym:
If the ["Strongly Typed Svelte Component Interface"](https://github.com/sveltejs/rfcs/pull/37) RFC is accepted and implemented in the language tools, writing component library definitions should be a lot easier. Attempting to auto-generate definitions may work on the surface for props/slots but is untenable in the long run. Especially since it's not the official way, it's prone to breakage if the internal API changes.

Again, consider the Link component as mentioned above. If `disabled` is `true`, the intrinsic element changes from `a` to `p`. If I were a TypeScript user, I'd be confused if a `disabled` Link permits anchor element attributes yet renders a paragraph element in the DOM.

Here is what a type definition for the Link component might look like if using the API proposed in the RFC:

```ts
// Link.d.ts
import { SvelteComponent } from "svelte";

// share across components
interface ForwardedEvents {
  click: MouseEvent;
  mouseover: MouseEvent;
  mouseenter: MouseEvent;
  mouseleave: MouseEvent;
}

interface SharedLinkProps {
  inline?: boolean;
  visited?: boolean;
}

interface DisabledLinkProps extends Partial<HTMLParagraphElement>, SharedLinkProps {
  disabled: true;
  ref?: HTMLParagraphElement;
}

interface EnabledLinkProps extends Partial<HTMLAnchorElement>, SharedLinkProps {
  disabled: false;
  ref?: HTMLAnchorElement;
}

type LinkProps = EnabledLinkProps | DisabledLinkProps;

export default class Link extends SvelteComponent<
  LinkProps,
  ForwardedEvents,
  { default: {} }
> {}
```

semenodp:
Hello. Is there any update on this issue? Would be happy to help adding `{ [key: string]: any; }` if that's the solution you're taking.

metonym:
[...]

PR #385, opened 2020-11-03
refactor(types): rewrite component TypeScript definitions

metonym:
This PR rewrites the TypeScript definitions from scratch.

**Purpose**

- account for intrinsic element attributes if `$$restProps` is spread to a native HTML Element (#304 )
- accurately represent the shape of the library so direct imports work
- improve typings for prop default values by deferring to the Svelte compiler instead of JSDocs
- improve typings for forwarded/dispatched events
- improve typings for slots, slot props

**Refactor**

- adjust existing JSDoc annotations to be more concise
- remove "." from component file names to improve direct imports
- update `typeof import("carbon-icons-svelte/lib/Add16").default` to `import("carbon-icons-svelte").CarbonIcon`
- remove type annotation for `id`, `ref` props where applicable

**Docs**

- replace PUBLIC_API.json with COMPONENT_API.json generated from "sveld"
- enhance component API documentation by linking to component source code, labelling reactive props, describing if the component spreads `$$restProps`

**Todo**

- [x] further improve slot prop typings (conservatively, some will fallback to `any`)
- [ ] ~~some interfaces should be exported for the consumer (i.e. DataTable `headers` in #364 )~~ (defer)
- [x] use `svelte-check` to validate TS definitions in the CI

**Errors**

- [x] props must extend `svelte.JSX.HTMLAttributes<HTMLElementTagNameMap["div"]>`; an intersection will cause errors
- [x] components without `$$restProps` (or if it's ambiguous like `Tag`) should conservatively allow any attributes
- [x] remove "undefined" dispatched events


albertms10:
Wow, nice job! 👏

Just asking… Will union types also be rewritten to TS syntax? (As do the `.d.ts` output type files)

Before:
```ts
/**
 * Specify alignment of accordion item chevron icon
 * @type {"start" | "end"}
 */
export let align = "end";
```

After:
```ts
/** Specify alignment of accordion item chevron icon */
export let align: "start" | "end" = "end";
```

metonym:
@albertms10 That's good thinking but unfortunately that would make the uncompiled Svelte components TS only.

src/Button/Button.svelte
```diff
@@ -1,92 +1,76 @@
...
-   * Set to `true` for the icon-only variant
-   * @type {boolean} [hasIconOnly=false]
-   */
+  /** Set to `true` for the icon-only variant */
   export let hasIconOnly = false;
 
   /**
    * Specify the icon from `carbon-icons-svelte` to render
-   * @type {typeof import("carbon-icons-svelte/lib/Add16").default} [icon]
+   * @type {typeof import("carbon-icons-svelte/lib/Add16").default}
```
albertms10, line 19: Instead of typing the icon prop with the default `Add16` icon, wouldn’t it be cleaner using the `typeof CarbonIcon` class?—it applies to the rest of icon imports.

```ts
/**
 * Specify the icon from `carbon-icons-svelte` to render
 * @type {typeof import("carbon-icons-svelte").CarbonIcon}
 */
export let icon = undefined;
```

Although in this way you can’t restrict the icon size to e.g. `16`… 🤔

metonym: Nice suggestion. The icon size should not matter.

PR #391, opened 2020-11-07
Remove bx--body--with-modal-open from body only when there are no visible modals

weaseldotro:
If multiple modals are used, this correctly handles the cleanup after the modals close.

### DIFF
--- a/src/ComposedModal/ComposedModal.svelte
+++ b/src/ComposedModal/ComposedModal.svelte
@@ -83,7 +83,9 @@
   });
 
   onDestroy(() => {
-    document.body.classList.remove("bx--body--with-modal-open");
+    if(document.querySelectorAll('.bx--modal.is-visible').length === 0) {
+      document.body.classList.remove("bx--body--with-modal-open");
+    }
   });
 
   afterUpdate(() => {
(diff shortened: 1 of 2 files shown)


metonym:
Not following this. How can there be multiple modals visible?

weaseldotro:
For example you click a button which opens a `ComposedModal`, which contains a button which opens a `Modal`, which could contain a confirmation message. After you close the confirmation modal, there is still the initial modal on the screen but the body doesn’t have the `bx--body--with-modal-open` class, so if you scroll the browser window then the body element scrolls, not the modal contents.

metonym:
So for that case there are nested modals?

weaseldotro:
Correct.

metonym:
IMO, dealing with nested modals isn't within the scope of this library.

weaseldotro:
So you won’t accept this pull request?

Because the library provides a modal component, the possibility of using nested modals sounds pretty reasonable to me, so why not support it? I don’t see how this could have a negative impact.

metonym:
From a design perspective, I'm opposed to adding logic outside of how the component is intended to be used. If having multiple modals visible at once is a requirement, you can always write your own logic to deal with this.

weaseldotro:
Alrighty then, I’ll close this.

metonym:
@weaseldotro If you strongly believe that nested modals should be supported, I'd recommend bringing this up to the core Carbon team in the [Carbon monorepo](https://github.com/carbon-design-system/carbon/issues) to see what they think. If the behavior is supported by the flagship Carbon library, then it's a no brainer that we support it here.

As always, I appreciate your contributions and feedback.

weaseldotro:
No worries, it’s your library and you do what you think it’s best. I am not going to take this up the ladder :P
I contribute because other people may have the same problems and my solutions could help them as well. Also, sometimes it’s less hacky / more elegant to implement the solutions in the root library than directly in the app, like in this instance.

PR #392, opened 2020-11-07
docs: add line replace method for code tags

### LINKED ISSUE #390: docs: use rich text formatting in Component API Props description
It may be better to parse the description field in the Component API Props documentation to show code text as `code` instead of \`code`:

docs with plain text

docs using code tag

A simple RegExp replace method should be enough, if a string can be parsed as HTML:

```ts
"`hasIconOnly` must be set to `true`.".replace(/`(.*?)`/g, "<code>$1</code>");
// Output: "<code>hasIconOnly</code> must be set to <code>true</code>."
```

So, in [ComponentApi.svelte:115](https://github.com/IBM/carbon-components-svelte/blob/8c304521691c3e25a1670e3f73d18cc9c59ee5cc/docs/src/components/ComponentApi.svelte#L115),

https://github.com/IBM/carbon-components-svelte/blob/8c304521691c3e25a1670e3f73d18cc9c59ee5cc/docs/src/components/ComponentApi.svelte#L114-L116

`line` should be appended with the replace method described above:

```svelte
<div class="description">{@html line.replace(/`(.*?)`/g, "<code>$1</code>")}.</div>
```

albertms10:
closes #390

---

- docs: add line replace method for code tags

### DIFF
--- a/docs/src/components/ComponentApi.svelte
+++ b/docs/src/components/ComponentApi.svelte
@@ -112,7 +112,9 @@
             </StructuredListCell>
             <StructuredListCell>
               {#each prop[1].description.split('\n') as line}
-                <div class="description">{line}.</div>
+                <div class="description">
+                  {@html line.replace(/`(.*?)`/g, '<code>$1</code>')}.
+                </div>
               {/each}
             </StructuredListCell>
           </StructuredListRow>


metonym (approved):
Looks good, thanks!

PR #393, opened 2020-11-09
feat(ComposedModal): allow using null for selectorPrimaryFocus to disable auto focus

weaseldotro:
Useful if you want to handle the focusing yourself or do not want the modal to auto focus on anything at all after mounting.
If you don't pass the prop `selectorPrimaryFocus`, everything works as before, the modal will focus on the primary button (`buttonRef`) after mounting.
If you pass a string value to the `selectorPrimaryFocus`, the modal will focus on the value using `querySelector` after mounting.
If you now pass `selectorPrimaryFocus={null}`, the modal will not focus on anything after mounting.

### DIFF
--- a/src/ComposedModal/ComposedModal.svelte
+++ b/src/ComposedModal/ComposedModal.svelte
@@ -69,7 +69,7 @@
   });
 
   function focus(element) {
-    if(selectorPrimaryFocus === null) {
+    if(selectorPrimaryFocus == null) {
       return
     }
     const node =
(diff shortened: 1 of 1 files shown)


src/ComposedModal/ComposedModal.svelte
```diff
@@ -69,6 +69,9 @@
   });
 
   function focus(element) {
+    if(selectorPrimaryFocus === null) {
```
metonym, line 72:
```suggestion
    if(selectorPrimaryFocus == null) {
```

A loose equals would suffice here

metonym (changes requested):

metonym (approved):

PR #407, opened 2020-11-22
fix(RadioButtonGroup): make selected prop reactive

### LINKED ISSUE #406: Selected prop on RadioButtonGroup is not reactive
Programmatically changing the `selected` prop does not trigger an update and the correct radio input is not checked.

Click on the select buttons: https://codesandbox.io/s/loving-grass-p466p?file=/App.svelte

weaseldotro:
Fixes #406

### DIFF
--- a/src/RadioButtonGroup/RadioButtonGroup.svelte
+++ b/src/RadioButtonGroup/RadioButtonGroup.svelte
@@ -40,7 +40,7 @@
       }
     },
     update: (value) => {
-      selectedValue.set(value);
+      selected = value;
     },
   });
 
--- a/src/RadioButtonGroup/RadioButtonGroup.svelte
+++ b/src/RadioButtonGroup/RadioButtonGroup.svelte
@@ -26,7 +26,7 @@
    */
   export let id = undefined;
 
-  import { createEventDispatcher, setContext } from "svelte";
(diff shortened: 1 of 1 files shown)


metonym (changes requested):
Nice work.

One requested change. Programmatically setting and binding `selected` works. However, if the user clicks one of the radio buttons, the update method from the "RadioButtonGroup" context must update `selected` and not `selectedValue`.

Otherwise, there may be two or more selected buttons:

Screen Shot 2020-11-22 at 6 23 04 AM

```diff
setContext("RadioButtonGroup", {
    selectedValue,
    add: ({ checked, value }) => {
      if (checked) {
        selectedValue.set(value);
      }
    },
    update: (value) => {
+      selected = value;
-       selectedValue.set(value);
    },
  });
```

weaseldotro:
Oops, somehow I missed testing that. I have added the requested changes.

To correctly test this, in my test program https://codesandbox.io/s/loving-grass-p466p?file=/App.svelte I have replaced ```<RadioButtonGroup selected={plan}>``` with ```<RadioButtonGroup bind:selected={plan}>``` to use two way binding, otherwise it won't work correctly when using both radios and buttons to change the options.

metonym (approved):

Issue #421, opened 2020-11-27
Fix FileUploader types


miedzikd:
FileUploader component types are not valid:

current state:
```
Files: string[];
```

should be:
```
Files: FileList
```

output from console:
![image](https://user-images.githubusercontent.com/21012696/100422555-d14c7100-308a-11eb-955e-6dc1c86742a4.png)

metonym:
Released in [v0.25](https://github.com/IBM/carbon-components-svelte/releases/tag/v0.25.0)

PR #422, opened 2020-11-27
fix types for FileUploader component

miedzikd:
Related issue: #421


miedzikd:
What I did:
- I changed the `Files` type from `string[]` to `FileList`.
- Then I realized that there is no need to overwrite `FileList` with `Files`, so I used `FileList` everywhere ...
- Since we also used `Files` in the accepted formats, I changed it to a plain `string[]`

You can find my changes in:
- src/FileUploader/FileUploader.svelte
- src/FileUploader/FileUploaderButton.svelte
- src/FileUploader/FileUploaderDropContainer.svelte

Other files are autogenerated

metonym (approved):
Thanks for fixing this

Issue #368, opened 2020-10-26
When adding new tabs, the selected tab has an incorrect index


weaseldotro:
The current implementation of Tabs assumes that new tabs are always added to the end of the tabs' list, it doesn't take into consideration that tabs can be inserted in the middle of the list. So the index of the new tab is always incremented based on the last index, it is not calculated based on the tab position.

Example: https://codesandbox.io/s/nervous-meitner-s0pju

Click on toggle green tab button, then click on the tab titled green. The green tab is the second one in the list, so it should have index of 1, however it has an index of 2.

Try playing with the buttons and toggling tabs then clicking on them. The tab index will be incremented each time a new tab is added, so the `selected` prop will also have an incorrect value.

The correct implementation should recalculate the indexes of the tabs each time a tab is added or removed.

metonym:
Agreed, this is unexpected.

For the case of inserting instead of appending a new tab – is it possible in the current approach of composition?

**Current**

```svelte
<Tabs>
  <Tab ... />
  <Tab ... />
<div slot="content">
  <TabContent .../>
  <TabContent .../>
</div>
</Tabs>
```

The API may need to change so that tabs are passed as a single prop, and display is controlled via slots:

```svelte
<Tabs tabs={[{id: 0, label: '1'}, {id: 1, label: '2'}]}>
<div slot="label" let:index let:id>
  Label {id} {index}
</div>
<div slot="content" let:index let:id>
  ...
</div>
</Tabs>
```

metonym:
If you're reading this from the future, the work around is to use the `#key` API to re-render the component.

```svelte
<script>
  import { Button, Tabs, Tab, TabContent } from "carbon-components-svelte";

  let tabs = [
    { title: "blue", visible: true },
    { title: "green", visible: false },
    { title: "red", visible: true },
  ];

  const toggleTab = (title) => {
    tabs = tabs.map((tab) => ({
      ...tab,
      visible: tab.title === title ? !tab.visible : tab.visible,
    }));
  };

  let selected = 0;
</script>

<svelte:options immutable />

{#key tabs}
  <Tabs bind:selected>
    {#each tabs as tab (tab.title)}
      {#if tab.visible}
        <Tab label="{tab.title}" />
      {/if}
    {/each}
    <div slot="content">
      {#each tabs as tab (tab.title)}
        {#if tab.visible}
          <TabContent>Content for {tab.title}</TabContent>
        {/if}
      {/each}
    </div>
[...]

PR #435, opened 2020-11-29
proof of concept for keyed tabs

metonym:
#368

**Problem**

Currently, the `Tabs` component cannot track if new tabs are dynamically added or removed.

The current design practices component composition. That is – props and slots could be defined for individual components instead of defining data through a single prop (typically an object or array).

The Svelte context API is used to "add" a `Tab` or `TabContent` to the parent `Tabs`. The problem with this approach is that newly added child components can disrupt the existing components tracked by `Tabs` when it is first mounted. Unlike React.js, for example, there isn't (currently) an API to programmatically access or manipulate slots.

```svelte
<Tabs>
  <Tab label="Tab label 1" />
  <Tab label="Tab label 2" />
  <Tab label="Tab label 3" />
  <div slot="content">
    <TabContent>Content 3</TabContent>
  </div>
</Tabs>
```

**Proposed Solution**

The solution is to forgo the "component composition" approach in favor of the "object configuration" route.

A single prop (e.g., "items") is used to render/track the tabs that can be programmatically updated. The prop requires an "id" key for each object.

In this refactored proof of concept, "TabsV2" offers two reactive props that the consumer can bind to:

- selectedIndex: the current tab index
- selectedId: the current tab id

```svelte
<script>
  let items = [
    { id: 0, label: "Tab 1" },
    { id: 1, label: "Tab 2", disabled: true },
    { id: 2, label: "Tab 3" },
  ];
</script>

<TabsV2
  bind:selectedId
  bind:selectedIndex
  {items}
  let:item
  let:id
  let:index
  on:change="{(e) => {
    console.log('change', e.detail);
  }}"
>
  <span slot="tab" style="{!item.disabled && 'color: blue'}">
    {#if index === 1}
      <Add16 />
    {/if}
    {item.label}
  </span>
  {#if selectedIndex === 0}Tab content {id} {index}{/if}
  {#if selectedIndex === 1}Tab content {id} {index}{/if}
  {#if selectedIndex === 2}Tab content {id} {index}{/if}
  {#if selectedIndex === 3}Tab content {id} {index}{/if}
  {#if selectedIndex === 4}Tab content {id} {index}{/if}
</TabsV2>
```

**Pros**

- Only one component to import (instead of 3)
- Two reactive props (instead of 1)
- No useless wrapping div `slot='content'`
- Keyed

**Cons**

- Tab label props have to be specified in the items array (e.g., "disabled": true)
- Consumer must check if item is disabled when programmatically updating the selected item

---

Possible names:

- TabsV2 (implies that it will be promoted in the next major version)
- TabsKeyed

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -1,6 +1,6 @@
 # Component Index
 
-> 155 components exported from carbon-components-svelte@0.25.1.
+> 156 components exported from carbon-components-svelte@0.25.1.
 
 ## Components
 
@@ -134,6 +134,7 @@
 - [`TableRow`](#tablerow)
 - [`Tabs`](#tabs)
 - [`TabsSkeleton`](#tabsskeleton)
+- [`TabsV2`](#tabsv2)
 - [`Tag`](#tag)
 - [`TagSkeleton`](#tagskeleton)
 - [`TextArea`](#textarea)
@@ -3581,6 +3582,44 @@ None.
 | mouseover  | forwarded | --     |
 | mouseenter | forwarded | --     |
 | mouseleave | forwarded | --     |
+
+## `TabsV2`
+
+### Types
+
+```ts
+export type TabsV2ItemId = number | string;
+
+export interface TabsV2Item {
+  id: TabsV2ItemId;
+  label?: string;
+  disabled?: boolean;
+}
+```
+
+### Props
+
+| Prop name       | Kind             | Reactive | Type                                      | Default value                    | Description                                 |
+| :-------------- | :--------------- | :------- | :---------------------------------------- | -------------------------------- | ------------------------------------------- |
+| selectedId      | <code>let</code> | Yes      | <code>TabsV2ItemId</code>                 | --                               | Specify the selected tab id                 |
+| selectedIndex   | <code>let</code> | Yes      | <code>number</code>                       | <code>0</code>                   | Specify the selected tab index              |
+| items           | <code>let</code> | No       | <code>TabsV2Item[]</code>                 | <code>[]</code>                  | Provide the tab items                       |
+| type            | <code>let</code> | No       | <code>"default" &#124; "container"</code> | <code>"default"</code>           | Specify the type of tabs                    |
+| iconDescription | <code>let</code> | No       | <code>string</code>                       | <code>"Show menu options"</code> | Specify the ARIA label for the chevron icon |
+| triggerHref     | <code>let</code> | No       | <code>string</code>                       | <code>"#"</code>                 | Specify the tab trigger href attribute      |
+
+### Slots
+
+| Slot name | Default | Props                                                                | Fallback                  |
+| :-------- | :------ | :------------------------------------------------------------------- | :------------------------ |
+| --        | Yes     | <code>{ id: TabsV2ItemId; index: number; item: TabsV2Item; } </code> | --                        |
+| tab       | No      | --                                                                   | <code>{item.label}</code> |
+
+### Events
+
+| Event name | Type       | Detail                                                                                    |
+| :--------- | :--------- | :---------------------------------------------------------------------------------------- |
+| change     | dispatched | <code>{ selectedIndex: number; selectedId: TabsV2ItemId; currentItem: TabsV2Item }</code> |
 
 ## `Tag`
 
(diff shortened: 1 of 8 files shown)


metonym:
@weaseldotro Curious to hear your feedback.

Rendered example: https://carbon-components-svelte-git-tabs-keyed.carbon-svelte.vercel.app/framed/Tabs/TabsV2

weaseldotro:
From a visual point of view, the current markup for tabs is way more elegant and beautiful. The new code looks very spaghetti like with all those `if`s for each tab content instead of semantic elements. If I saw both these markups side by side, I would think v2 is actually v1 and vice versa.

From a programming point of view, the proposed solution seems more powerful. I don't understand why you kept using both a numeric index and a string key when an `id` key is required, I would have only used a string key and that's it. The selected index prop seems redundant and useless when you have a key.

There is a bug if I click on a tab then change focus using the arrow keys on the keyboard: the selected tab changes, but the focus remains on the initial tab. Also, if I press enter then the selected tab reverts to the focused tab.

metonym:
Thanks for the feedback.

> From a visual point of view, the current markup for tabs is way more elegant and beautiful.

Agreed. I think a lot of users (myself included) prefer component composition.

In fairness, the above example was intended to be exhaustive.

This would be the basic usage where the tab slot is not used and only `selectedIndex` is bound:

```svelte
<TabsV2 bind:selectedIndex {items}>
  {#if selectedIndex === 0}Tab content 1{/if}
  {#if selectedIndex === 1}Tab content 2{/if}
  {#if selectedIndex === 2}Tab content 3{/if}
</TabsV2>
```

> If I saw both these markups side by side, I would think v2 is actually v1 and vice versa.

This is important. V2 probably isn't the best name. Maybe `TabsKeyed` and only document it in the existing `Tabs` docs.

>  don't understand why you kept using both a numeric index and a string key when an id key is required, I would have only used a string key and that's it. The selected index prop seems redundant and useless when you have a key.

Yes, it's redundant – `selectedIndex` just provides an alternative (and vice versa).

For example, if I only bind the `id`, I would need to write my own logic to get its index from the items array (although this isn't a big deal).

> There is a bug if I click on a tab then change focus using the arrow keys on the keyboard: the selected tab changes, but the focus remains on the initial tab. Also, if I press enter then the selected tab reverts to the focused tab.

It needs to be aligned with the existing implementation

ispyinternet:
> Currently, the Tabs component cannot track if new tabs are dynamically added or removed.

Would they not only be dynamically added / removed by an API exposed by <Tabs>,  and therefore trackable? We shouldn't be manipulating the DOM directly?

ispyinternet:
Or am I missing the point, and they are added by conditional runtime svelte?

weaseldotro:
> In fairness, the above example was intended to be exhaustive.
> 
> This would be the basic usage where the tab item is not customized and only `selectedIndex` is bound:
> 
> ```svelte
> <TabsV2 bind:selectedIndex {items}>
>   {#if selectedIndex === 0}Tab content 1{/if}
>   {#if selectedIndex === 1}Tab content 2{/if}
>   {#if selectedIndex === 2}Tab content 3{/if}
> </TabsV2>
> ```
I meant that it looks like ugly and like spaghetti because of all the `if`s, not because of the props added to the `TabsV2` element.

> 
> > If I saw both these markups side by side, I would think v2 is actually v1 and vice versa.
> 
> This is important. V2 probably isn't the best name. Maybe `TabsKeyed` and only document it in the existing `Tabs` docs.
> 
I meant that this "new" implementation looks like it was actually the first (previous) version of the component and the current implementation is actually the the second (new & better) version.

> > don't understand why you kept using both a numeric index and a string key when an id key is required, I would have only used a string key and that's it. The selected index prop seems redundant and useless when you have a key.
> 
> Yes, it's redundant – `selectedIndex` just provides an alternative (and vice versa).
> 
> For example, if I only bind the `id`, I would need to write my own logic to get its index from the items array (although this isn't a big deal).
> 
Yes I know it's redundant because some people might want to use a numeric index and not a key, but this also means more things to break and more code to maintain. Also, when dynamically changing the tabs the developer must take into consideration that the numeric index might also change. I think using just a key would simplify things a lot.

metonym:
@weaseldotro @ispyinternet 

It might be better to just use the `#key` block API to recreate the component if the tabs change.

See [PR #438](https://github.com/IBM/carbon-components-svelte/pull/438)

Note that the `selectedIndex` is not normalized if the tabs shrink in number and the previous selected index is greater than the current length.

PR #437, opened 2020-11-30
fix FileUploader

AsakuraMizu:
Related PR: #422 

`files` should be a FileList, not a list of files' names.


metonym:
Thanks for the PR.

It seems that the mismatch is because [FileList type](https://developer.mozilla.org/en-US/docs/Web/API/FileList) is an object of Symbol iterators (files) and spreading `target.files` (`[...target.files]`) converts it into an array.

Two options:

1) To keep with the `files: FileList` type, do not spread `target.files`

This would require updating the rendering logic within `FileUploader` (`{#each files as { name }, i (name)}`)

2) Change files type to `File[]`

I'm inclined to go with option 1 because the `files` prop would be consistent with the native value of `e.target.files`.

src/FileUploader/FileUploader.svelte
```diff
@@ -95,11 +95,11 @@
     kind="{kind}"
     on:change
     on:change="{({ target }) => {
-      files = [...target.files].map(({ name }) => name);
+      files = [...target.files];
```
metonym, line 98:
```suggestion
      files = target.files;
```

To align with the `FileList` type, `target.files` should not be spread.

src/FileUploader/FileUploader.svelte
```diff
@@ -95,11 +95,11 @@
...
     }}"
   />
   <div class:bx--file-container="{true}">
-    {#each files as name, i (name)}
+    {#each files as { name }, i (name)}
```
metonym, line 102: ... but if the `FileList` type is kept intact, the rendering logic would need to be updated, which expects `files` to be an array

metonym (changes requested):

src/FileUploader/FileUploader.svelte
AsakuraMizu: This is not only the logic to update, but also at anywhere which use files as an array, such as   [L71](https://github.com/AsakuraMizu/carbon-components-svelte/blob/410a805191350e9537d1f77455cf76536291fa87/src/FileUploader/FileUploader.svelte#L71)(`files.includes`), [L112](https://github.com/AsakuraMizu/carbon-components-svelte/blob/410a805191350e9537d1f77455cf76536291fa87/src/FileUploader/FileUploader.svelte#L112) and [L117](https://github.com/AsakuraMizu/carbon-components-svelte/blob/410a805191350e9537d1f77455cf76536291fa87/src/FileUploader/FileUploader.svelte#L117)(`files.filter`). So maybe changing its type to `File[]` is a better solution.

metonym: @AsakuraMizu makes sense – another point is that the end user probably expects an array type anyway.

metonym (approved):
excellent

Issue #428, opened 2020-11-27
RadioTile support 'multiple' prop?


ispyinternet:
Initially I was looking at SelectableTile to control a selection of filters, firstly with there being nothing reactive about it, it looked like a bit of clunky end user wiring would be required - that's another issue I wonder should be looked at?

Anyway for my use case, I was happy to allow only 1 filter, so I opted for the much more succint RadioTile, but it did get me thinking if I wanted to support more than one selection, should this component support a `multiple` prop?

If you did want to implement, how would you like to handle the value? Array type when multiple, or Array type always?

metonym:
> ...should this component support a multiple prop?

Radio buttons must only have one selection ([Nielsen Norman](https://www.nngroup.com/articles/radio-buttons-default-selection/)).

I'll review the SelectableTile.

ispyinternet:
I guess I've also just discovered that at least one must be selected as I look for a way to empty the filters!

Yes I think if SelectableTile could fill the gap that would be great. If you want to suggest a solution I can look at it. Would an optional wrapper component be a feesible solution?

metonym:
Sure – that approach is sound.

Similar to RadioTileGroup, the wrapper component (e.g., SelectableTileGroup) would manage the selected ids using the context API.

ispyinternet:
1. Do you want to rename TileGroup -> RadioTileGroup ?
2. Or what do you think about keeping TileGroup and using a prop to specify radio or select? that would then make RadioTile and SelectableTile one in the same -> Tile with `selectable` prop?

metonym:
I like your thinking of reusing `TileGroup`.

One issue though: the `selected` prop type would need to change to be an array of values. So for a RadioTile you would need to do `selected[0]`.

I'm leaning towards keeping them separate. Preserve the `TileGroup` component but add two new ones dedicated to the SelectableTile, RadioTile respectively:

- SelectableTileGroup
- RadioTileGroup

So we can deprecate TileGroup in the next major release in one fell swoop.

ispyinternet:
At the moment, I guess these tile groups are suffering the same fate as the Tabs in that they are not programmatically mutable?

Might be good to try and address this with the newer versions? Although I wasn't keen on the sight of the all the if statements! Are there no other options / techniques on the table?

ispyinternet:
[...]

Issue #423, opened 2020-11-27
Inconsistent use of rest props with DataTable and DataTableSkeleton


ispyinternet:
For me this appeared because I wanted to control the class on the DataTable container to add x-scroll. I would like that to be consistent across the DataTable and the DataTableSkeleton, however in the skeleton, you pass the rest props directly to the table bypassing the data table container (`bx--data-table-container`), whereas in the DataTable you apply rest props to the container.

I would submit a fix, but I'm unsure how you want to address this?

metonym:
I agree – spreading rest props to the same class name "bx--data-table-container" makes sense.

I think that spreading rest props in `DataTableSkeleton` to the top-level "bx--data-table-container" element instead of the `table` element would be the best approach. If I recall correctly, the initial top-level element in `DataTableSkeleton` was the table element. The container div was added later.

ispyinternet:
I've actioned that, do I need to do anything to keep the git commits / history clean?

https://github.com/IBM/carbon-components-svelte/compare/master...ispyinternet:master

metonym:
Feel free to submit a PR.

metonym:
This was released in v0.26.0

PR #439, opened 2020-11-30
WIP: SelectableTileGroup

ispyinternet:


src/Tile/TileGroup.svelte
```diff
@@ -17,7 +17,7 @@
   const dispatch = createEventDispatcher();
   const selectedValue = writable(selected);
 
-  setContext("TileGroup", {
+  setContext("RadioTileGroup", {
```
metonym, line 20: Refrain from making changes to TileGroup if there is already a dedicated RadioTileGroup component

src/Tile/RadioTile.svelte
```diff
@@ -20,24 +23,31 @@
...
 
 <input
+  bind:this="{ref}"
   type="radio"
   id="{id}"
   name="{name}"
   value="{value}"
   checked="{checked}"
-  tabindex="{tabindex}"
+  tabindex="-1"
```
metonym, line 48: What's the reason for changing the tabindex?

src/Tile/RadioTile.svelte
```diff
@@ -8,6 +8,9 @@
   /** Specify the value of the radio input */
   export let value = "";
 
+  /** Specify the title of the selectable tile */
+  export let title = "title";
```
metonym, line 12: What's this prop for?

src/DataTableSkeleton/DataTableSkeleton.svelte
```diff
@@ -41,7 +41,11 @@
   );
 </script>
 
-<div class:bx--skeleton="{true}" class:bx--data-table-container="{true}">
+<div
```
metonym, line 44: I'd rather this be in a separate PR

src/Tile/RadioTile.svelte
```diff
@@ -20,24 +23,31 @@
...
   export let name = "";
 
+  /** Obtain a reference to the input HTML element */
+  export let ref = null;
+
   import { getContext } from "svelte";
   import CheckmarkFilled16 from "carbon-icons-svelte/lib/CheckmarkFilled16";
 
-  const { add, update, selectedValue } = getContext("TileGroup");
+  const { _light, add, update, selectedValue } = getContext("RadioTileGroup");
```
metonym, line 32: Note: this is another breaking change if a RadioTile can no longer be used with a TileGroup

metonym (changes requested):

metonym:
In addition to passing the CI/deploy preview, could you remove the committed `package-lock.json` files?

src/Tile/TileGroup.svelte
ispyinternet: I did update TileGroup to use this key so its backwards compatible? (just aligning naming convention for when it depreciates)

src/Tile/RadioTile.svelte
ispyinternet: I noticed in SelectableTile you place a -1 on the input and tabindex on the label and I thought I would align this.

src/Tile/RadioTile.svelte
ispyinternet: aligning with your SelectableTile

src/DataTableSkeleton/DataTableSkeleton.svelte
ispyinternet: ok

src/Tile/RadioTile.svelte
ispyinternet: Ah.

ispyinternet:
> In addition to passing the CI/deploy preview, could you remove the committed `package-lock.json` files?

Should I update .gitignore?

I also inadvertantly added palimpsest, should I add this to gitignore?

src/Tile/RadioTile.svelte
metonym: They're two different components.

The canonical implementation has it as "-1": https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/components/tile/tile--selectable.hbs#L8

ispyinternet: OK

src/Tile/RadioTile.svelte
ispyinternet: What do you think to this feature? Support it?

src/Tile/RadioTile.svelte
ispyinternet: Is there a rationale for one to have and the other not?

src/Tile/RadioTile.svelte
metonym: IMO this already deviates from the initial issue to improve the DX for SelectableTile better.

ispyinternet: ok

ispyinternet:
BTW, you said in the issue comments to use `selectedItem` `selectedItems` which also is breaking? Wasn't sure if you realised?

metonym:
How about creating separate PRs that address the issues individually?

- DataTableSkeleton #423
- SelectableTileGroup #428

metonym:
> BTW, you said in the issue comments to use `selectedItem` `selectedItems` which also is breaking? Wasn't sure if you realised?

`selectedValue` and `selectedValues` would be new props if `SelectableTileGroup` and `RadioTileGroup` components are introduced.

src/Tile/SelectableTileGroup.svelte
```diff
@@ -0,0 +1,52 @@
+<script>
+  /**
+   * Specify the selected tile's
+   * @type {string}
```
metonym, line 4:
```suggestion
   * @type {string[]}
```

ispyinternet: Do you think that should be `any` array? Since the values represent the values you pass to the SelectableTile, and you could give it any type of value?

metonym: Yes, good point. We should loosen the type here

ispyinternet: would that be written `@type {any[]}` ?

metonym: Yes

ispyinternet:
Wow, that was painful. Mainly due to my own ineptitude, but having to push to debug. Can you advise on a setup to test the components locally.

I'm sorry If you've been getting pinged senseless over these commits!

1 or 2 unanswered above and advise me of any other changes you require.

src/Tile/RadioTile.svelte
metonym: Remove

src/Tile/SelectableTileGroup.svelte
```diff
@@ -0,0 +1,48 @@
+<script>
+  /**
+   * Specify the selected tile's
+   * @type {any[]}
+   */
+  export let selectedValues = [];
+
+  /** Set to `true` to disable the tile group */
+  export let disabled = false;
```
metonym, line 9: remove

src/Tile/SelectableTileGroup.svelte
```diff
@@ -0,0 +1,48 @@
...
+    } else if (!selected && i > -1) {
+      a.splice(i, 1);
+    }
+    return a;
+  }
+
+  setContext("SelectableTileGroup", {
+    selectedValues: _selectedValues,
+    update: ({ selected, value }) =>
+      _selectedValues.set(newArray(value, selected)),
```
metonym, line 34: - Use the store `.update` method
- Use filter/map Array methods

metonym (changes requested):
- remove any changes to TileGroup/RadioTile
- remove RadioTileGroup
- restore the "light" variant example for SelectableTile

metonym:
For testing locally, you can use the doc website as a UI.

Refer to CONTRIBUTING.md: https://github.com/IBM/carbon-components-svelte/blob/master/CONTRIBUTING.md#documentation-set-up

src/Tile/SelectableTileGroup.svelte
ispyinternet: just 8/9?

metonym: the `disabled` prop

ispyinternet:
> * remove any changes to TileGroup/RadioTile
> * remove RadioTileGroup
> * restore the "light" variant example for SelectableTile

You did originally ask to bring in the RadioTileGroup? I know i've made this painful - but do I need to strip out RadioTileGroup?

src/Tile/SelectableTileGroup.svelte
ispyinternet: I've updated the code to:

```js
  setContext("SelectableTileGroup", {
    selectedValues: _selectedValues,
    update: ({ selected, value }) =>
      _selectedValues.update(v => {
        const copy = [...v];
        const exists = copy.indexOf(value) > -1;
        if (selected && !exists) {
          copy.push(value);
        }
        else if (!selected && exists) {
          copy.splice(i, 1);
        }
        return copy;
      })
  });
```

What would your suggestion be?

ispyinternet:
> * remove any changes to TileGroup/RadioTile
> * remove RadioTileGroup
> * restore the "light" variant example for SelectableTile

I've restored the light, I think the reason I stripped it because when I reviewed the live docs, it wasn't actually being demod properly:

https://carbon-svelte.vercel.app/components/RadioTile

src/Tile/SelectableTileGroup.svelte
metonym: Something like

```
_selectedValues.update(_ => {
  if (_.includes(value)) {
    return _.filter(i => i !== value);
  } else {
   // since order does not matter
    return [..._, value]
  }
});
```

ispyinternet: 👍 jedi master.

ispyinternet:
i haven't removed RadioTileGroup yet - I'm fighting to keep it - everything is still backwards compatible?

src/Tile/SelectableTileGroup.svelte
ispyinternet: Hold on, what if update receives the value but with `selected=false` and it doesn't already exist on the array. Your going to add the item, even if the selected state is false?

metonym:
> i haven't removed RadioTileGroup yet - I'm fighting to keep it - everything is still backwards compatible?

Save it for another PR

src/Tile/SelectableTileGroup.svelte
ispyinternet: Yh the deployed code demonstrates that bug. I've redeployed a fix.

ispyinternet:
> > i haven't removed RadioTileGroup yet - I'm fighting to keep it - everything is still backwards compatible?
> 
> Save it for another PR

grunting to self.

src/Tile/RadioTile.svelte
ispyinternet: Ok actioned, but why so hard on the prop!?

docs/src/pages/components/RadioTile.svx
```diff
@@ -24,13 +25,13 @@ components: ["TileGroup", "RadioTile"]
 ### Light variant
 
 <TileGroup legend="Service pricing tiers">
-  <RadioTile value="0" checked>
+  <RadioTile light value="0" checked>
```
metonym, line 28: Thanks for fixing this in the docs

src/Tile/SelectableTileGroup.svelte
```diff
@@ -0,0 +1,39 @@
...
+
+  $: selectedValues = $_selectedValues;
+  $: dispatch("select", $_selectedValues);
+</script>
+
+<fieldset class:bx--tile-group="{true}" {...$$restProps}>
+  {#if legend}
+    <legend class:bx--label="{true}">{legend}</legend>
+  {/if}
+  <div>
```
metonym, line 36:
```suggestion
  <div role="group" aria-label="{$$props['aria-label'] || 'Selectable tiles'}">
```

`div` should have a group role; allow the user to override the aria label.

metonym:
Are you able to get two way binding on `selectedValues` to work?

```svelte
<script>
  let selectedValues = ["1"]
</script>

<SelectableTileGroup bind:selectedValues legend="Select the options you require">
  <SelectableTile title="Option 1" value="1">
    Option 1
  </SelectableTile>
  <SelectableTile title="Option 2" value="2">
    Option 2
  </SelectableTile>
  <SelectableTile title="Option 3" value="3">
    Option 3
  </SelectableTile>
</SelectableTileGroup>
```

src/Tile/SelectableTileGroup.svelte
ispyinternet: Revisiting: Why don't you want to support the disabled prop on the fieldset?

ispyinternet:
> Are you able to get two way binding on `selectedValues` to work?
> 
> ```svelte
> <script>
>   let selectedValues = ["1"]
> </script>
> 
> <SelectableTileGroup bind:selectedValues legend="Select the options you require">
>   <SelectableTile title="Option 1" value="1">
>     Option 1
>   </SelectableTile>
>   <SelectableTile title="Option 2" value="2">
>     Option 2
>   </SelectableTile>
>   <SelectableTile title="Option 3" value="3">
>     Option 3
>   </SelectableTile>
> </SelectableTileGroup>
> ```

I actually can't and I have no idea why the two-way binding is not working. I can't see how it differs from this REPL: 

https://svelte.dev/repl/hello-world?version=3.30.1

metonym:
@ispyinternet take a look at this: https://svelte.dev/repl/bf4a1ebbe12b49d38f0fcd15ca782a04?version=3.30.1

It supports three use cases using component composition:

1) Binding w/ a single prop

```svelte
<script>
  let sel = ['0']
</script>

<Parent bind:sel>
  <Child id="0" />
  <Child id="1" />
  <Child id="2" />
</Parent>
```

2) Binding w/ a single prop (initially selected children)

```svelte
<script>
  let sel = [];
</script>

<Parent bind:sel>
  <Child id="0" selected />
  <Child id="1" />
  <Child id="2" />
</Parent>
```

3) "controlled"

```svelte
<script>
  let controlled = [
    { id: "0", selected: true },
    { id: "1", selected: false },
    { id: "2", selected: false },
  ];
</script>

<Parent sel={controlled.filter(i => i.selected).map(i => i.id)}>
  {#each controlled as item}
    <Child id="{item.id}" bind:selected="{item.selected}" />
  {/each}
</Parent>
```

ispyinternet:
> @ispyinternet take a look at this: https://svelte.dev/repl/bf4a1ebbe12b49d38f0fcd15ca782a04?version=3.30.1
> 
> 
> 
> It supports three use cases using component composition:
> 
> 
> 
> ## 1) Binding w/ a single prop
> 
> 
> 
> ```svelte
> 
> <script>
> 
>   let sel = ['0']
> 
> </script>
> 
> 
> 
> <Parent bind:sel>
> 
>   <Child id="0" />
> 
>   <Child id="1" />
> 
>   <Child id="2" />
> 
> </Parent>
> 
> ```
> 
> 
> 
> ## 2) Binding w/ a single prop (initially selected children)
> 
> 
> 
> ```svelte
> 
> <script>
> 
>   let sel = [];
> 
> </script>
> 
> 
> 
> <Parent bind:sel>
> 
>   <Child id="0" selected />
> 
>   <Child id="1" />
> 
>   <Child id="2" />
> 
> </Parent>
> 
> ```
> 
> 
> 
> ## 3) "controlled"
> 
> 
> 
> ```svelte
> 
> <script>
> 
>   let controlled = [
> 
>     { id: "0", selected: true },
> 
>     { id: "1", selected: false },
> 
>     { id: "2", selected: false },
> 
>   ];
> 
> </script>
> 
> 
> 
> <Parent sel={controlled.filter(i => i.selected).map(i => i.id)}>
> 
>   {#each controlled as item}
> 
>     <Child id="{item.id}" bind:selected="{item.selected}" />
> 
>   {/each}
> 
> </Parent>
> 
> ```
> 
> 

Yh, I was being a complete donut I was binding to the wrong variable name 😩

ispyinternet:
So I've fixed the binding bug, since this is a new component, do you want to include support for `light` at the group level?

ispyinternet:
Hi @metonym did you get a chance to review this and do you want me me to look at anything? Cheers.

metonym:
@ispyinternet Does it fulfill the three test cases mentioned above?

1) bind w/ a single prop
2) bind w/ a single prop with initially selected children
3) "controlled"

ispyinternet:
I'm pretty sure it does, but I assumed that you would want to validate that yourself before commiting to master.

I had asked myself if there were any way to write tests against this? I don't think the current project supports this does it?

If you want me to prepare a demo to validate the three cases I would be happy to (in the coming days)

ispyinternet (commented):
> @ispyinternet Does it fulfill the three test cases mentioned above?
> 
> 1. bind w/ a single prop
> 2. bind w/ a single prop with initially selected children
> 3. "controlled"

@metonym Do you think I can demonstrate answers to these 3 questions by writing appropriate demo's in the docs?

metonym:
@ispyinternet I added an example that tests the three use cases: https://carbon-components-svelte-kqfunrl8x.vercel.app/framed/SelectableTile/SelectableTileReactive

Click around in the first two. The bound `selectedValues` is indeed updating correctly by clicking tiles. However, I'm unable to programmatically update the `selectedValues` by clicking "reset selectedValues."

The issue here is that `SelectableTileGroup` has to update the internal selected values store based on the exported selectedValues prop. This is illustrated in the [REPL link](https://svelte.dev/repl/bf4a1ebbe12b49d38f0fcd15ca782a04?version=3.30.1) that I'd shared.

ispyinternet:
Ok ill get this sorted.

ispyinternet:
> In the third use case ("controlled"), I am able to reset the selectedValues, but when I click on the one of the tiles the selectedValues is incorrectly updated.

@metonym what I'm struggling with atm, is the additional (guessable) api of the `selectedValues` prop.

That is, we can either supply an array of scalars, or an array of `{ value, selected}` objects.

In the later case, suppose we empty the array, (unselect all), then select, we will need to know what the user expects back - an array of scalar values or an array of object types. We would have to infer by the initial incoming state and potentially 'remember'. There is of course the potential then to allow users to provide:

```
let selectedValues = [
1, // assume { value: 1, selected: true }
2,
{ value: 'a', selected: true }
]
```

I guess if we detect any item of the 'object form' in the initial creation, then always return that form?

metonym:
@ispyinternet Perhaps I was unclear. Take the Svelte REPL as a direct reference.

> That is, we can either supply an array of scalars, or an array of { value, selected} objects.

Not sure what you mean by this.

`selectedValues` should be `string[]`. It shouldn't be an array of objects. See the Svelte REPL.

ispyinternet:
> @ispyinternet Perhaps I was unclear. Take the Svelte REPL as a direct reference.
> 
> 
> 
> > That is, we can either supply an array of scalars, or an array of { value, selected} objects.
> 
> 
> 
> Not sure what you mean by this.
> 
> 
> 
> `selectedValues` should be `string[]`. It shouldn't be an array of objects. See the Svelte REPL.

Ok (array any!) however your framed example did bind to selectedValues3 which is the array of objects!

docs/src/pages/framed/SelectableTile/SelectableTileReactive.svelte
```diff
@@ -0,0 +1,86 @@
...
+    on:click="{() => {
+      selectedValues2 = [];
+    }}"
+  >
+    Reset selectedValues2
+  </Button>
+</div>
+
+<SelectableTileGroup
+  selectedValues="{selectedValues3}"
```
ispyinternet, line 64: @metonym here the incoming is array of objects - was that a mistake?

metonym: Yes, this was a mistake. It should be mapping the `value` from the object. The REPL has the correct example

ispyinternet:
@metonym I have to admit I'm stumped on the last example. I can't get an atomic update of the store. Basically, I'm reasoning that as the control updates the each loop, when the first item updates, the store updates, which passes an update back down to the children - it will only reset one item at a time.

PR #441, opened 2020-12-01
Move rest props in DataTableSkeleton

ispyinternet:
move restprops from `<table>` to table container `div` to align with DataTable usage


metonym (approved):
Looks good – thanks!

PR #444, opened 2020-12-04
fix(number-input): decimal input

AsakuraMizu:
It is really annoying with inputting decimals in NumberInput because it "auto" remove the suffix zeros.

~~And run `prepack` because @ispyinternet had missed it~~


metonym (changes requested):
The non-mobile input should be updated as well: https://github.com/IBM/carbon-components-svelte/blob/e3a152fc7d083dd22a2132d553e4f6a94eae3e89/src/NumberInput/NumberInput.svelte#L235

AsakuraMizu:
Wait a minute, I have found a bug.

metonym (approved):

Issue #476, opened 2021-01-03
OverflowMenu styling and focus invalid when opened for the first time


ebeuker:
When the overflow menu is opened for the first time:
- The menu contains a border
- The first menu item is not selected
![image](https://user-images.githubusercontent.com/754901/103478646-6ea46e80-4dc8-11eb-9f78-8ed654b12db8.png)

When you open the over flow menu for a second or third time:
- The menu no longer contains a border
- The first menu item is selected
![image](https://user-images.githubusercontent.com/754901/103478730-fab69600-4dc8-11eb-8d95-69b6e9bfbf65.png)

I think that this issue has been introduced with the fix for issue #191

metonym:
Published to NPM in v0.27

PR #477, opened 2021-01-03
Overflow menu first time open behavior is not correct

ebeuker:
This PR contains a fix for issue #476 

The current implementation prevents the focus to be set when it is opened for the first time. The code in this PR prevents the focus to be set during the initial after update lifecycle method call. The initial lifecycle call is called after the component is mounted.


metonym (approved):
Awesome! Thanks for getting to the bottom of this.

PR #539, opened 2021-02-26
Allow disable attribute for SelectableTile

lalilaloe:
I have seen #439 but tiles should be able to be disabled per tile. If for example one item in the list of selectable tiles is not available.


metonym:
I agree. However, the issue is that there is no styling for the disabled state.

I believe this will be supported in Carbon v10.32 though.

Refs: https://github.com/carbon-design-system/carbon/commit/91b6c05a59ede97b34aaa79b982d702a440b92d5

PR #546, opened 2021-02-28
Replace onDestroy with onMount for SSR support in ComposedModal

### LINKED ISSUE #200: Modal dialog is incompatible with Sapper SSR
When using SSR, the Modal component doesn't work because it has afterUpdate and onDestroy hooks that reference `document` leading to SSR errors in Sapper. 

https://github.com/IBM/carbon-components-svelte/blob/64e914ff8c6556cb12ec29401ddddbd8a342011e/src/components/Modal/Modal.svelte#L41

ItsiAdam:
Fixes #200 for ComposedModal, same as #207 

Could this be a problem for any additional components?

### DIFF
--- a/src/ComposedModal/ComposedModal.svelte
+++ b/src/ComposedModal/ComposedModal.svelte
@@ -36,7 +36,6 @@
     setContext,
     onMount,
     afterUpdate,
-    onDestroy,
   } from "svelte";
   import { writable } from "svelte/store";
 
@@ -75,10 +74,9 @@
   onMount(async () => {
     await tick();
     focus();
-  });
-
-  onDestroy(() => {
-    document.body.classList.remove("bx--body--with-modal-open");
+    return () => {
+      document.body.classList.remove("bx--body--with-modal-open");
+    };
   });
 
   afterUpdate(() => {


metonym:
Can you elaborate on the issue? How can I reproduce it? Are you using Sapper?

src/ComposedModal/ComposedModal.svelte
```diff
@@ -75,10 +74,9 @@
   onMount(async () => {
     await tick();
     focus();
-  });
-
-  onDestroy(() => {
-    document.body.classList.remove("bx--body--with-modal-open");
+    return () => {
```
metonym, line 77: The return function will not be triggered if the `onMount` callback is async

ItsiAdam:
> Can you elaborate on the issue? How can I reproduce it? Are you using Sapper?

Yes, I believe this is only an issue when using Sapper.

nickolasgregory:
I've encountered this when using [SvelteKit](https://kit.svelte.dev). Weirdly it doesn't happen all the time.
Was fixed in #200 returned in #348

src/ComposedModal/ComposedModal.svelte
ItsiAdam: Okay, does it need to be async?

Florian-Schoenherr: in Modal it's not async either

metonym:
Closing this as the code in discussion is long out-of-date.

Both `Modal` and `ComposedModal` toggle the body class in the `onMount` lifecycle method with a synchronous callback.

If you do encounter this again, a minimal repro with the Sapper/SvelteKit versions would aid in debugging.

PR #547, opened 2021-03-01
feat(ComposedModal): forward keydown event and add shouldSubmitOnEnter

weaseldotro:


metonym:
The dispatched "submit" event fires twice:

Screen Shot 2021-03-20 at 10 51 50 AM


https://github.com/IBM/carbon-components-svelte/blob/master/src/ComposedModal/ComposedModal.svelte#L54

weaseldotro:
The event would fire twice when pressing enter if the focus was on the modal footer primary button. I added an exception for this case.

metonym:
Can you provide steps that I can test? What is the expected behavior?

weaseldotro:
I'm using a `Form` inside a `ComposedModal` `ModalBody` and I want to submit the form when I press enter while the focus is on one of the form's components.

Here's some code: https://codesandbox.io/s/objective-chaplygin-9k645?file=/App.svelte
I type the username, then I press tab and and I type the age, then I want to press enter to submit the form, while the focus is still in on the age input.

If you have a better solution than my PR, I'm open to suggestions.

metonym:
IMO, this relates more to how you're using the form element. I believe that if you have two or more input elements within a form, the "submit" event will not fire when pressing enter unless you have a "submit" button.

What about forwarding the keydown event to the Form component?

weaseldotro:
> IMO, this relates more to how you're using the form element. I believe that if you have two or more input elements within a form, the "submit" event will not fire when pressing enter unless you have a "submit" button.

Yes, pressing enter if I don't have a submit button inside the form doesn't submit the form. The problem with `ComposedModal` is that the primary button in the footer doesn't have a `submit` type, so even if I put the modal inside a form then it still wouldn't submit the form when I press enter.

> What about forwarding the keydown event to the Form component?

I don't understand how this would help. Can you please elaborate?

metonym:
> I don't understand how this would help. Can you please elaborate?

```svelte
<!-- Form.svelte -->
<form on:keydown>...</form>
```

weaseldotro:
That's what I tried at first but it didn't work. Now I see that the `Form` component in the library is not forwarding the `keydown` event :( I replaced the `Form` in my code with a regular `form` HTML element and it now works.
Perhaps there should be a `keydown` event in the `Form` component?

Anyway, the whole point of this PR is now moot, I'll close it.

metonym:
> Perhaps there should be a keydown event in the Form component?

Yes, that's what I was suggesting

weaseldotro:
Should I make a PR for this?

metonym:
Yes, I'd accept a PR for forwarding the keydown event to the `Form` component

PR #568, opened 2021-03-18
NumberInput: forward name to non-mobile variant

jhwz:
NumberInput does not set `name` property correctly on the non-mobile variant.


metonym:
The mobile variant of the number input is deprecated.

jhwz:
I understand that, apologies if I wasn't clear enough. If you look at the changes you will understand but currently the `name` property is only forwarded to the (deprecated) mobile variant and not the normal version of number input

Issue #597, opened 2021-04-14
Sortable DataTable, sorting does not work if any item is undefined


alkanna:
```
TypeError: can't access property "toString", itemB is undefined
```

Shouldn't undefined items be placed either at the top or the end of the sorted array ?
This forces us to manipulate data that should be immutable.

Error is in `DataTable.svelte:182`

alkanna:
Proposed fix :
```javascript
          return itemA 
                  ? itemA.toString().localeCompare(itemB 
                          ? itemB.toString() 
                          : '', 'en', {numeric: true}) 
                  : '';
```

alkanna:
I made a pull request #602 that fixes this issue and adds the ability to bind nested objects/arrays.

Issue #564, opened 2021-03-11
Allow nested values for DataTable header key


Trystan-SA:
If I have data ordered like this : 

```js 
[{ 
    id: "123456"
    identity: {
       firstName: "John",
       lastName: "Doe"
    }
 }]
```

It's impossible to select "lastName" as a key

```js

<DataTable
   headers={[
      {value:"Last Name"  key: "identity.lastName"} ???
   ]}
/>
```

The values inside the DataTable will be undefined.

Trystan-SA:
If I use the **display** and **sort** parameters, I can make it run almost correctly.

```js 
<DataTable
   headers={[
      {value:"Last Name", key: "identity", 
      display: (identity)=> identity.lastName, 
      sort:(a,b)=>{ 
         if(a.lastName > b.lastName){return 1}
         else if(a.lastName < b.lastName{ return -1}
      }
   }]
/>
```

But if I want to have the same thing with the firstName field, a glitch will appear. This will mix the order of the column when sorting them (probably because they have the same key as reference) 

- When I sort any column for the first time (the result is the same for all of them) . the lastName column will get the values of the firstName column.
- When I sort it a second time. The batchExpansion column will get the value of the firstName column. The firstName column will get the value of the batchExpansion column (an arrow on every field) and the lastName column stay with the values of the firstName column.
- Nothing changes if I do it again.


So I don't think there is another option for me (apart from changing the value of the array directly)

Trystan-SA:
I found a way to fix my issue.

I can flatten my JS object (by using the [JS flat package](https://www.npmjs.com/package/flat) or doing it myself)
It will then work as expected.

The issue itself is not solved, but this is a nice workaround.

alkanna:
Hey @Trystan-SA, a PR was just merged that fixes this issue : #602

Trystan-SA:
Great work ! Thank you for this PR :)

PR #602, opened 2021-04-16
Add support for nested object fields in DataTable

alkanna:
I took the liberty to provide a way to access nested objects by a string path.
I also changed the way undefined items are handled : instead of showing up as "undefined" in the DataTable, the cells will simply be empty, which makes it so that it does not crash with the default sorting behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- ## Unreleased -->
 
+## [0.32.2](https://github.com/IBM/carbon-components-svelte/releases/tag/v0.32.2) - 2021-04-23
+
+**Fixes**
+
+- prevent multiple Dropdown items from being selected by updating the `selectedId`
+- prevent the default keydown behavior so that pressing "Enter" should open the Dropdown menu
+- resolve Carbon icon imports to Svelte files in `DataTable`, `HamburgerMenu` components
+
+**Housekeeping**
+
+- patch `carbon-components` to version 10.32.1
+
 ## [0.32.1](https://github.com/IBM/carbon-components-svelte/releases/tag/v0.32.1) - 2021-04-02
 
 **Fixes**
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -1,6 +1,6 @@
 # Component Index
 
-> 167 components exported from carbon-components-svelte@0.32.1.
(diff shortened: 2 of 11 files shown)


alkanna:
@metonym Hey, could you take a look at this ?

metonym:
Thanks for taking the time to put this PR together. Referencing #597, I agree that an undefined value should default to an empty string.

For the part about nested objects/arrays, could you provide a simple example?

alkanna:
Hey, it's pretty straightforward.
Let's say we have this array of objects :
```javascript
[{
    id: 'a',
    name: 'Load Balancer 3',
    network: {
		protocol: "HTTP",
		port: 3000
	},
    rule: 'Round robin'
}, {
    id: 'b',
    name: 'Load Balancer 1',
    network: {
		protocol: "HTTP",
		port: 443
	},
    rule: 'Round robin'
}, {
    id: 'c',
    name: 'Load Balancer 2',
    network: {
		protocol: "HTTP",
		port: 80
	},
    rule: 'DNS delegation'
}]
```

Before, we could not access `network.protocol` in a `DataTable`, with this little PR, we can do so like this :

```html
<DataTable
  headers={[{ key: 'name', value: 'Name' }, { key: 'network.protocol', value: 'Protocol' }, { key: 'network.port', value: 'Port' }, { key: 'rule', value: 'Rule' }]}
  rows={}
/>
```

This would also work if `network` was an array (also works with indexes) and not an object :

```javascript
data = [{
    id: 'a',
    name: 'Load Balancer 3',
    network: ["HTTP", 3000],
    rule: 'Round robin'
}, {
    id: 'b',
    name: 'Load Balancer 1',
    network: ["HTTP", 443],
    rule: 'Round robin'
}];

console.log(resolvePath(data[0], "network.1", ""));
```

Output : 
```shell
$ node index.js 
3000
```

alkanna:
If there is anything missing don't hesitate to point it out, like tests or documentation.

metonym:
@alkanna I like that example. Could you add an example "Sortable with nested object values" to the doc site after the [Sortable with custom display and sort methods](https://github.com/IBM/carbon-components-svelte/blob/master/docs/src/pages/components/DataTable.svx#L612)?

alkanna:
Hey @metonym, I added the example as requested.

metonym (approved):
Great work – excited to use this feature.

Also, this should close related issue #564

metonym:
Published to NPM in [version 0.33.0](https://github.com/IBM/carbon-components-svelte/releases/tag/v0.33.0)

Issue #607, opened 2021-04-21
can't open a dropdown with the keyboard


jasonlyu123:
## Step to reproduce

It can be reproduced on the docs site. 
1. Navigate to any dropdown example
2. Hold the tab key until it focuses on a dropdown
3. Press the enter key to activate the dropdown

It would open and be closed immediately. I think it's because the button also bubbles a click event so that it got toggled twice. One in click handler and another in a keydown handler. Seems like a preventDefault can fix this. I can open a PR for this if you like to.

metonym:
Thank you for raising this issue – PR would be appreciated.

PR #610, opened 2021-04-22
fix dropdown keyboard usage

jasonlyu123:
Fixes can't open a dropdown with keyboard #607 
I didn't use the preventDefault modifier because it would also prevent the default of the tab. Which interferes with the tab navigation. 

While on this I found another inconsistency with other implementations. The selected options of a dropdown can be switch by arrow down and arrow up when the dropdown isn't opened. Other implementations like the vue and react version would open the Dropdown and switch highlighted options. Should I open another PR for it? Or I can just do it here.

### DIFF
--- a/src/Dropdown/Dropdown.svelte
+++ b/src/Dropdown/Dropdown.svelte
@@ -203,7 +203,11 @@
       tabindex="0"
       role="button"
       aria-expanded="{open}"
-      on:keydown="{({ key }) => {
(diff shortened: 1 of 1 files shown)


metonym (approved):
Thank you for fixing this.

metonym:
Published to NPM in [version v0.32.2](https://github.com/IBM/carbon-components-svelte/releases/tag/v0.32.2)

PR #618, opened 2021-04-30
Fix TableCell "colspan" when a DataTable is Expandable and Selectable

mabentley85:
Small fix to correct the width of the `expanded-row` slot when a DataTable is both expandable and selectable.

The `colspan` of TableCell that wraps the `expanded-row` slot is now set to `header.length + 2` instead of `header.length + 1` to accommodate the additional selectable column.

### DIFF
--- a/src/DataTable/DataTable.svelte
+++ b/src/DataTable/DataTable.svelte
@@ -410,7 +410,7 @@
               parentRowId = null;
             }}"
           >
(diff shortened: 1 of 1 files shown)


metonym (approved):
Nice one

metonym:
Published to NPM in [version 0.33.0](https://github.com/IBM/carbon-components-svelte/releases/tag/v0.33.0)

PR #647, opened 2021-05-14
Add html input attributes to ToolbarSearchProps

### LINKED ISSUE #646: $$restProps Types
It would be great to have type definitions for `$$restProps`. For example, `placeholder` is not defined on `ToolbarSearch`. 

`svelte-check` states the following problem: 

```
Type '{ placeholder: string; value: string; persistent: true; }' is not assignable to type 'IntrinsicAttributes & ToolbarSearchProps'.
  Property 'placeholder' does not exist on type 'IntrinsicAttributes & ToolbarSearchProps'.
```

gregorw:
Fixes #646

ToolbarSearchProps are missing type definitions for $$restProops. With this PR `svelte-check` no longer complains and editors can show propose code completions. E.g. VS Code:

image

*Limitations*

This is just a very specific case. The proper definition of `$$restProps` in other components is subject to further work.
As an alternative to `svelte.JSX.HTMLAttributes` I would like to suggest [`Partial`](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype) in a separate PR.

### DIFF
--- a/docs/src/COMPONENT_API.json
+++ b/docs/src/COMPONENT_API.json
@@ -11016,7 +11016,7 @@
         { "type": "forwarded", "name": "blur", "element": "Search" }
       ],
       "typedefs": [],
-      "rest_props": { "type": "InlineComponent", "name": "Search" }
+      "rest_props": { "type": "Element", "name": "input" }
     },
     {
       "moduleName": "Tooltip",
--- a/src/DataTable/ToolbarSearch.svelte
+++ b/src/DataTable/ToolbarSearch.svelte
@@ -1,4 +1,6 @@
 <script>
+  /** @restProps {input} */
+
   /**
    * Specify the value of the search input
    * @type {number | string}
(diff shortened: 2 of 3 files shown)


metonym (changes requested):
Thanks for catching this.

The TypeScript definitions are auto-generated and should not be edited directly.

I'd suggest adding `@restProps {input}` as a comment to `ToolbarSearch.svelte`. 

Here's an example: https://github.com/IBM/carbon-components-svelte/blob/master/src/Button/Button.svelte#L4

gregorw:
Thanks for taking the time to respond. I have updated the PR as you suggested. So, it’s analogous to:

https://github.com/IBM/carbon-components-svelte/blob/cff16e5d7de19652df6a905360bbfe01ea6d0d7f/src/Tag/Tag.svelte#L2

Great tooling by the way. I’m Impressed by sveld.

metonym (approved):
Great work – thank you for your patience.

metonym:
Available in v0.34.0 ([release notes](https://github.com/IBM/carbon-components-svelte/releases/tag/v0.34.0))

Issue #567, opened 2021-03-18
Tab focus isn't trapped in Modal/ComposedModal


dqzx:
Both the Modal and ComposedModal components allow a user to transfer focus to other page elements via the keyboard. The focus should be trapped within the modal, as per https://www.carbondesignsystem.com/components/modal/usage#focus.

metonym:
Fixed in v0.38.2

PR #716, opened 2021-07-02
Fix: trap tab focus within modal

dqzx:
Fix for issue #567


metonym (changes requested):
@dqzx Great work! Thanks for taking this one.

Could you apply the same trap focus behavior to `ComposedModal`?

metonym (approved):

PR #733, opened 2021-07-07
Breakpoint

metonym:
**Features**

- add `Breakpoint` component

---

The [Carbon Design System grid implementation](https://carbondesignsystem.com/guidelines/2x-grid/implementation#responsive-options) defines five responsive breakpoints:

- Small (`sm`): less than 672px
- Medium (`md`): 672 - 1056px
- Large (`lg`): 1056 - 1312px
- X-Large (`xlg`): 1312 - 1584px
- Max (`max`): greater than 1584px

This utility component uses the [Window.matchMedia API](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) to declaratively determine the current Carbon breakpoint size.

### Usage

```svelte
<script>
  import { Breakpoint } from "carbon-components-svelte";

  let size;
  let events = [];
</script>

<Breakpoint bind:size on:match="{(e) => (events = [...events, e.detail])}" />

<h1>{size}</h1>

<pre>
  {JSON.stringify(events,null, 2)}
</pre>

```

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -1,6 +1,6 @@
 # Component Index
 
-> 170 components exported from carbon-components-svelte@0.39.0.
+> 171 components exported from carbon-components-svelte@0.39.0.
 
 ## Components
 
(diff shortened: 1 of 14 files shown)


keehun:
@metonym, I can't seem to import `Breakpoint`. It shows error: `Module '"carbon-components-svelte"' has no exported member 'Breakpoint'. ts(2305)`

I am importing it along with other carbon components. Should it be imported from a different package?

metonym:
@keehun How are you importing it?

See the [REPL](https://svelte.dev/repl/6545ff1a1aef44de97fea024ca9d9573) for an example.

keehun:
The exact same way as the REPL. Still getting the same problem. In `node_modules`, I can see the source for `Breakpoint` and type exports for it and such. Unsure where it's going wrong. All the other components I can use just fine... Will have to do more digging. Might have to install a fresh version of SvelteKit+CDS and see if the issue persists.

PR #745, opened 2021-07-13
fix(DataTable): undefined cell value fallback

truongminh:
Zero cell value should be as-is.
Only fallback on `undefined` property.

Link to issue:
- https://github.com/IBM/carbon-components-svelte/issues/666

### DIFF
--- a/src/DataTable/DataTable.svelte
+++ b/src/DataTable/DataTable.svelte
@@ -127,14 +127,11 @@
       .map(({ key }, i) => ({ key, id: $headerItems[i] }))
       .reduce((a, c) => ({ ...a, [c.key]: c.id }), {})
   );
-  const resolvePath = (object, path, defaultValue) =>
+  const resolvePath = (object, path) =>
     path
       .split(/[\.\[\]\'\"]/)
       .filter((p) => p)
-      .reduce(
-        (o, p) => ((o && typeof o === "object") ? o[p] : defaultValue),
-        object
-      );
+      .reduce((o, p) => (o && typeof o === "object" ? o[p] : o), object);
 
   setContext("DataTable", {
     sortHeader,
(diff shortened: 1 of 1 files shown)


metonym (changes requested):
Thanks for taking this on.

It appears that the site is not rendering cell values: https://carbon-components-svelte-git-fork-truongmi-04150b-carbon-svelte.vercel.app/components/DataTable

See the suggested change.

src/DataTable/DataTable.svelte
```diff
@@ -132,7 +132,7 @@
       .split(/[\.\[\]\'\"]/)
       .filter((p) => p)
       .reduce(
-        (o, p) => (o && typeof o === "object" && o[p] ? o[p] : defaultValue),
+        (o, p) => ((o && typeof o === "object" && typeof o[p] === 'undefined') ? o[p] : defaultValue),
```
metonym, line 135: Allow whatever value `o` is to render as is if it's not an object.

```suggestion
        (o, p) => ((o && typeof o === "object") ? o[p] : o),
```

truongminh: Code updated. I just found `cell.display` useful also.
Thank @metonym for the change.

src/DataTable/DataTable.svelte
```diff
@@ -132,7 +132,7 @@
       .split(/[\.\[\]\'\"]/)
       .filter((p) => p)
       .reduce(
-        (o, p) => (o && typeof o === "object" && o[p] ? o[p] : defaultValue),
+        (o, p) => ((o && typeof o === "object") ? o[p] : defaultValue),
```
metonym, line 135: I'm not sure if this solves the issue. Having it fallback to the defaultValue would result in the empty string.

metonym (changes requested):
See my comment.

Issue #315, opened 2020-10-08
App switcher in UI Shell doesn't always close when clicking outside it


weaseldotro:
Go to this demo: https://carbon-svelte.vercel.app/framed/UIShell/HeaderSwitcher
Open the app switcher in the UI Shell (top right button).
Click in the main window, anywhere between a few centimeters below the welcome text to the bottom of the page. So basically anywhere below `<main>`

This can be fixed by changing `<svelte:body` to `<svelte:window` in this line:
https://github.com/IBM/carbon-components-svelte/blob/master/src/UIShell/GlobalHeader/HeaderAction.svelte#L54

The app switcher is now closing, however I don't know if this has other consequences.

metonym:
Open to PRs

weaseldotro:
It seems that `<svelte:body>` is limited to the page body inside de browser window, while `<svelte:window>` uses the entire window. So if the page contents is smaller than the window, the rest of the window will not trigger the events defined on `<svelte:body>`

It looks like the same problem happens with the ComboBox component. So I did a search and the `<svelte:body>` is used in a bunch of components:
- ComboBox/ComboBox.svelte
- DatePicker/DatePicker.svelte
- Dropdown/Dropdown.svelte
- MultiSelect/MultiSelect.svelte
- OverflowMenu/OverflowMenu.svelte
- Slider/Slider.svelte
- Tooltip/Tooltip.svelte
- TooltipDefinition/TooltipDefinition.svelte
- TooltipIcon/TooltipIcon.svelte

Most of them use the click event, TooltipDefinition and TooltipIcon use keydown, while Slider uses a few others.
Should I make a PR and replace all the instances of `<svelte:body>` with `<svelte:window>`?

metonym:
Start with the UI Shell header switcher

phil-mayer:
@metonym I can still see this behavior in `MultiSelect/MultiSelect.svelte`. The MultiSelect control does not close as expected if the body is shorter than the window and the user clicks an area between the bottom of the body and the bottom of the window.

Can I open a PR to switch out `<svelte:body>` for `<svelte:window>` on this component?

metonym:
@phil-mayer Yes

weaseldotro:
Might as well replace all the instances of `<svelte:body>` that I listed in my last message :)

phil-mayer:
Sure, happy to do so as long as @metonym is ok with the increase in scope.

metonym:
[...]

PR #761, opened 2021-07-22
fix: close overlays on window click

phil-mayer:
Replace all occurrences of `svelte:body` with `svelte:window` to resolve the issue described in #315.

All occurrences listed in [this comment](https://github.com/IBM/carbon-components-svelte/issues/315#issuecomment-705669503) were replaced, except for `Dropdown/Dropdown.svelte`, as it was already replaced.


metonym (approved):
Thanks for taking care of this!

PR #785, opened 2021-08-11
Set a condition for showing legend in FormGroup

ibilux:
We shouldn't add a `<legend>` If the the legend text specified.


metonym (approved):
Good catch – thanks for taking care of this.

Issue #753, opened 2021-07-16
DatePicker Popup is positioned wrong when inside dialog. 


mikerussellnz:
When a datepicker is placed inside a modal / composed dialog the picker part is positioned incorrectly. 

Screenshot below, it seems that the calculation of the position of the picker is being calculated relative to the top left of the screen, but it is being positioned relative to the top left of the dialog. :-(

![datepicker](https://user-images.githubusercontent.com/627852/125935327-507e3687-b881-4636-b661-aa31531e5078.png)

mikerussellnz:
I found that the fix is to set static: true on the calendar config options passed to flatpickr. 
` calendar = await createCalendar({
      options: {
        appendTo: datePickerRef,
        dateFormat,
        defaultDate: $inputValue,
        locale,
        maxDate,
        minDate,
        mode: $mode,
	static: true  // <<< This line
      },`

Result:
![datepicker2](https://user-images.githubusercontent.com/627852/125937074-4450ba8b-3f01-46a1-889e-3a02096245cc.png)

insasho:
> I found that the fix is to set static: true on the calendar config options passed to flatpickr.

@mikerussellnz I am also experiencing this problem. Can you share your workaround showing how to pass static: true through to the flatpickr instance used by carbon-components-svelte?

metonym:
Supported in [v0.44.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.44.0).

Override/customize the `flatpickr` calendar options using `flatpickrProps`.

```svelte
<script>
  import { DatePicker, DatePickerInput } from "carbon-components-svelte";
</script>

<DatePicker
  datePickerType="single"
  flatpickrProps="{{ static: true }}"
  on:change
>
  <DatePickerInput labelText="Meeting date" placeholder="mm/dd/yyyy" />
</DatePicker>
```

See https://flatpickr.js.org/options for a full list of options.

PR #794, opened 2021-08-22
added the staticProp property for the DatePicker component to work correctly

stilet:
added the staticProp property for the Modal component to work correctly #753

### DIFF
--- a/src/DatePicker/DatePicker.svelte
+++ b/src/DatePicker/DatePicker.svelte
@@ -57,7 +57,7 @@
   export let id = "ccs-" + Math.random().toString(36);
 
   /** Set to `true` to use in component Modal */
-  export let staticProp = false;
+  export let flatpickrProps = {};
 
   import {
     createEventDispatcher,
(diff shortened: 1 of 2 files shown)


src/DatePicker/DatePicker.svelte
```diff
@@ -56,6 +56,9 @@
   /** Set an id for the date picker element */
   export let id = "ccs-" + Math.random().toString(36);
 
+  /** Set to `true` to use in component Modal */
+  export let staticProp = false;
```
metonym, line 60:
```suggestion
  export let flatpickrProps = {};
```

src/DatePicker/DatePicker.svelte
```diff
@@ -138,6 +141,7 @@
         maxDate,
         minDate,
         mode: $mode,
+        static: staticProp,
```
metonym, line 144:
```suggestion
        ...flatpickrProps
```

metonym (changes requested):
This is a great initiative. I think it would be better just to be able to pass any flatpickr prop to the calendar. See my comments.

metonym (approved):

PR #816, opened 2021-09-13
feat: improve typing for DataTable row and cell

### LINKED ISSUE #667: DataTable runtime row type is lost
The DataTable component uses a spread operation internally which loses all non-enumerated properties and the original type:

```js
$: rows = rows.map((row) => ({
  ...row,
  cells: headerKeys.map((key, index) => ({
    key,
    value: resolvePath(row, key, ""),
    display: headers[index].display,
  })),
}));
```

E.g. functions of the prototype will be gone:

```js
var x = { log() { console.log('!') } }
var y = { ...x }
y.log();
// logs !

class X { log() { console.log('!') } }
var x = new X();
var y = { ...x }
y.log()
// TypeError: y.log is not a function
```

I would recommend assigning the original row to a property instead of or additionally to spreading it to prevent this.

In my particular use case the row is a `Proxy`, which becomes mostly useless after the spread. Having the original type might also be useful for type-based operations using `instanceof`.

KoichiKiyokawa:
fix #667 (maybe)

Thank you for such a nice library :tada:
When using DataTable component, it would be great if the editor autocompletes `headers` prop or slot values(`cell.key` or `row`).

I make it possible with [Generics](https://www.typescriptlang.org/docs/handbook/2/generics.html)(named `Row`). For example,
```ts
export interface DataTableEmptyHeader<Row extends DataTableRow = DataTableRow>
```
`extends DataTableRow` restrict Row type (e.g. it should have `id` property). `= DataTableRow` is to avoid breaking change.

Feel free to:
- Close this PR if not suitable.
- Change my code.

without say anything.

## demo (autocomplete is working correctly by this PR)

https://user-images.githubusercontent.com/40315079/133122454-829fbe75-b052-44f1-8230-f3eca94640cc.mp4

<details>
<summary>code in demo</summary>

```svelte
<script>

  const rows = [
    { id: 1, name: "user1", age: 10, gender: "male" },
    { id: 2, name: "user2", age: 20, gender: "female" },
  ];
</script>

<DataTable
  rows="{rows}"
  headers="{[{ key: 'name', value: 'Name' }, { key: 'age', value: 'Age' }, { key: 'gender', value: 'Gender' }]}"
>
  <div slot="cell" let:cell let:row>
    {#if cell.key === 'age'}{row.age}(years old){:else}{cell.value}{/if}
  </div>
</DataTable>

```
</details>

### DIFF
--- a/types/DataTable/DataTable.svelte.d.ts
+++ b/types/DataTable/DataTable.svelte.d.ts
@@ -5,8 +5,8 @@ export type DataTableKey = string;
 
 export type DataTableValue = any;
 
-export interface DataTableEmptyHeader {
-  key: DataTableKey;
+export interface DataTableEmptyHeader<Row extends DataTableRow = DataTableRow> {
+  key: keyof Row;
   empty: boolean;
   display?: (item: Value, row: DataTableRow) => DataTableValue;
   sort?: false | ((a: DataTableValue, b: DataTableValue) => number);
@@ -15,8 +15,10 @@ export interface DataTableEmptyHeader {
   minWidth?: string;
 }
 
-export interface DataTableNonEmptyHeader {
-  key: DataTableKey;
+export interface DataTableNonEmptyHeader<
+  Row extends DataTableRow = DataTableRow
+> {
+  key: keyof Row;
   value: DataTableValue;
   display?: (item: Value, row: DataTableRow) => DataTableValue;
   sort?: false | ((a: DataTableValue, b: DataTableValue) => number);
@@ -25,7 +27,9 @@ export interface DataTableNonEmptyHeader {
   minWidth?: string;
 }
 
-export type DataTableHeader = DataTableNonEmptyHeader | DataTableEmptyHeader;
+export type DataTableHeader<Row extends DataTableRow = DataTableRow> =
+  | DataTableNonEmptyHeader<Row>
+  | DataTableEmptyHeader<Row>;
 
 export interface DataTableRow {
   id: any;
@@ -34,8 +38,8 @@ export interface DataTableRow {
 
 export type DataTableRowId = any;
 
-export interface DataTableCell {
-  key: DataTableKey;
+export interface DataTableCell<Row extends DataTableRow = DataTableRow> {
+  key: keyof Row;
   value: DataTableValue;
   display?: (item: Value, row: DataTableRow) => DataTableValue;
 }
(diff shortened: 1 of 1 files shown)


metonym:
Started testing locally – this looks very promising! 

This may also address this issue: https://github.com/carbon-design-system/carbon-components-svelte/issues/364

brunnerh:
There is a bit of wrinkle with `key: keyof Row;` in that keys can also access nested objects or not exist at all. But I would love to have generics for the slot properties, otherwise it requires a lot of type assertions when working in TypeScript.

[REPL demonstrating other key uses](https://svelte.dev/repl/aab4495dd1d148238e3307b5566ff9e5?version=3.47.0)

Using interpolated string literal types it it actually possible to get nested keys as well. The types may be a bit hard to read and it does not handle the synthetic keys case, though.

[TS Playground example for nested key resolution](https://www.typescriptlang.org/play?#code/C4TwDgpgBACgThAblAvFA2gOyROAaKABgIEYCAmAgZgIBYCBWAgNgIHYCAOAgTlOICwAKCiioJMuMria4+uKbjW4juO7i+UcsSgA6fYXQBdI8OGhIUAFIB7AJaYAPAGkCMAHyoozqBAAewBCYACYAzlChwHAOAOZQAD5QmACuALYARrhQAPzCYrC+AUFhEVGxCUlpmXA5eWIABgAkAN7OAL4tAOSdhYEh4TA5UN1QAFzDup0dzTBt9XWi49iIWUs4cADcZkIW0DAAhsAAFqGOACoEACK9xeEpGVloJISeaOiXRjf9GMu4n9lJdZjKBnL4lGzpABWEAAxsBaiIxM0MD4HFAANYQEA2ABmIKMAFpsuMfP4+iVItFMHFEvdqgj8vkmq05hVbA4XG5DidzuhnEY3AhEO8jO53AtGWsVnAJW10JjsXizp9xt0tkIJcIYTZMJEoHAbAB3cJvCXNCX5OzBcaCRGMpL7VIQcbInF2OCRVW2I6YToEAA2+09w0uNggPTaeAtYn2MWdUFolFlUbtUHNqct1vEKftokwjvjrvdwc6Vn22D9UEDJdD4agkejolj8aoSdTDaERnV5nA0AASkavLtcfqjaEsFU-t2dr2oAPDc4sV4DsdTvP3OqAPSbsQAPWyQA)

PR #824, opened 2021-09-21
fix(DataTable): useStaticWidth fix for title/description

### LINKED ISSUE #795: DataTable useStaticWidth title/description style bug
image
will try to do a PR
edit: seems like this is an upstream issue and it will probably not get merged here if I "simply" fix it

Florian-Schoenherr:
Closes #795 
- does the same thing as upstream, will work once the version of carbon is bumped
- docs updated, the diff is just bad (I just added title&description to the useStaticWidth example)


metonym (approved):
Thank you @Florian-Schoenherr for opening the initial issue and following through to fix this upstream and in this implementation.

I've upgraded `carbon-components` to v10.44.0 and have verified that `useStaticWidth` now works correctly provided a title/description.

Note: we also have to upgrade `sass` to version >= 1.33 as `carbon-components@10.44` uses the new [`math.div` API](https://github.com/carbon-design-system/carbon/pull/9508).

PR #836, opened 2021-10-04
fix: menuOptionsClass in OverflowMenu

KoichiKiyokawa:
Thanks.
I think that the `menuOptionsClass` prop (in OverflowMenu component) is not work correctly.

```svelte
<OverflowMenu menuOptionsClass="foo-class" open>
  <OverflowMenuItem text="foo" />
  <OverflowMenuItem text="bar" />
</OverflowMenu>
```

## before
image

## after(this PR)
image


metonym (approved):
Thanks for catching this!

metonym:
Released in [v0.44.5](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.44.5)

Issue #839, opened 2021-10-08
Multiple <HeaderAction /> problem


miedzikd:
When you open HeaderAction, and you click anywhere then sidebar should be hidden.
But... when you have more then one <HeaderAction /> in <HeaderUtilities> and you first open second HeaderAction, then first HeaderAction you will stuck on second one... 

Steps to reproduction:
1. Open REPL: https://svelte.dev/repl/300f4d72b0854ed3a7e33442c6baca0a?version=3.35.0
1. Open second Header action, and don't click anywhere.
1. Open first Header action.
1. You should still see second Header, first one is below second and is still open.

metonym:
Related #246

PR #840, opened 2021-10-08
Add closeIcon option and remove stopPropagation on:click in HeaderAction

miedzikd:
Remove stopPropagation on:click helps when you have more then one acction to switch between actions tabs

Releated issue: #839


miedzikd:
@metonym sure I can do this... 
but did you consider to add maybe just an attirubute for `HeaderAction` component? Something like 
- `notChangeIconWhenOpen` or
- `openIcon` (with option to use custom Icon - default close) or
- `closeIcon` (same as above) ?

Naming and English generally is not my strong side :P

Thanks for that we can avoid braking changes, but have this new functionality

miedzikd:
Right now we have `icon` attribute already... To have back compatibility we can leave this, and add only one new `closeIcon` attribute... Do you agree?

PR #879, opened 2021-10-26
document static property on datepicker modal

cupcakearmy:


cupcakearmy:
@metonym sry forgot about the pr.

metonym:
Thanks for taking the time – I added some missing props to the Modal/DatePicker.

PR #880, opened 2021-10-27
feat: allow paginated sorting data table

kwieszalka-maystreet:
This MR is introducing  `pageSize` and `page` props, allowing  `DataTable` to sort the whole dataset but still display paginated content. It can be also used with non-Carbon pagination components.

How it works:
- pass the whole dataset in the `rows` prop, 
- limit displayed content by using `pageSize` and `page` props  (which are corresponding to the same props in the `Pagination` component).

### DIFF
--- a/src/DataTable/DataTable.svelte
+++ b/src/DataTable/DataTable.svelte
@@ -100,6 +100,12 @@
   /** Set to `true` to use static width */
   export let useStaticWidth = false;
 
+  /** Set to `number` to limit the output */
+  export let pageSize = 0;
+
+  /** Set to `number` to set current page */
+  export let page = 0;
+
   import { createEventDispatcher, setContext } from "svelte";
   import { writable, derived } from "svelte/store";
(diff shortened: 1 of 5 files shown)


kwieszalka-maystreet:
@metonym

src/DataTable/DataTable.svelte
```diff
@@ -100,6 +100,12 @@
   /** Set to `true` to use static width */
   export let useStaticWidth = false;
 
+  /** Set to `number` to limit the output */
```
metonym, line 103: Let's make the `pageSize` description consistent with that from `Pagination.svelte`.

```suggestion
  /** Specify the number of items to display in a page */
```

metonym (commented):
This is fantastic work. Looking at the example you added, the API is beautifully concise.

This approach is better than maintaining a separate `PaginationDataTable` component.

metonym:
Nice. Could you run "yarn build:lib" one more time so the `COMPONENT_INDEX.md` and TypeScript definitions are updated?

metonym (approved):

Issue #872, opened 2021-10-21
[Regression] Reactive tooltip example doesn't work


metonym:
https://carbon-components-svelte.onrender.com/components/Tooltip#reactive-example

Clicking the button to toggle the tooltip cannot close it.

malinowskip:
This turned out to be really tricky, but I came up with a possible solution: https://github.com/malinowskip/carbon-components-svelte/commit/257cf816dce8909690a98a108877645d141fb6f7. All the examples seem to work.

metonym:
@malinowskip Could you submit a PR? Would love to give it a try

metonym:
Released in [v0.47.5](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.47.5)

Issue #869, opened 2021-10-21
Tooltip is not closing when click on tooltip


miedzikd:
When you click on tooltip trigger, next on tooltip... then event for closing tooltip stop working

Steps to reproduction:
1. Open: https://svelte.dev/repl/57e64a2a5378433099879637ebac9539?version=3.35.0
1. Click on TEST1
1. Click on TOOLTIP1
1. Click on TEST2

You will see that tooltip1 and tooltip2 are both open

PR #885, opened 2021-11-01
fix(ToolTip): better toggling logic

malinowskip:
Proposal to fix https://github.com/carbon-design-system/carbon-components-svelte/issues/872

### DIFF
--- a/src/Tooltip/Tooltip.svelte
+++ b/src/Tooltip/Tooltip.svelte
@@ -89,8 +89,17 @@
     }
   }
 
-  function openMenu() {
+  function onFocus() {
     open = true;
+  }
+
+  function onMousedown() {
+    // determine the desired state before the focus event triggers.
+    const shouldClose = open;
+    // ensure changes are scheduled at the end, i.e. after the possible focus event.
+    setTimeout(() => {
+      open = shouldClose ? false : true;
+    });
   }
 
   afterUpdate(() => {
@@ -174,7 +183,7 @@
         }
       }
 
-      open = false;
+      if (!ref.contains(target)) open = false;
     }
   }}"
 />
@@ -190,8 +199,8 @@
         bind:this="{refIcon}"
         {...buttonProps}
         aria-describedby="{tooltipId}"
-        on:click|preventDefault|stopPropagation="{openMenu}"
-        on:focus="{openMenu}"
+        on:mousedown="{onMousedown}"
+        on:focus="{onFocus}"
         on:blur="{onBlur}"
         on:keydown="{onKeydown}"
       >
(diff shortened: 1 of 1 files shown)


metonym:
Thank you for taking the time. I'm still encountering the following behavior.

Clicking the button once will double trigger the tooltip – it will open and close it in fast succession.

https://user-images.githubusercontent.com/10718366/139747955-eceb0c38-6430-4205-a0bf-eeb572f94dfb.mov

malinowskip:
Thanks for reviewing. How can I open that example? I don’t know what the route should be.

metonym:
> Thanks for reviewing. How can I open that example? I don’t know what the route should be.

The route should be `http://localhost:3000/framed/Tooltip/TooltipReactive`

I tested it with the previous code (before the [commit](https://github.com/malinowskip/carbon-components-svelte/commit/f155486c373d6b0ddedd89284a8e02b34ff98458) that modified the example):

```svelte
<!-- TooltipReactive.svelte -->
<script>
  import { Tooltip, Button } from "carbon-components-svelte";
  let open = true;
</script>

<style>
  div {
    margin-top: var(--cds-spacing-05);
  }
</style>

<Tooltip bind:open triggerText="Resource list" align="start">
  <p>Resources are provisioned based on your account's organization.</p>
</Tooltip>

<div style="margin-top: var(--cds-spacing-12);">
  <Button size="small" on:click="{() => (open = !open)}">
    {open ? 'Close tooltip' : 'Open tooltip'}
  </Button>
</div>

<div>Open: {open}</div>

```

malinowskip:
Okay, I think I get it now. The "Open tooltip" button opens the tooltip, but it still triggers this mousedown handler in the ToolTip component:

```svelte
<svelte:window
  on:mousedown="{({ target }) => {
    if (open) {
      if (target.contains(refTooltip)) {
        if (refIcon) {
          refIcon.focus();
        } else if (ref) {
          ref.focus();
        }
      }

      if (!ref.contains(target)) open = false;
    }
  }}"
/>
```

metonym:
Unfortunately, it reintroduces this issue where multiple tooltips can be open at the same time: https://github.com/carbon-design-system/carbon-components-svelte/issues/869

malinowskip:
Thank you for reviewing. I tried a bunch of different things, but it seems like each fix introduces a regression one way or another – in most cases due to event handlers firing at different times, doing conflicting things. For example, in the reactive example, the `mousedown` event on the button closes the tooltip, so the “Close tooltip” button becomes an “Open tooltip” button mid-click.

The only thing I can think of at this point is to use a store to ensure that only one ToolTip is open at any given time, but that might be overkill if there’s a simpler solution.

malinowskip:
I have tested this with all the examples in the docs, the reactive example, as well as the “multiple tooltips” example, in Firefox and Chrome. To deal with the race conditions, I registered the new click event in the capture phase, and used a timeout. @metonym when you have a chance, can you have a look? I hope I didn’t miss anything obvious this time.

There is an unrelated minor issue where if you open a tooltip, click on it, and then tab out, then it remains open.

metonym (approved):
Well done!

Issue #894, opened 2021-11-08
Tooltip: reactive toggling doesn’t work if the tooltip icon is in focus


malinowskip:
Unfortunately, I didn’t cover one possibility in https://github.com/carbon-design-system/carbon-components-svelte/pull/885. In the [reactive tooltip example](https://carbon-components-svelte.onrender.com/components/Tooltip/#reactive-example), toggling doesn’t work as intended if the icon is in focus.

https://user-images.githubusercontent.com/47485431/140752758-3306bdf4-d50a-4122-ae28-2401d46ce4dc.mov

metonym:
Patched in [v0.47.6](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.47.6)

PR #895, opened 2021-11-08
fix(Tooltip): fix two way binding in cases where icon is in focus

malinowskip:
Proposal to fix https://github.com/carbon-design-system/carbon-components-svelte/issues/894


metonym (approved):
Nice catch

PR #900, opened 2021-11-10
Add ref prop to `Form`

### LINKED ISSUE #877: Add ref prop to Form?
A lot of components have a `ref` prop to get the underlying HTML element, but `Form` does not.
Getting the element would be helpful for programmatic submission or validation of the form.

metonym:
Closes #877

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -1447,7 +1447,9 @@ None.
 
 ### Props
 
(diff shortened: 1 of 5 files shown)


metonym:
Released in [v0.48.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.48.0)

PR #901, opened 2021-11-10
Add autoWidth prop to `Tabs`

### LINKED ISSUE #899: Allow Tabs to have an auto width
Related #774

Currently, the width of `Tab` is hardcoded to be `10rem`. I often find myself manually overriding the CSS so that tabs with longer text have an automatic width.

Because `carbon-components` currently does not support it, we can inline the rule using the `style` prop. In the flagship repo, Tabs will have an auto width in v11 https://github.com/carbon-design-system/carbon/issues/10031

metonym:
Closes #899

By default, the width of each tab is set to `10rem`.

Set `autoWidth` to `true` for tabs to have an automatically set width.

```svelte
<Tabs autoWidth>
  <Tab label="Tab label 1" />
  <Tab label="Tab label 2" />
  <Tab label="Tab label 3" />
  <div slot="content">
    <TabContent>Content 3</TabContent>
  </div>
</Tabs>
```

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -3984,12 +3984,13 @@ None.
 
 ### Props
 
-| Prop name       | Kind             | Reactive | Type                                      | Default value                    | Description                                 |
-| :-------------- | :--------------- | :------- | :---------------------------------------- | -------------------------------- | ------------------------------------------- |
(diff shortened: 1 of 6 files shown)


metonym:
Released in [v0.48.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.48.0)

PR #905, opened 2021-11-12
hotfix: fix remaining imports from carbon-icons-svelte

metonym:
#904 inlined icons from `carbon-icons-svelte`. Unfortunately, many references were missed in the script.

This didn't throw errors when building the library because `carbon-icons-svelte` is installed as a devDependency.


metonym:
Fixed in [v0.48.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.48.1)

PR #911, opened 2021-11-16
Correct page number when dispatching update

ibilux:
When in last page of data and the page size get changed to a bigger size the current page number will be out of range of total pages.
This code will set the page number in range of total pages.

### DIFF
--- a/src/Pagination/Pagination.svelte
+++ b/src/Pagination/Pagination.svelte
@@ -70,7 +70,7 @@
   /** Set an id for the top-level element */
   export let id = "ccs-" + Math.random().toString(36);
 
-  import { createEventDispatcher } from "svelte";
+  import { afterUpdate, createEventDispatcher } from "svelte";
   import CaretLeft16 from "../icons/CaretLeft16.svelte";
   import CaretRight16 from "../icons/CaretRight16.svelte";
   import Button from "../Button/Button.svelte";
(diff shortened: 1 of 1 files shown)


metonym:
Could you provide a repro?

ibilux:
> Could you provide a repro?

[Here](https://svelte.dev/repl/d7d6bbe381be49d39fb042c4c9aabd63?version=3.44.1),
If you select **10** items per page and select page **5**,
And then select **100** items per page,
The current page will stay **5**, and the "Next page" button will still work and increasing the page number.
# I hope I made it clear?

metonym:
Nice catch – thanks for providing a repro.

As for the failing CI, it seems that Svelte has detected a circular dependency between `page` and `totalPages`.

I think that using the `afterUpdate` lifecycle method should work:

```js
afterUpdate(() => {
  if (page > totalPages) {
    page = totalPages;
  }
});
```

ibilux:
> ```js
> afterUpdate(() => {
>   if (page > totalPages) {
>     page = totalPages;
>   }
> });
> ```

Looks very good.

metonym:
Thanks again for reporting this!

metonym:
Fixed in [v0.49.2](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.49.2)

Issue #811, opened 2021-09-05
ContextMenu, limit to div


Sheepux:
I've done a contextualised contextmenu, meaning that i can handle on wich div it appears instead of appearing on all my page (outside of my rendered component).
I was wondering if it's the correct way to handle it or if i should have used the contextMenu differently.

That's probably something that could be introduced as a new prop "limitTo" (HtmlElement).
```javascript
<script>
// Context Menu
    let itemsdiv: HTMLElement;
    let pos = { x: 0, y: 0 };
    let showMenu = false;

    async function onRightClick(event: any) {
        if (showMenu) {
            showMenu = false;
            await new Promise((res) => setTimeout(res, 100));
        }
        pos = { x: event.clientX, y: event.clientY };
        showMenu = true;
    }

    document.addEventListener(
        "contextmenu",
        function (event: any) {
            if (event.target === itemsdiv || itemsdiv.contains(event.target)) {
                event.preventDefault();
                return;
            }
            showMenu = false;
        },
        true
    );
</script>
```
manually handing menu opening because of manual event handling
```html
{#if showMenu}
    <ContextMenu open={showMenu} x={pos.x} y={pos.y}>
 ```
 Specifying the desired div
 ```html
 <div  bind:this={itemsdiv} on:contextmenu={onRightClick} >
 ```
 I havent tried to PR
 I guess i would have to change[ something here ](https://github.com/carbon-design-system/carbon-components-svelte/blob/master/src/ContextMenu/ContextMenu.svelte#L83) to check if we're inside the limitTo prop

metonym:
In [version 0.50](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.50.0), you can now pass custom elements as trigger targets to `ContextMenu`

Example: https://carbon-components-svelte.onrender.com/components/ContextMenu#custom-target

Sheepux:
I believe the example should have 3 div to show that you can have different menus (and share 1):
DIV1 (menuA) DIV2 (menuA) DIV3 (menuB)
I'll test it later. Thanks a lot for this one 👍

Issue #620, opened 2021-05-02
ContextMenu can only be used on the whole page


NyxCode:
Currently, `ContextMenu` can only be used on the whole page.
Sadly, this is pretty limiting - I'd like to use it only on an element on my page.

Could `ContextMenu` be split into a dumb sub-component which doesn't listen for events itself?
I would like to just give it `x`, `y` as props and handle everything else myself.

metonym:
I agree that there should be an option to use it on an arbitrary element.

I think that adding a `target` prop (`HTMLElement`) to `ContextMenu` would be a good approach. The default would still be the `window`.

NyxCode:
I guess that would work for most cases, but I'm not sure that would be flexible enough.
At least for my usecase, I need to have a context menu for a *lot* of elements, and I wouldn't want a seperate `ContextMenu` for every one of them.

I think this would be a sane change for `ContextMenu`, but I still believe that something like `ContextMenuInner` (#622) would be great for more compilcated usecases.

geoidesic:
Should just be called `Menu` if it can only be used on the whole page. The whole point of adding the world "Context" before menu is, well.. for a menu that differs by context. Boggles the mind :)

Trystan-SA:
I also don't understand why ContextMenu is indeed not contextualized based on what item you right-click. Please at least populate the passed on:open forwarded event with a reference to the DOM element we have the cursor over.

 ```js
<ContextMenu on:open={(...e)=>{console.log(e); }}>
	<ContextMenuOption indented labelText="Copy"/>
	<ContextMenuOption indented kind="danger" labelText="Delete" />
</ContextMenu>
```

is currently returning this : 
![Capture d’écran 2021-11-17 194339](https://user-images.githubusercontent.com/43876044/142262693-f4be0310-64a9-4522-affb-78e7711af76a.png)

**Comment by metonym** [comment]
In [version 0.50](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.50.0), you can now pass custom elements as trigger targets to `ContextMenu`

Example: https://carbon-components-svelte.onrender.com/components/ContextMenu#custom-target

@Trystan-SA The `event.target` is passed to the `on:open` event detail as well.

```svelte
<ContextMenu on:open={(e) => {
  console.log(e.detail); // HTMLElement
}}>
  ...
</ContextMenu>
```

geoidesic:
[...]

PR #916, opened 2021-11-18
Support custom target(s) for `ContextMenu`

metonym:
Related #620, #811

**Features**

- support custom target elements for `ContextMenu` using the `target` prop
- pass clicked element as `event.detail` in `ContextMenu` "open" event

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -780,12 +780,13 @@ None.
 
 ### Props
 
-| Prop name | Kind             | Reactive | Type                                      | Default value      | Description                                                                      |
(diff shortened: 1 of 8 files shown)


jeasoncc:
what should i do for ContextMenu in an arrays elements

metonym:
@jeasoncc Does this [example](https://carbon-components-svelte.onrender.com/components/ContextMenu#multiple-targets) help?

geoidesic:
@metonym this still isn't useful because it's still a global menu, not a context menu. It's global in the sense that it is positioned globally, outside of the structure of your site's HTML, out of context of the triggering element. It's now just triggered contextually... but it remains a global element. It offers no way of attaching the positioning of the element to an element within your HTML structure. Therefor  it should be called GlobalMenu.

PR #926, opened 2021-12-01
fix(combobox): filtered ComboBox handling with keyboard

patoi:
Filtered ComboBox handling with keyboard has a bug, that prevent using.

**Reproducing:**
On demo site ComboBox/Filterable https://carbon-components-svelte.onrender.com/components/ComboBox click on input field, type Fax, press ArrowDown, and press Enter: the input value will be Slack. Expected: Fax


metonym (commented):
Thanks for reporting this and taking the time to work on this PR.

This does indeed select the correct item when using keyboard navigation. However, I encountered the following regressions:

**"select" event is dispatched on every key stroke after an initial selection**

https://user-images.githubusercontent.com/10718366/144509757-7ba60387-8424-4f98-b4bf-387e95dae436.mov

**Initial `selectedIndex` regression**

When testing the following example, no item is selected. The expected behavior is that the second item should be initially selected.

```svelte
<ComboBox
  titleText="Contact"
  selectedIndex={1}
  items={[{ id: '0', text: 'Slack' }, { id: '1', text: 'Email' }, { id: '2', text: 'Fax' }]}
/>
```

patoi:
Thanks! I fixed these issues.

May I have noticed another previous problem: it doesn't send "select" event (selecting the nothing), when click on clear "X" button. May be a dispatch call is missing from the clear method?

patoi:
Hi! I think this patch works fine. Is it mergeable?

metonym (commented):
Ignore... I'd previously broken the symlink when testing.

Re-reviewing.

metonym:
@patoi The keyboard selection works better. Encountering the following issue.

1. Search "Fa" and select "Fax" using the keyboard
2. Clear the selection so that all options are available
3. Select "Email"
4. The first option ("Slack") is selected

I wonder if some value needs to be reset for this to work with clicking?

```svelte
<script>
  import { ComboBox } from "carbon-components-svelte";

  function shouldFilterItem(item, value) {
    if (!value) return true;
    return item.text.toLowerCase().includes(value.toLowerCase());
  }
</script>

<ComboBox
  titleText="Contact"
  items={[
    { id: "0", text: "Slack" },
    { id: "1", text: "Email" },
    { id: "2", text: "Fax" },
  ]}
  on:select={(e) => {
    console.log(e.detail);
  }}
  {shouldFilterItem}
/>

```

https://user-images.githubusercontent.com/10718366/147888866-1ee3755f-d715-4fa7-aecf-90f74cc462aa.mov

patoi:
When you clicked on a row, the on:click doesn't run, because on:blur catch the event:

```javascript
        on:blur
        on:blur="{({ relatedTarget }) => {
          if (inputValue.length === 0 && selectedIndex > -1) {
            if (filteredItems[selectedIndex]) {
              inputValue = filteredItems[selectedIndex].text;
            }
          }
          if (!open || !relatedTarget) return;
          if (
            relatedTarget &&
            !['INPUT', 'SELECT', 'TEXTAREA'].includes(relatedTarget.tagName) &&
            relatedTarget.getAttribute('role') !== 'button' &&
            relatedTarget.getAttribute('role') !== 'searchbox'
          ) {
            ref.focus();
          }
```

If I delete on:blur event handlers, then it works fine.

- What is the purpose of the blur feature?
- Is it removable?

Thanks for your answer!

metonym:
It seems that the bug is actually unrelated – I'll address this in a separate PR.

Could you investigate the following case? Two-way binding using `selectedIndex` doesn't work as expected. Currently, "Email" should be set as the default, but no value is set. Similarly, programmatically setting `selectedIndex` has no effect.

```svelte
<script>
  import { ComboBox } from "carbon-components-svelte";

  let selectedIndex = 1;
</script>

<ComboBox
  titleText="Contact"
  bind:selectedIndex
  items={[
    { id: "0", text: "Slack" },
    { id: "1", text: "Email" },
    { id: "2", text: "Fax" },
  ]}
/>

<div on:click={() => (selectedIndex = 0)}>Set to 0</div>
```

patoi:
I've worked on it for several hours, and it's ok finally. But the blur handler is still a problem.
When I deleted the blur handler the problem has gone.

Here's how the program works:
## ✅ without filtered items
https://user-images.githubusercontent.com/1846548/148238293-d819f4f1-02b6-4bce-9d83-aeea65b891ae.mp4

## ✅ with filtered items and deleted blur handler
https://user-images.githubusercontent.com/1846548/148238731-6f9516af-b1c7-4920-a30f-068f3bd657df.mp4

## ❌ with filtered items (blur handler untouched):
https://user-images.githubusercontent.com/1846548/148238964-04e14572-14fe-4de3-9f60-fd9c8dda9f24.mp4

You can see on the last video:
1. typing 'email'
2. select with keyboard: email
3. delete selection with Backspace
4. select Fax **by mouse** from pop-up list
5. result: Email ❌

But if you're using just keyboard:
1. typing 'email'
2. select with keyboard: email
3. delete selection with Backspace
4. select Fax **by keyboard** from pop-up list
5. result: Fax ✅

After deleting blur handler, the first case has worked.

Some design notes: in my opinion, the operation of the model and the view model should be more strongly separated. The model consists of the items and the selected item's id, and the view model is arbitrary. The component converts between them. In this case:

```
<ComboBox
  bind:selectedId
  {items}
/>
```
where `selectedId` is item's id, and item is { id: string, text: string } (same as now).
In this case the `selectedIndex` is missing from properties, and user can't access it. (Maybe it's not even needed as an internal variable.)

I think this design is more reliable, and easier to use from the user side.

metonym:
@patoi This is looking great. I agree on the part about updating the `on:blur` logic that is interfering with the clearing.

Could you remove the following from the `on:blur`?

```
if (inputValue.length === 0 && selectedIndex > -1) {
            if (filteredItems[selectedIndex]) {
              inputValue = filteredItems[selectedIndex].text;
              selectedItem = filteredItems[selectedIndex];
              selectedId = filteredItems[selectedIndex].id;
            }
          }
```

patoi:
Thanks! It works fine! 🍾
Pushed and merged master.

metonym (approved):
Great work – thanks for seeing this through!

metonym:
Released in [v0.51.2](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.51.2)

PR #946, opened 2021-12-24
Modify tab click events to prevent navigation.

### LINKED ISSUE #626: Tabs issue on responsive / mobile
You can observe this on the documentation site, either on mobile, or using dev tools in responsive mode, if you click on the actual a link with the `href='#'` you get an undesirable navigation to the top of the current page - its fine if you click the `li` avoiding the `a` tag

brunnerh:
Without `preventDefault` the link will navigate the page instead of opening the dropdown when the tabs are in the compact mode.
Because the parent element also toggles on click, there is an additional `stopPropagation` to not toggle twice which would be a no-op.

Fixes #626.

### DIFF
--- a/src/Tabs/Tabs.svelte
+++ b/src/Tabs/Tabs.svelte
@@ -130,8 +130,8 @@
       tabindex="-1"
       class:bx--tabs-trigger-text="{true}"
       href="{triggerHref}"
-      on:click
-      on:click="{() => {
+      on:click|preventDefault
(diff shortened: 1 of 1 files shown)


metonym (approved):
Well done – a very elegant solution

metonym:
Released in [v0.50.3](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.50.3)

PR #947, opened 2021-12-27
Enhance Checkbox to use bind:group

### LINKED ISSUE #612: Checkbox Bind Group
Thank you for creating this awesome library for Svelte.

I had a question regarding the `Checkbox` component. Is there any variation of the checkbox or similar checkbox like component that supports [bind:group](https://svelte.dev/docs#bind_group)? It looks like `value` is not exposed by `Checkbox`. Or is it intended to use [SelectableTile](https://carbon-svelte.vercel.app/components/SelectableTile) for that type of functionality?

Thanks!

gregorw:
Fixes #612. 

https://user-images.githubusercontent.com/988276/147611961-85231b95-3159-471f-93a9-e87507f6d314.mov

This adds Svelte-like [`bind:group`](https://svelte.dev/tutorial/group-inputs) behavior to Checkbox.

In the first commit I tried to leverage Svelte’s [own](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality#comparing_objects) `bind:group`. Simply forwarding this to the native input element did not work due to [strict equality issues](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality#comparing_objects). See also “custom” checkboxes in [this REPL](https://svelte.dev/repl/b68755d741664ddab2bf1be263d0058c?version=3.44.3). 

So, this left us with a custom implementation of the same logic that Svelte already implements. It’s straightforward though and only a few lines long. I hope this is a viable improvement. Of course, everything is open for discussion and I am happy to hear and incorporate any feedback. 

I was uncertain about testing. What do we do there?


gregorw:
Rebased onto master.

gregorw:
@metonym What do you think? :)

metonym (changes requested):
Thank you for taking the initiative – `bind:group` would be a really useful feature!

When testing locally, the `bind:group` example that you provided works perfectly. However, the single checkbox case where `checked` is initially true no longer works.

In the following example, the single checkbox is not toggled by default.

```svelte
<script>
  import { Checkbox } from "carbon-components-svelte";
  
  let checked = true;
</script>

<Checkbox
  bind:checked
  labelText="value"
  value="value"
  on:check={(e) => {
    console.log("on:check", e.detail);
  }}
/>
```

This is a bit tricky because – from what I've tried – it's not possible to use reactive statements/assignments for both `checked` and `group` without causing a cyclical dependency error.

As such, I propose the following approach to support two-way binding:

1. determine if the `input` is checked if either `checked` or `group.includes(value)` is truthy

```diff
- checked="{checked}"
+ checked="{checked || group.includes(value)}"
```

2. Forgo reactive ($:) syntax in favor of setting `checked` and `group` directly in the `on:change` handler

```diff
on:change="{() => {
    checked = !checked;
+ group = group.includes(value)
+    ? group.filter((_value) => _value !== value)
+     : [...group, value];
 }}"
```

gregorw:
> When testing locally, the `bind:group` example that you provided works perfectly. However, the single checkbox case where `checked` is initially true no longer works.

Great catch! 😅

> This is a bit tricky because – from what I've tried – it's not possible to use reactive statements/assignments for both `checked` and `group` without causing a cyclical dependency error.

Totally agree with your proposed approach here. I just pushed it.

Of course, the cyclic dependency could be resolved by braking one direction of the cycle:

```
checked -----

File: src/Checkbox/Checkbox.svelte [hunk 1]
```diff
@@ -3,12 +3,21 @@
    * @event {boolean} check
    */
 
-  /** Specify the value of the checkbox */
+  /**
+   * Specify the value of the checkbox
+   * @type {string | any}
```
metonym, line 8: Nice catch, typing this as any. We can remove the redundant "string" type.

```suggestion
   * @type {any}
```

metonym (changes requested):
Great work!

I've tested:

1. Uncontrolled, single checkbox

```svelte
<Checkbox labelText="value" value="value" />
```

2. Two-way binding, single checkbox

```svelte
<Checkbox
  bind:checked
  labelText="value"
  value="value"
  on:check={(e) => {
    console.log("on:check", e.detail);
  }}
/>
```

3. Two-way binding using `bind:group` with initial values (your example)
4. Two-way binding using `bind:group` with no initial values

One requested change is to update the `value` prop type and re-generate the types/docs.

src/Checkbox/Checkbox.svelte
gregorw: Absolutely. Done.

metonym (approved):
💯

gregorw:
@metonym Thanks for reviewing 💪

PR #962, opened 2021-12-30
Fix docs dev/build on Windows/newer Node version

### LINKED ISSUE #959: Docs build system out of date?
The `svite` package is on an old version that causes errors with newer versions of node:

> error svite@0.7.2: The engine "node" is incompatible with this module. Expected version "^12||^14". Got "16.7.0"

There appears to be a vite plugin that is [supposed to replace this packge](https://github.com/svitejs/svite/tree/main/packages/svite#migration-from-08).

brunnerh:
Switch from svite to vite + vite-plugin-svelte.
Use `cross-env` to set environment variables platform independently in scripts.

Fixes #959

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -2,20 +2,22 @@
   "private": true,
   "scripts": {
     "dev": "run-p dev:*",
-    "dev:routify": "NODE_ENV=development routify run",
-    "dev:svite": "svite",
+    "dev:routify": "cross-env NODE_ENV=development routify run",
+    "dev:svite": "vite dev",
     "build": "run-s build:*",
     "build:routify": "routify run -b",
-    "build:svite": "svite build"
+    "build:svite": "vite build"
   },
   "devDependencies": {
     "@sveltech/routify": "^1.9.9",
+    "@sveltejs/vite-plugin-svelte": "^1.0.0-next.33",
     "autoprefixer": "^10.2.3",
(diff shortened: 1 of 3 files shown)


brunnerh:
If that should be split or `cross-env` should not be used at all this can be adjusted/rejected as required.

brunnerh:
Also, I do not know what the performance of the `svite` system was like; it seems rather slow now.
This might need additional adjustments to `vite.config.js`.

metonym (approved):
Awesome – thanks for getting to the bottom of this!

I tested this locally and was able to run `yarn dev` and `yarn build`.

I also deployed it on Render and it works as expected: https://ccs-docs-win.onrender.com/

The build time seems to have drastically improved; it's about ~21% faster.

brunnerh:
@metonym If that is "fast", I am not impressed 😅
The hot module reload in dev modes takes a very long time, not the "almost instant" I am used to from other projects.
I suspect that it rebuilds too many files unnecessarily.

metonym:
@brunnerh True – I was referring to the production build. The current `yarn dev` has been consistently slow as of late. I suspect the bottleneck is related to how many passes the various preprocessors do on all the files.

When testing/developing this library locally, I've mainly been using a fresh SvelteKit project that symlinks "carbon-components-svelte."

Eventually, the goal is to move to SvelteKit.

PR #968, opened 2022-01-03
Add `required` prop to Select.

### LINKED ISSUE #964: Select $$restProps on wrong element/required missing.
The `Select` component currently cannot be made required for form submissions. I intended to create a pull request to fix this, but noticed that this could either be achieved by...
- having the `$$restProps` apply to the `<select>` element rather than the root wrapper
- or add an explicit `required` property that is then applied to the `<select>`

Maybe even both. The code seems a bit inconsistent, e.g. TextInput has the props on the input while Checkbox has it on a wrapper element. Personally, I think it makes the most sense to always apply the rest props to the interactive element (input/select/textarea/button).

brunnerh:
Upgrade sveld to set paths correctly on Windows.
Make Prettier script compatible with Windows and run it.

Fixes #964


brunnerh:
Having an explicit prop seems helpful, but feel free to reject.
Could also move `$$restProps` in addition to the new prop, that would be a breaking change, though.

pheuter:
I was _just_ about to report an issue for this, 👍 would be great to be able to mark `<select>` elements as required for form validation.

metonym (approved):
👍

PR #971, opened 2022-01-05
Update TileGroup on programmatic selection change.

### LINKED ISSUE #859: Selected tile in TileGroup does not update on programmatic selection change
When changing the value of a variable that is used for the `selected` property, the UI will not update.

Example:

```svelte
<script>
	// ...
	let selected = undefined;
	let show = true;
</script>

{#if show}
	<TileGroup bind:selected>
		<RadioTile value="1">Tile 1</RadioTile>
		<RadioTile value="2">Tile 2</RadioTile>
		<RadioTile value="3">Tile 3</RadioTile>
	</TileGroup>
{/if}

<p>Selection: {selected}</p>

<Button on:click={() => selected = "2"}>Set to 2</Button>

<Button on:click={() => {
	show = false;
	setTimeout(() => show = true);
}}>Force re-render</Button>
```

[REPL](https://svelte.dev/repl/690f0f2f9c314184bed46bb16c457ea2?version=3.44.3)

The button can be used to set the value directly, but the component will only reflect the change upon re-rendering.

---

Looking at the source, the radio tiles use a store from the context defined by the group. But the updates only go one-way, so the store is not updated:

```
$: selected = $selectedValue;
```

Changes to `selected` should also update `$selectedValue` and thus in turn the tiles. Not sure how to best resolve this, though; Svelte does not like cyclic dependencies.

brunnerh:
Also only fire `select` event for user interaction (from RadioTile).

Uses `set` function instead of `$` to prevent cyclic dependency error.

Fixes #859.

### DIFF
--- a/src/Tile/TileGroup.svelte
+++ b/src/Tile/TileGroup.svelte
@@ -26,11 +26,12 @@
     },
     update: (value) => {
       selectedValue.set(value);
+      dispatch("select", value);
     },
   });
 
   $: selected = $selectedValue;
-  $: dispatch("select", $selectedValue);
+  $: selectedValue.set(selected);
 </script>
 
 <fieldset disabled="{disabled}" class:bx--tile-group="{true}" {...$$restProps}>


metonym (approved):
Excellent

PR #973, opened 2022-01-05
Propagate `name` and `required` from TileGroup.

### LINKED ISSUE #972: Native RadioTile required validation not supported
Currently a selection of a radio tile cannot be enforced using native form validation. To use this the same `name` has to be set on all radio inputs along with the `required` attribute.

Suggested changes:
- `name` should be propagated from `TileGroup` to `RadioTile`. It makes no sense for the elements of the group to have different names; they would no longer be part of the same group. The `name` property would then be removed from `RadioTile`.
- `required` should also be propagated down.

I started to implement this, but currently is still an [upstream layout issue](https://github.com/carbon-design-system/carbon/issues/10343). (The validation message is displayed in the wrong place.)

brunnerh:
Remove `name` from `RadioTile`.

Fixes #972.

Probably causes conflicts with pull request #971

### DIFF
--- a/src/Tile/RadioTile.svelte
+++ b/src/Tile/RadioTile.svelte
@@ -20,13 +20,11 @@
   /** Set an id for the input element */
   export let id = "ccs-" + Math.random().toString(36);
 
-  /** Specify a name attribute for the input */
-  export let name = "";
-
   import { getContext } from "svelte";
   import CheckmarkFilled16 from "../icons/CheckmarkFilled16.svelte";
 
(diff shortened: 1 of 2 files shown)


theetrain (commented):
Tested, and overall these changes work well. 👍 

All that remains are some documentation updates. I do not see any conflicts upstream nor elsewhere. cc @metonym

brunnerh:
Obsoleted by:
- #1818

PR #991, opened 2022-01-10
breaking: drop support for ES, UMD formats

### LINKED ISSUE #896: TypeError: Cannot read properties of undefined (reading '$$') when using 0.42.0 or newer
When I'm using any components from carbon-components-svelte. It trigger the following error :

chrome_bxTO9mFTBa
chrome_YRSjjmqjap

Here is the code that trigger the error. It happen when I'm using any components (buttons, forms, accordions): 
Code_R6CQCZ0gOO

I don't have this issue if I'm using the 0.41.0 version. It start breaking at 0.42.0 or newer.

My package versions :
svelte: 3.44.1
esbuild: 0.13.12
esbuild-svelte: 0.5.7  (and tested with 0.5.3 too)

### LINKED ISSUE #1427: Upgrading from 0.41 results in "Cannot read properties of undefined (reading ‘$$’)"
Hi there.

I'm trying to upgrade my (previously working) project to use newer versions of the dependencies. I've been using Carbon components and am running into the above error message. I've done a fair bit of Googling, without success so far, but I have manged to make a set of minimal steps to reproduce the issue:

1. With Meteor 2.7.3 (I'm using the git repo, on tag release/METEOR@2.7.3), run meteor create test --svelte
2. meteor run
3. meteor npm i --save carbon-components-svelte
4. Edit imports/ui/App.svelte, adding:
    import { Button } from 'carbon-components-svelte';
5. Modify button to Button in both places on line 14.

Things that I've tried:
1. Update svelte to ^3.48 (https://github.com/carbon-design-system/carbon-components-svelte says it requires this)
2. Downgrade carbon-components-svelte to 0.60 or 0.50, both without success.

The working version is using carbon-components-svelte 0.41.

I'm thinking the issue is in carbon-components-svelte because if I just upgrade carbon-components-svelte, that's what breaks my build. I can of course find the issue reported elsewhere, but not a solution that is helping in this case.

Can someone help, please? Thanks in advance!

metonym:
Fixes #896, fixes #1427

This PR drops support for the compiled ES, UMD formats. Instead, only the uncompiled Svelte source code will be shipped.

This should have no impact for users of Vite/SvelteKit/Webpack/Rollup. This solves issues for set-ups that experience errors using the compiled code (#896). A bonus is that the `carbon-components-svelte` package will be considerably lighter in size.


dummdidumm:
FYI there's a feature request in SvelteKit which would enable also bundling to ESM/UMD: https://github.com/sveltejs/kit/issues/2572 . There are valid cases where you would want people to be able to use your library standalone. If that's the case for `carbon-components-svelte` is another question.

metonym:
@dummdidumm Thank you for the info! Good to know that there are potential use cases for the bundled code.

Issue #1003, opened 2022-01-13
Probably `selectedId` is better than `selectedIndex` in ComboBox and Dropdown


KoichiKiyokawa:
First, thank you for working on this library.
I use this library in production and it is very useful 🎉 
But, I found that the reactive prop of `selectedIndex` in `ComboBox` and `Dropdown` was difficult to use.
It would be better if it is `selectedId`.

The reasons are as follows.
## Matches with other select like components.
- `Select` component has a `selected` prop (it is equivalent to selectedId).
- `MultiSelect` component has a `selectedIds` prop.

## Easier to connect with server side
When sending values to the server side, it is often necessary to use the `selectedId` instead of the `selectedIndex`.

## Matches HTML `<select>` tags.
Since `ComboBox` and `Dropdown` are used as alternatives to the `<select>` tag, it is better if the APIs match, IMO.

```svelte
<!-- select tag use case in svelte-->
<script>
let selectedFruit
</script>

<select bind:value={selectedFruit}>
  <option value="apple">Apple</option>
  <option value="banana">Banana</option>
</select>
```

Currently, it is necessary to declare an extra variable when using `ComboBox` or `Dropdown`.

```svelte
<script>
const items = [{ id: 'apple', text: 'Apple' }, { id: 'banana', text: 'Banana' }]
let selectedFruitIndex = -1 // the extra variable
$: selectedFruit = items[selectedIndex] // we need this to send the server side.
</script>

<Dropdown {items} bind:selectedIndex={selectedFruitIndex} />
```

or, to avoid extraneous variables, do the following.

```svelte
<script>
const items = [{ id: 'apple', text: 'Apple' }, { id: 'banana', text: 'Banana' }]
let selectedFruit
</script>

<Dropdown {items} selectedIndex={items.findIndex(item => item.id === selectedFruit)} on:select={e => selectedFruit = e.detail.selectedId} />
```

If `ComboBox` and `Dropdown` have a reactive `selectedId`, we can do the following. It is very smart.

```svelte
<script>
let selectedFruit
</script>

<Dropdown items={[{ id: 'apple', text: 'Apple' }, { id: 'banana', text: 'Banana' }]} bind:selectedId={selectedFruit} />
```

This issue is just a suggestion, so feel free to close 👍

metonym:
I agree.

[...]

PR #1004, opened 2022-01-14
feat(Dropdown): selectedIndex -> selectedId

KoichiKiyokawa:
part of #1003

Modify `selectedIndex` to `selectedId` in Dropdown component.
(I will create another PR for ComboBox later to make the diff smaller.)

### DIFF
--- a/src/Dropdown/Dropdown.svelte
+++ b/src/Dropdown/Dropdown.svelte
@@ -18,8 +18,11 @@
    */
   export let itemToString = (item) => item.text || item.id;
 
-  /** Specify the selected item index */
-  export let selectedIndex = -1;
+  /**
+   * Specify the selected item id
+   * @type {DropdownItemId}
+   */
+  export let selectedId = undefined;
 
   /**
    * Specify the type of dropdown
@@ -108,7 +111,6 @@
 
   const dispatch = createEventDispatcher();
 
-  let selectedId = undefined;
   let highlightedIndex = -1;
 
   function change(dir) {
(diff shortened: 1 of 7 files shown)


src/Dropdown/Dropdown.svelte
```diff
@@ -123,12 +125,12 @@
     highlightedIndex = index;
   }
 
-  $: if (selectedIndex > -1) {
+  $: if (selectedId !== undefined) {
     dispatch("select", { selectedId, selectedIndex, selectedItem });
```
KoichiKiyokawa, line 129: I'm not sure if we should dispatch `selectedIndex` 🤔

metonym: Let's leave it in for now

metonym (changes requested):
This looks great!

I agree that `selectedIndex` can be omitted from the `select` event.

src/Dropdown/Dropdown.svelte
KoichiKiyokawa: Done in d52828a 🙆‍♂️

metonym (approved):
Excellent work! I tested two-way binding and the `on:select` worked as expected.

I'd like to get this merged/released in the next minor version (0.53) along with the changes to ComboBox.

KoichiKiyokawa:
@metonym 
Thank you for reviewing :tada:
I created the PR for `ComboBox` in #1016

metonym:
Released in [v0.53.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.53.0).

PR #1016, opened 2022-01-18
feat(ComboBox): selectedIndex -> selectedId 

### LINKED ISSUE #1003: Probably `selectedId` is better than `selectedIndex` in ComboBox and Dropdown

KoichiKiyokawa:
fix #1003
continued from #1004


src/ComboBox/ComboBox.svelte
```diff
@@ -269,29 +267,25 @@
...
+              highlightedIndex > -1 &&
+              filteredItems[highlightedIndex]?.id !== selectedId
+            ) {
               open = false;
-              highlightedIndex = -1;
-              if (filteredItems[selectedIndex]) {
-                inputValue = filteredItems[selectedIndex].text;
-                selectedItem = filteredItems[selectedIndex];
-                selectedId = filteredItems[selectedIndex].id;
+              if (filteredItems[highlightedIndex]) {
```
KoichiKiyokawa, line 275: We can use `highlightedIndex` because `highlightedIndex = -1;` was moved to line 288.

src/ComboBox/ComboBox.svelte
```diff
@@ -149,34 +150,31 @@
...
       if (!selectedItem) {
         selectedId = undefined;
-        selectedIndex = -1;
         inputValue = "";
         highlightedIndex = -1;
         highlightedId = undefined;
-        prevSelectedIndex = undefined;
       } else {
-        // programmatically set selectedIndex
+        // programmatically set inputValue
```
metonym, line 157: Good catch

metonym (approved):
Thank you for seeing this through, and for also updating the examples.

I tested:

- two-way binding using `selectedId`
- filtering
- keyboard navigation
- clearing selection using the "X"
- programmatically clearing using the accessor

PR #1017, opened 2022-01-18
fix(Dropdown): list icon was clickable in disabled state

patoi:
https://user-images.githubusercontent.com/1846548/149938326-e2ab9415-3178-429d-bbf7-0f0cdd66a529.mp4

### DIFF
--- a/src/Dropdown/Dropdown.svelte
+++ b/src/Dropdown/Dropdown.svelte
@@ -176,6 +176,7 @@
       {disabled && 'bx--dropdown--disabled'}
       {light && 'bx--dropdown--light'}"
     on:click="{({ target }) => {
+      if (disabled) return;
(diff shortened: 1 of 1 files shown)


patoi:
There is one more bug...

patoi:
ListBox was selectable, when extended Dropdown component.

patoi:
MacOS 12.1:
The bug does **not** reproducable in Firefox 96.0.1.
Affected browser is Chrome 97.0.4692.71 (x86_64)

metonym (approved):
Nice catch 🙏

metonym:
Fixed in [v0.53.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.53.0)

PR #1044, opened 2022-01-27
breaking(number-input): type `value` as `null | number`

### LINKED ISSUE #1039: `NumberInput` change event data or type is off
The `change` event dispatches the `value` property as its payload.
The event is typed as `number`, the property as `number | string`.

`value`, can implicitly be `undefined` as well, maybe that needs to be considered, too.

metonym:
Fixes #1039

This uses `null` instead of an empty string to denote "no value."

```svelte
<script>
  import { NumberInput, Button } from "carbon-components-svelte";

  let value = null;

  $: console.log("bind:value", value, typeof value);
</script>

<NumberInput
  allowEmpty
  bind:value
  on:change={(e) => console.log("on:change", e.detail, typeof e.detail)}
/>

<Button on:click={() => (value = null)}>Set to null</Button>
<Button on:click={() => (value = 0)}>Set to 0</Button>
```

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -2560,32 +2560,32 @@ export type NumberInputTranslationId = "increment" | "decrement";
 
 ### Props
 
-| Prop name       | Kind               | Reactive | Type                                                            | Default value                                                    | Description                                     |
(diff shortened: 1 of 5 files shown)


metonym:
@brunnerh does this work with your use case?

brunnerh:
@metonym Yes, looks good 👍

PR #1047, opened 2022-01-29
Allow modals to have button icons

gregorw:
Buttons may have icons. Modals (mostly) have buttons. Should modal buttons have icons?

In this PR I would like to allow an icon on the _primary_ button of a modal. I’m happy to propose a separate PR if we want to consider secondary buttons as well.

image


gregorw:
Fixes #1048

metonym (approved):
Looks great – thanks for also including an example in the docs!

PR #1049, opened 2022-01-29
Allow modals to have button props

### LINKED ISSUE #1048: Modal button icons
We would like to use a primary button _icon_ in modals.

In Carbon Design System buttons already have icons. Modals have buttons most of the time. So, this seems to be a logic addition. The [guidelines](https://www.carbondesignsystem.com/components/modal/usage/) do not cover this, though. The closest thing I could find was a [loading indicator on a modal button](https://www.carbondesignsystem.com/components/modal/usage/#loading).

image

What we want is this:

image

What do you think?

I have already implemented this in #1047.

gregorw:
This is an alternative — and more generic — implementation to #1047 that fixes #1048.

@metonym Which one do you prefer?


gregorw:
Closing in favor of #1047

PR #1051, opened 2022-01-30
Check number input value against `""`.

### LINKED ISSUE #1050: NumberInput not binding `null` and defaulting to `0`
The null check introduced in the recent fix (https://github.com/carbon-design-system/carbon-components-svelte/commit/402de56029f6c400df335e8182be65c79cc5dd9f) is invalid due to the fact that an input's value is always a string.

This causes `''` to be coerced to `0` instead of returning `null`.

[REPL](https://svelte.dev/repl/7408eee2e9f044269504e6e99dd6aa96?version=3.46.3)

brunnerh:
Fixes #1050.

### DIFF
--- a/src/NumberInput/NumberInput.svelte
+++ b/src/NumberInput/NumberInput.svelte
@@ -135,7 +135,7 @@
   $: dispatch("change", value);
   $: incrementLabel = translateWithId("increment");
   $: decrementLabel = translateWithId("decrement");
-  $: value = inputValue != null ? Number(inputValue) : null;
+  $: value = inputValue != "" ? Number(inputValue) : null;
   $: error =
     invalid || (!allowEmpty && value == null) || value > max || value < min;
   $: errorId = `error-${id}`;


src/NumberInput/NumberInput.svelte
```diff
@@ -135,7 +135,7 @@
   $: dispatch("change", value);
   $: incrementLabel = translateWithId("increment");
   $: decrementLabel = translateWithId("decrement");
-  $: value = inputValue != null ? Number(inputValue) : null;
+  $: value = inputValue != "" ? Number(inputValue) : null;
```
metonym, line 138: A couple of things:

1. A strict inequality check is needed here.

Otherwise, a value of `0` will be converted to `null`.

```
0 != ""; // false
0 !== ""; // true
```

2. We still need a null check to account for the following cases

This should render an empty value with an error outline:

```svelte
<NumberInput value={null} />
```

This should render an empty value without an error:

```svelte
<NumberInput allowEmpty value={null} />
```

```suggestion
  $: value = inputValue !== "" && inputValue != null ? Number(inputValue) : null;
```

metonym:
I see you raised #1053 as an alternative.

src/NumberInput/NumberInput.svelte
brunnerh: Ah, yes, because there is the initial assignment `inputValue = value`, so the variable could contain `string`, `number` and `null`. #1053 no longer has this issue.

brunnerh:
For #1053 I had to refactor the offending line and it probably would not merge with this correctly, so I integrated the change there. I guess this can be closed, if you approve of the other pull request.

metonym:
Superseded by #1053

PR #1053, opened 2022-01-30
Align NumberInput `change`/`input` with native.

### LINKED ISSUE #1052: Align NumberInput events with native inputs?
Currently `change` fires on input and the `input` event is simply forwarded (i.e. does not parse).

For the native inputs the firing semantics are:
- [`change`](https://html.spec.whatwg.org/multipage/indices.html#event-change): Fired at controls when the user commits a value change
- [`input`](https://html.spec.whatwg.org/multipage/indices.html#event-input): Fired at controls when the user changes the value

It would be nice if the component only fires `change` at the same time the input fires `change` in case uncommitted value changes are not of interest. Also, both `input` and `change` should provide the parsed value (`string` => `number | null`).

(Steppers should fire `change` and `input`, the native steppers do that as well.)

[REPL illustrating differences](https://svelte.dev/repl/6193507d3aea47dfbbddc0db381df6bd?version=3.46.3)

brunnerh:
Fixes #1052.
Obsoletes #1051.

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -2600,11 +2600,11 @@ export type NumberInputTranslationId = "increment" | "decrement";
 | Event name | Type       | Detail                          |
 | :--------- | :--------- | :------------------------------ |
 | change     | dispatched | <code>null &#124; number</code> |
+| input      | dispatched | <code>null &#124; number</code> |
 | click      | forwarded  | --                              |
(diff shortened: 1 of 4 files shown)


metonym (approved):
Nice – this works great

brunnerh:
Thank you for considering and even thoroughly testing this!

PR #1058, opened 2022-02-01
fix(tooltip): `Tooltip.svelte.d.ts` is missing open/close events

metonym:
**Fixes**

- type open/close events for `Tooltip`


metonym:
Fixed in [v0.57.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.57.1)

Issue #1062, opened 2022-02-02
DataTable sorting indicator appears in wrong TableColumnHead


berndfuhrmann:
Demo of problem:
https://svelte.dev/repl/3a27073d9b4f48f383b90c1dfd7395e2?version=3.46.3

Steps to reproduce:
1. load REPL
2. click on table header "Name"
3. Sorting indicator appear in column "Rule" instead of "Name"

Version tested: 0.57.1

metonym:
Nice find – the reactivity of the `headers` prop is buggy

PR #1063, opened 2022-02-02
Fix issue by adding id's in a different order

berndfuhrmann:
This should fix the issue I just reported: https://github.com/carbon-design-system/carbon-components-svelte/issues/1062
I'm not sure if this is the best way to do it or if changing headers should be legal at all. On the other hand, I don't see, how this would break anything.
Thanks for reviewing in advance!

### DIFF
--- a/src/DataTable/DataTable.svelte
+++ b/src/DataTable/DataTable.svelte
@@ -155,7 +155,7 @@
       if (refSelectAll) refSelectAll.checked = false;
     },
     add: (id) => {
-      headerItems.update((_) => [..._, id]);
+      headerItems.update((_) => [id, ..._]);
     },
   });
 


metonym (commented):
> changing headers should be legal at all

I definitely think `headers` should be reactive.

However, the off by one error still occurs for `headers` that aren't dynamically updated.

metonym (changes requested):
Looking at the code, I don't think there's any reason to generate an id per `TableHeader`.

We can instead just use the Header `key` as the id:

```diff
const thKeys = derived(headerItems, () =>
    headers
-     .map(({ key }, i) => ({ key, id: $headerItems[i] }))
+     .map(({ key }, i) => ({ key, id: key }))
      .reduce((a, c) => ({ ...a, [c.key]: c.id }), {})
  );
```

PR #1070, opened 2022-02-07
fix: duplicated SkeletonText heading removed

canerakdas:
Hello,

While reading the docs, I found duplication on SkeletonText


metonym (approved):
Nice catch, thank you

Issue #1069, opened 2022-02-05
Improve `Breakpoint` ergonomics?


brunnerh:
One of the main questions usually is: Is the viewport larger/smaller than *X*?
Currently this requires either explicitly listing all valid sizes or resolving the size from the read-only `breakpoints` prop to compare it.

e.g.

```svelte
<script>
	import { Breakpoint } from 'carbon-components-svelte';
	
	let size;
	let breakpoints = {};
	$: tooSmall = breakpoints[size] < breakpoints.lg;
</script>

<Breakpoint bind:size bind:breakpoints />
```

This is not great because:
1. The values are only available after the component is mounted.
2. The values are only available within components and not code in imported JS files (e.g. extracted layout logic).
3. Component is redundant because it does not do anything component-specific (like adding elements to the DOM).

At first I thought that maybe it should remain a component because of the event handler lifecycle management, but it can be converted to a store without sacrificing that: `onMount` and `onDestroy` can be used in external files, as long as the code is called from a component.

[Proof of concept](https://svelte.dev/repl/1cc1ff0aa32e457388149262f9453670?version=3.46.4) (Toggle checkbox and see console drawer for lifecycle outputs)

I would propose to convert the component into:
1. A function that creates a readable store that returns the current breakpoint size.
2. An export (e.g. `breakpoints`) that contains the object from the `breakpoints` prop.

This would allow changing the above example to something like this:

```svelte
<script>
	import { breakpointObserver, breakpoints } from 'carbon-components-svelte';

	const breakpoint = breakpointObserver();
	$: tooSmall = breakpoints[$breakpoint] < breakpoints.lg;
</script>
```

The returned store could also expose utility functions that create derived stores for checking smaller/larger. This would simplify the code even further:

```svelte
<script>
	import { breakpointObserver } from 'carbon-components-svelte';

	const tooSmall = breakpointObserver().smallerThan('lg');
</script>
```

metonym:
This is great, I love the idea.

Does it have to replace the existing component? I think it can exist separately.

brunnerh:
I guess it would not hurt to have both if the core logic is extracted, thus not causing additional maintenance or potential for inconsistencies.

metonym:
Released in [v0.61.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.61.0)

brunnerh:
[...]

PR #1092, opened 2022-02-12
Implement breakpointObserver.

brunnerh:
For #1069

Drop breakpoints constant export from Breakpoint component.
Expose breakpoints constant separately and freeze it.
Change `match` event to `change`, because it is now based on store and captures initial value change.

Naming and file organization is up for debate of course as this has no precedent in this repository...

### DIFF
--- a/src/Breakpoint/index.d.ts
+++ b/src/Breakpoint/index.d.ts
@@ -1,2 +1,3 @@
+export { default as Breakpoint } from "./Breakpoint.svelte";
 export { breakpointObserver } from "./breakpointObserver";
-export { breakpoints } from "./constants";
+export { breakpoints } from "./breakpoints";
--- a/src/Breakpoint/Breakpoint.svelte
+++ b/src/Breakpoint/Breakpoint.svelte
@@ -26,7 +26,7 @@
 
   import { createEventDispatcher } from "svelte";
   import { breakpointObserver } from "./breakpointObserver";
-  import { breakpoints } from "./constants";
+  import { breakpoints } from "./breakpoints";
 
   const dispatch = createEventDispatcher();
   const observer = breakpointObserver();
--- a/src/Breakpoint/breakpointObserver.d.ts
+++ b/src/Breakpoint/breakpointObserver.d.ts
@@ -1,19 +1,22 @@
+import type { Readable, Subscriber, Unsubscriber } from "svelte/store";
+import type { BreakpointSize, BreakpointValue } from "./breakpoints";
+
 /**
- * Creates a readable store that returns the current {@link BreakpointSize}.
+ * Creates a readable store that returns the current breakpoint size.
  * It also provides functions for creating derived stores used to do comparisons.
  */
 export function breakpointObserver(): {
-    subscribe: (this: void, run: import("svelte/store").Subscriber<any>, invalidate?: (value?: any) => void) => import("svelte/store").Unsubscriber;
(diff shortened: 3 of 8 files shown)


brunnerh:
@metonym If you have suggestions on how to proceed from here, that would be appreciated. Everything except documentation seems to be tied to automated processes which are not set up to handle this.

metonym (changes requested):
This is looking good.

I see what you mean about the preprocessor breaking and inability to generate TypeScript definitions for the observer/constants.

1. As a workaround, let's remove the newly added exports from `index.js` so that direct importing works.
2. I'd prefer the `BreakpointSize` and `BreakpointValue` interfaces be preserved in the Svelte file/TypeScript definitions. This is useful because it shows the possible values for `BreakpointSize` in the component index and docs.
3. Manual TypeScript definitions for `constants.js` and `breakpointObserver.js` should work for now.

brunnerh:
Ideally the types could be sourced from a single definition. Maybe additional syntax could be introduced so the types can be re-exported from a `.d.ts` file or another module without duplication.

Removing the index export is not great. I tried to mitigate this by adding a separate build output, so the observer could be imported from the `lib` directory (to keep the separation of potentially internal code and public package API), but ran into trouble with the Svelte output (`'Function called outside component initialization'`).

metonym (changes requested):
Nits:

1. rename "constants.js" to "breakpoints.js"

I was expecting that "breakpoints" would be in a file of the same name so that I could get TS definitions when importing directing:

```
import { breakpoints } from "carbon-components-svelte/src/Breakpoint/breakpoints";
```

It's also more descriptive than "constants."

2. I'd prefer to avoid reference types where possible

It's easier to find `BreakpointSize` and `BreakpointValue` types if they are in `breakpoints.d.ts`

`breakpoints.d.ts` should house the types:

```ts
export type BreakpointSize = "sm" | "md" | "lg" | "xlg" | "max";
// ...
```

And then breakpointObserver.js can access it using import types:

```js
/** @type {Record<import("./breakpoints").BreakpointSize, MediaQueryList>} */
```

Finally, in `breakpointObserver.d.ts`, I'd prefer the types be imported at the top:

```ts
import type { Readable, Subscriber, Unsubscriber } from "svelte/store";
import type { BreakpointSize, BreakpointValue } from "./breakpoints";
```

src/Breakpoint/index.d.ts
```diff
@@ -0,0 +1,2 @@
+export { breakpointObserver } from "./breakpointObserver";
+export { breakpoints } from "./constants";
```
metonym, line 2: Missing the `Breakpoint` Svelte component.

```suggestion
export { default as Breakpoint } from "./Breakpoint.svelte";
export { breakpointObserver } from "./breakpointObserver";
export { breakpoints } from "./constants";

```

brunnerh:
> I was expecting that "breakpoints" would be in a file of the same name so that I could get TS definitions when importing...

To my knowledge, the naming should not matter, as long as the files have their respective typings; I renamed it though.

Regarding the inline use of `import("./breakpoints").BreakpointSize` instead of `<reference ...`, personally I am not a fan because of its verbosity and duplication, but then again, I would write everything directly in TypeScript, where that could just be a regular import statement.

The component export missing along with having all those inline Svelte imports was a side effect of using `tsc` to generate the declaration files without giving it enough context. Now that the `BreakpointSize`/`BreakpointValue` types are part of one of these files that is probably no longer an option if anything changes. It would be nice if all that would just be part of the build process, though.

metonym (approved):
Nice work – thanks for also updating the docs.

PR #1100, opened 2022-02-16
fix(multi-select): fix keyboard behavior

### LINKED ISSUE #938: MultiSelect intercepts keyboard events, interfering with browser
When a MultiSelect has focus, it eats keyboard events. This makes it impossible to switch to a different tab (with Cmd+1, Cmd+2, etc.), or open Chrome dev tools (Cmd+Shift+C or F12), and so on.

Possibly related: clicking outside the MultiSelect (e.g., on a blank part of the page) has no effect. Compare this to a Search bar, for instance, where it is still possible to use keyboard shortcuts, and clicking away cancels the search.

Tested with Chrome on a Mac.

metonym:
Fixes #938

- remove `preventDefault|stopPropagation` from `ListBoxField` forwarded keydown event
- pressing "Escape" should close an open menu
- pressing "Enter" should toggle the menu
- pressing "Space" should select an item

### DIFF
--- a/src/ListBox/ListBoxField.svelte
+++ b/src/ListBox/ListBoxField.svelte
@@ -60,7 +60,7 @@
   on:mouseover
   on:mouseenter
   on:mouseleave
-  on:keydown|preventDefault|stopPropagation
+  on:keydown
   on:focus
   on:blur
 >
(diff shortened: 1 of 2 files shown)


metonym:
Closed this because I was experiencing regressions. Replaced by #1101 and will record these as separate issues.

PR #1105, opened 2022-02-18
chore(examples):remove config.kit.target from sveltekit example

rdela:
to prevent this error that halts dev or build

> config.kit.target is no longer required, and should be removed


metonym (approved):
Thanks!

PR #1108, opened 2022-02-19
Update PasswordInput.svelte

hnzlmnn:
If no labelText is given, label should not be rendered. This behavior is present in TextInput.svelte, but not in PasswordInput.svelte


metonym (approved):
Good catch – thanks!

metonym:
Fixed in [v0.58.3](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.58.3)

PR #1110, opened 2022-02-20
fix: slottable labelText should render correctly

metonym:
Currently, in `TextInput` and `PasswordInput`, the "labelText" slot will not be rendered as the condition only checks for the `labelText` prop.

```svelte
<TextInput placeholder="Enter user name...">
  <!-- doesn't render -->
  <b slot="labelText" style="color: red">Label</b>
</TextInput>
```

**Fixes**

- correctly render "labelText" slot in `TextInput` and `PasswordInput`

### DIFF
--- a/src/TextInput/PasswordInput.svelte
+++ b/src/TextInput/PasswordInput.svelte
@@ -129,7 +129,7 @@
       </div>
     {/if}
   {/if}
-  {#if !inline && labelText}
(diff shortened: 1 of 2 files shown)


metonym:
Fixed in [v0.58.3](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.58.3)

PR #1122, opened 2022-02-21
fix(file-uploader): update `clearFiles` accessor description

metonym:
The `clearFiles` prop description currently reads "Override the default behavior of clearing the array of uploaded files."

This is misleading as `clearFiles` is a component accessor, not a `let` prop.


metonym:

PR #1123, opened 2022-02-21
refactor(types): remove unnecessary JSDoc default type

metonym:
The JSDoc default type (e.g., `[size="xl"]`) can be omitted as it is already determined by sveld.

```js
/*
 * @type {"sm" | "lg" | "xl"} [size="xl"]
 */
```


metonym:

PR #1124, opened 2022-02-21
fix(types): dispatched events without a detail value should be `null`

metonym:
Currently, dispatched events without `e.detail` have a type of `any`.

We can be more specific as a `CustomEvent` detail value defaults to `null`.


metonym:

PR #1125, opened 2022-02-21
feat(file-uploader-drop-container): rework `FileUploaderDropContainer`

metonym:
**Features**

- convert `FileList` to `File[]` to be consistent with `FileUploader` and `FileUploaderButton`
- add `files` prop for two-way binding
- dispatch instead of forward the `change` event (detail signature: `File[]`)

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -1328,18 +1328,19 @@ None.
 
 ### Props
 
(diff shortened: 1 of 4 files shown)


metonym:
Released in [v0.59.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.59.0)

PR #1149, opened 2022-03-08
Fix type for validateFiles prop on FileUploaderDropContainer

DanielPower:
`validateFiles` gets called with an array of files. But the type declaration incorrectly indicates that it takes a single file.


metonym (changes requested):
Nice catch.

The TypeScript/JSON/Markdown are auto-generated.

Could you run "yarn build:docs" locally?

DanielPower:
Docs built :slightly_smiling_face:

metonym (approved):
Thanks!

metonym:
Released in [v0.61.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.61.0)

Issue #935, opened 2021-12-10
TextInput bind variable set to undefined from code results "undefined" text in input field


patoi:
Example:

https://svelte.dev/repl/130500ff57c1414ba4c0acfae1dfd42f?version=3.44.2

![image](https://user-images.githubusercontent.com/1846548/145528777-a13d36d6-abaf-40a8-9d4a-d2d7bbf63420.png)

metonym:
Released in [v0.50.3](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.50.3)

PR #1169, opened 2022-03-13
refactor(text-area): use native binding for `value` prop

metonym:
Tested to make sure #935 didn't regress.

As expected, setting `value` to undefined will empty the `textarea` value, not set it to be "undefined."

```svelte
<script>
  import { Button, TextArea } from "carbon-components-svelte";

  let value;
</script>

<TextArea bind:value />

<Button on:click={() => (value = undefined)}>Click</Button>
```


metonym:

PR #1170, opened 2022-03-13
breaking(form): do not prevent default behavior on submit event

### LINKED ISSUE #1134: Form submit not triggered / forwarded
Hello,

https://carbon-components-svelte.onrender.com/components/Form
I looked the example above, and form submit is not triggered when submit Button is clicked.

But when I do this then it works:

```
<Form 
        id="myform"
	on:submit={() => {
		document.getElementById('myform').submit();
	}}
...
<FormGroup legendText="Checkboxes">
    <Checkbox id="checkbox-0" labelText="Checkbox Label" checked name="checkbox-0" /> // <--- added name attribute
    <Checkbox id="checkbox-1" labelText="Checkbox Label" name="checkbox-1" />
    <Checkbox id="checkbox-2" labelText="Checkbox Label" disabled name="checkbox-2"/>
</FormGroup>
...
</Form>
```

Question: is this the proper way? 

Best Regards,
Krešimir

### LINKED ISSUE #1140: Should `Form` prevent the default submit behavior?
Related #1134

Currently, the `submit` event forwarded in `Form` has a `preventDefault` modifier. This can be restrictive as there is no way for the consumer to *not* prevent the default behavior.

I wonder if this assumes too much by "forcing" a SPA-like behavior.

It might be better to leave the decision of preventing the default behavior to the consumer.

metonym:
Closes #1140
Closes #1134

The modifier blocks the consumer from using the native form submission behavior. The consumer still has the ability to prevent the default behavior if needed.

```svelte
<Form
  on:submit={(e) => {
    e.preventDefault();
  }}
/>
```

### DIFF
--- a/docs/src/pages/components/Form.svx
+++ b/docs/src/pages/components/Form.svx
@@ -60,3 +60,17 @@ components: ["Form", "FormGroup"]
   </FormGroup>
   <Button type="submit">Submit</Button>
 </Form>
+
+### Prevent default behavior
+
+The forwarded `submit` event is not modified. Use `e.preventDefault()` to prevent the native form submission behavior.
+
+```
+<Form on:submit={e => {
+  e.preventDefault();
+  console.log("submit", e);
+}}>
+   <Checkbox id="checkbox-0" labelText="Checkbox Label" checked />
+  <Button type="submit">Submit</Button>
+</Form>
+```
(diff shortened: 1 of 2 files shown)


metonym:
Released in [v0.62.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.62.0).

PR #1173, opened 2022-03-14
refactor(search): use native binding for `value` prop

metonym:
- use the native Svelte `bind:` directive for the value prop
- remove the exported `type` prop as Svelte disallows `bind:value` if `type` is dynamic (this isn't a breaking change as `$$restProps` is forwarded to `input`)
- type the `value` prop as `any`


metonym:

PR #1174, opened 2022-03-14
breaking: delete `Copy` component

### LINKED ISSUE #1156: Copy component does not copy and is also missing from the docs
[Copy](https://github.com/carbon-design-system/carbon-components-svelte/tree/master/src/Copy) is exported and available to use, but it's not actually documented. Running `yarn build:docs` doesn't seem to create any docs either.

`Copy` is also a poor name since it doesn't actually have any of the copying logic. It appears to be a component that provides feedback via a popover (not the popover component, to be clear) when clicked. `CopyButton` uses `Copy` as a wrapper and provides the actual [copying logic here](https://github.com/carbon-design-system/carbon-components-svelte/blob/853942f998b656593afc8801970489aa10369ac9/src/CopyButton/CopyButton.svelte#L17).

metonym:
Closes #1156

The `Copy` component is used internally by `CodeSnippet` and `CopyButton`. It is not meant to be used or exposed to the end user. Having it exist in auto-generated documentation can be misleading.

Releasing this breaking change also requires an update to `carbon-preprocess-svelte`.


metonym:
Released in [v0.62.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.62.0).

PR #1179, opened 2022-03-19
feat(toolbar-search): support auto-filterable rows

### LINKED ISSUE #591: DataTable Toolbar Search not working
I'm writing Svelte app that requires table row filtering and Carbon `DataTable` with `ToolbarSearch` looked very promising.

```svelte
<script>
  import {
    DataTable,
    Toolbar,
    ToolbarContent,
    ToolbarSearch
  } from "carbon-components-svelte";

  const headers = [
  ];

  const rows = [
    { id: "a", name: "Load Balancer 3", port: 3000, rule: "Round robin" },
    { id: "b", name: "Load Balancer 1", port: 443, rule: "Round robin" },
    { id: "c", name: "Load Balancer 2", port: 80, rule: "DNS delegation" },
    { id: "d", name: "Load Balancer 6", port: 3000, rule: "Round robin" },
    { id: "e", name: "Load Balancer 4", port: 443, rule: "Round robin" },
    { id: "f", name: "Load Balancer 5", port: 80, rule: "DNS delegation" }
  ];
</script>

<DataTable {headers} {rows} >
  <Toolbar>
    <ToolbarContent>
      <ToolbarSearch />
    </ToolbarContent>
  </Toolbar>
</DataTable>
```

Unfortunately `ToolbarSearch` is not working - is this broken or not implemented? And if not implemented when is it coming?

metonym:
Closes #591

Currently, `ToolbarSearch` does not filter `rows` provided to the `DataTable`. This PR adds a `shouldFilterRows` prop that supports auto-filtering in addition to custom filtering. The use case is to support client-side, in-memory search (e.g., a table with fewer than 500 rows).

By default ,`shouldFilterRows` is `false`.

If  `shouldFilterRows` is `true`, it will use `value`  to perform a basic string comparison on cell values that are of a string or a number type.

```svelte
<ToolbarSearch shouldFilterRows />
```

For custom filtering, `shouldFilterRows` accepts a function and passes it the current row and value. It expects the function to return a boolean.

```svelte
<ToolbarSearch
  shouldFilterRows={(row, value) => {
    // only test the value against a specific cell
    return row.port.includes(value);
  }}
/>
```

Technically, there is a breaking change, which is flagged when generating documentation.

`DataTable` rows will no longer be reactive because it is written to a Svelte store and passed down via the `"DataTable"` context.

Currently, if you `bind:rows`, an [extra field is written to each row](https://github.com/carbon-design-system/carbon-components-svelte/pull/1179/files#diff-e1e14fa75f6c4b3da71cb7eff121f0ce58ba5f7d37999df963c5bb55c90c1002L198) called "cells," which is used when rendering the data table. However, this is an acceptable change as `row.cells` should not be exposed in the first place.

---

TODO:

- [ ] document internal row keys (`id,` `cells`)
- [x] add basic filterable example (`shouldFilter`)
- [x] add custom filterable example


metonym:

PR #1181, opened 2022-03-19
feat(combo-box): make `ComboBox` slottable 

### LINKED ISSUE #1176: Multiline ComboBox
There's currently no component that support multiline content in a combobox / dropdown. I couldn't find anything pointing in this direction on https://www.carbondesignsystem.com either.

A potential solution could be to introduce a variant for the `ComboBox` component. I played around with it, and the following would roughly cover my current use case (being a 3-line item entry):

```diff
diff --git a/node_modules/carbon-components-svelte/src/ComboBox/ComboBox.svelte b/node_modules/carbon-components-svelte/src/ComboBox/ComboBox.svelte
index be693f3..f92fcae 100644
--- a/node_modules/carbon-components-svelte/src/ComboBox/ComboBox.svelte
+++ b/node_modules/carbon-components-svelte/src/ComboBox/ComboBox.svelte
@@ -363,7 +363,7 @@
               highlightedIndex = i;
             }}"
           >
-            {itemToString(item)}
+            {@html itemToString(item)}
             {#if selectedItem && selectedItem.id === item.id}
               <Checkmark16 class="bx--list-box__menu-item__selected-icon" />
             {/if}
```

It also requires some additional css, but that doesn't necessarily have to live in [carbon-components-svelte](https://github.com/carbon-design-system/carbon-components-svelte).

```scss
.bx--list-box__menu-item,
.bx--list-box__menu-item__option {
  // carbon-components/scss/globals/scss/vendor/@carbon/styles/scss/components/list-box/_list-box.scss
  height: rem(2 * 40px);
  line-height: 1.2rem;
}
```

The result is such we can have something like this:
image

The minimal change in [carbon-components-svelte](https://github.com/carbon-design-system/carbon-components-svelte) to support my use case would be to allow `@html` as the item content.

What do you think about something like this? Do you have any other plans? Is there another way to achieve this?

metonym:
Closes #1176

This PR adds a default slot to `ComboBox` so the consumer can customize each item slot using their own components or elements. This is more expressive than the `itemToString` prop.

The default slot includes the `item` and `index` as props.

```svelte
<ComboBox
  titleText="Contact"
  items={[
    { id: "0", text: "Slack" },
    { id: "1", text: "Email" },
    { id: "2", text: "Fax" },
  ]}
  let:item
  let:index
>
  {item.text} {item.id} {index}
</ComboBox>
```

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -686,7 +686,9 @@ export interface ComboBoxItem {
 
 ### Slots
 
-None.
+| Slot name | Default | Props                                               | Fallback                          |
+| :-------- | :------ | :-------------------------------------------------- | :-------------------------------- |
+| --        | Yes     | <code>{ item: ComboBoxItem; index: number } </code> | <code>{itemToString(item)}</code> |
 
 ### Events
 
--- a/docs/src/COMPONENT_API.json
+++ b/docs/src/COMPONENT_API.json
@@ -1639,7 +1639,14 @@
         }
       ],
       "moduleExports": [],
-      "slots": [],
+      "slots": [
+        {
+          "name": "__default__",
+          "default": true,
+          "fallback": "{itemToString(item)}",
+          "slot_props": "{ item: ComboBoxItem; index: number }"
+        }
+      ],
       "events": [
         {
           "type": "dispatched",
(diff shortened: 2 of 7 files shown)


metonym:

PR #1186, opened 2022-03-19
feat: make icons slottable

metonym:
Currently, icons are not slottable. Making icons slottable would support make it easier to pass custom styles and events. The workaround currently would be to create a "wrapper" component, which is rather inconvenient.

```svelte
<!-- icon as prop -->
<Tag icon={Add16}>Text</Tag>

<!-- slottable icon -->
<Tag>
  Text
  <Add16 slot="icon" style="fill: red" on:click />
</Tag>
```

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -860,11 +860,12 @@ None.
 
 ### Slots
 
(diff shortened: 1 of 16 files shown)


metonym:

PR #1187, opened 2022-03-19
feat(ui-shell): make `SideNavLink` text slottable

metonym:
This adds a default slot to `SideNavLink` so that `text` is slottable.

```svelte
<SideNavLink href="/" text="Text" />

<SideNavLink href="/">
  <strong>Text</strong>
</SideNavLink>
```


metonym:

PR #1198, opened 2022-03-23
Allow flatpickrProps to override default flatpickr options

### LINKED ISSUE #1197: flatpickrProps on DatePicker do not override default options
I wanted to be able to set `disableMobile: false` on [flatpickr options](https://flatpickr.js.org/options/) for my DatePicker component. I used the `flatpickrProps` property like so:

```
<DatePicker datePickerType="single" flatpickrProps={{disableMobile: false}}>
  <DatePickerInput labelText="..." placeholder="mm/dd/yyyy" />
</DatePicker>
```

These options are properly forwarded to `createCalendar.js`, but the [options are spread](https://github.com/carbon-design-system/carbon-components-svelte/blob/912d3796f0b011648a055ed14d30b765100bb1b1/src/DatePicker/createCalendar.js#L71) before applying the default options from `carbon-components-svelte`. In order to be able to override the defaults from this library, the `...options` statement needs to be placed below the default options.

flut1:
Fixes #1197

### DIFF
--- a/src/DatePicker/createCalendar.js
+++ b/src/DatePicker/createCalendar.js
@@ -68,7 +68,6 @@ async function createCalendar({ options, base, input, dispatch }) {
   }
 
   return new flatpickr(base, {
-    ...options,
     allowInput: true,
     disableMobile: true,
     clickOpens: true,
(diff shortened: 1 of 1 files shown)


metonym (approved):
Thank you!

PR #1203, opened 2022-03-24
feat: add WebStorage to address localStorage or sessionStorage

wilriker:
feat: make LocalStorage use WebStorage
feat: add SessionStorage using WebStorage
feat: make key in WebStorage reactive
feat: add updateKey event in case key changed


wilriker:
This fixes #1204 and expands WebStorage API availability to user's choice which one to use (session or local).

PR #1205, opened 2022-03-24
fix(ComboBox): adding title prop to ListBoxMenuItem

Borog:
This PR add title prop to ListBoxMenuItem.

Design documentation:
"Avoid having multiple lines of text in a dropdown. If the text is too long for one line, add an ellipsis (…) for overflow content, _**and accompany with a browser-based tooltip to show the full string of text.**_"
https://www.carbondesignsystem.com/components/dropdown/usage/#overflow-content

### DIFF
--- a/src/ComboBox/ComboBox.svelte
+++ b/src/ComboBox/ComboBox.svelte
@@ -350,6 +350,7 @@
         {#each filteredItems as item, i (item.id)}
           <ListBoxMenuItem
             id="{item.id}"
+            title="{itemToString(item)}"
             active="{selectedId === item.id}"
             highlighted="{highlightedIndex === i}"
             on:click="{() => {


metonym:
I like this a lot.

It seems that Carbon conditionally adds a [title only for truncated text](https://github.com/carbon-design-system/carbon/blob/008a7df0a28d95e608dfcfbc466544ef725d4c3f/packages/react/src/components/ListBox/ListBoxMenuItem.js#L13). They have a hook that determines if it's truncated based on if the scrollWidth exceeds the offsetWidth. I think it would be nice to only add the title if the text is truncated.

What do you think?

metonym:
Released in [v0.64.3](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.64.3)

PR #1213, opened 2022-03-26
Align 10.56

metonym:
**Features**

- add `kind` prop to `ProgressBar` to support inline, indented variants

**Documentation**

- add `ProgressBar` examples "Inline variant" and "Indented variant"

**Housekeeping**

- upgrade `carbon-components` to v10.56.0

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -2788,15 +2788,16 @@ None.
 
 ### Props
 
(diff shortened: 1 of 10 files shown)


metonym:
Released in [v0.63.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.63.0). If using `carbon-preprocess-svelte`, upgrade to version 0.9.0.

PR #1230, opened 2022-04-05
fix(data-table): prevent selectable datatable with sticky header from jumping

metonym:
This PR fixes a bug where selecting a datatable row will cause the page to scroll back to the top. It applies a fix suggested in #1230.

Sample code to repro:

1. Scroll down the page
2. Select a row
3. The page should scroll to the top

```svelte
<script>

  const headers = [
  ];

  const rows = [
    { id: "a", name: "Load Balancer 3", port: 3000, rule: "Round robin" },
    { id: "b", name: "Load Balancer 1", port: 443, rule: "Round robin" },
    { id: "c", name: "Load Balancer 2", port: 80, rule: "DNS delegation" },
    { id: "d", name: "Load Balancer 6", port: 3000, rule: "Round robin" },
    { id: "e", name: "Load Balancer 4", port: 443, rule: "Round robin" },
    { id: "f", name: "Load Balancer 5", port: 80, rule: "DNS delegation" },
  ];
</script>

<div style="min-height: 200vh">
  <DataTable stickyHeader batchSelection {headers} {rows} />
</div>

```

### DIFF
--- a/src/Checkbox/InlineCheckbox.svelte
+++ b/src/Checkbox/InlineCheckbox.svelte
@@ -18,20 +18,22 @@
   export let ref = null;
 </script>
 
-<input
-  bind:this="{ref}"
-  type="checkbox"
-  class:bx--checkbox="{true}"
-  checked="{indeterminate ? false : checked}"
-  indeterminate="{indeterminate}"
-  id="{id}"
-  {...$$restProps}
-  aria-label="{undefined}"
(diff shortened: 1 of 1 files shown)


metonym:
Released in [v0.63.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.63.0).

Note on upgrading: if using `carbon-preprocess-svelte`, upgrade to version 0.9.0.

PR #1265, opened 2022-04-23
feat(data-table): allow custom column widths

### LINKED ISSUE #790: Feature Request: Set the width of a column in DataTable
Hello,
It would be nice if we can set the width of a column when setting the headers in `DataTable`. For example:
```
  headers={[
    { key: 'name', value: 'Name' },
    { key: 'protocol', value: 'Protocol', width='15%' },
    { key: 'port', value: 'Port', width='10rem' },
    { key: 'rule', value: 'Rule', minWidth='10rem' }
  ]}
```

metonym:
Closes #790

Specify an optional `width` or `minWidth` property in the `headers` object to control the width of the column.

In the example below, we set the "Name" column width to be 50%, with a minimum width of 200px.

Note that a [table-layout: fixed](https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout#values) rule will be applied to the `table` element when using custom widths.

```svelte
<DataTable
  headers="{[
    { key: 'name', value: 'Name', width: '50%', minWidth: '200px' },
    { key: 'protocol', value: 'Protocol', width: '60px' },
    { key: 'port', value: 'Port', width: '60px' },
    { key: 'rule', value: 'Rule', width: '10rem' },
  ]}"
  {rows}
/>
```

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -927,8 +927,8 @@ export interface DataTableEmptyHeader {
   display?: (item: Value) => DataTableValue;
   sort?: false | ((a: DataTableValue, b: DataTableValue) => 0 | -1 | 1);
   columnMenu?: boolean;
-  minWidth?: string;
   width?: string;
+  minWidth?: string;
 }
 
 export interface DataTableNonEmptyHeader {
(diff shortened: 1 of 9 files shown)


ibilux:
Thank you @metonym for the amazing work.

PR #1292, opened 2022-05-13
fix(code-snippet): use "small" size for expand/collapse button

metonym:
The expand/collapse button for multi-line code snippets should use the "small" size, not the "field" size.

Ref: https://github.com/carbon-design-system/carbon/pull/11353


metonym:

PR #1293, opened 2022-05-13
fix(header-action): type missing "open" event

metonym:
`HeaderAction` typedefs should include the dispatched "open" event.


metonym:

PR #1295, opened 2022-05-13
fix(types): narrow dispatched events without detail to use `null` instead of `any`

metonym:
Currently, dispatched events without a detail type are typed as `CustomEvent<any>`. Although this isn't wrong, we can narrow the `e.detail` type to be `null`.


metonym:

PR #1299, opened 2022-05-14
fix(types): use `@see` tag for flatpickr options link

metonym:
Using the JSDoc `@see` tag will give better Intellisense in VSCode when hovering over the `flatpickrProps` prop.

Screen Shot 2022-05-14 at 12 40 57 PM


metonym:

PR #1317, opened 2022-05-28
feat: set title attribute in `CheckBox`, `MultiSelect`, `Dropdown`, `ComboBox`  if label is truncated

metonym:
Closes #1205

The `title` attribute should be set if an item's label text is truncated.

This implementation using the `element.innerText` method since the text may not be provided via a prop (e.g., slots).

Screen Shot 2022-05-28 at 10 06 18 AM


metonym:
Making a note for versioning that this should be a fix, not a feature.

PR #1318, opened 2022-05-28
refactor: prefer class directive over class attribute

metonym:
This PR refactors components to use the [class directive](https://svelte.dev/docs#template-syntax-element-directives-class-name) where possible to make conditionally applied classes easier to read.

**Before**

```svelte
class="{datePickerType && `bx--date-picker--${datePickerType}`} {datePickerType === 'range' && $labelTextEmpty && 'bx--date-picker--nolabel'}"
```

**After**

```svelte
class:bx--date-picker--simple="{datePickerType === 'simple'}"
class:bx--date-picker--single="{datePickerType === 'single'}"
class:bx--date-picker--range="{datePickerType === 'range'}"
class:bx--date-picker--nolabel="{datePickerType === 'range' && $labelTextEmpty}"
```

### DIFF
--- a/src/TextInput/TextInput.svelte
+++ b/src/TextInput/TextInput.svelte
@@ -120,7 +120,8 @@
           class:bx--visually-hidden="{hideLabel}"
           class:bx--label--disabled="{disabled}"
           class:bx--label--inline="{inline}"
-          class="{inline && !!size && `bx--label--inline--${size}`}"
+          class:bx--label--inline--sm="{size === 'sm'}"
+          class:bx--label--inline--xl="{size === 'xl'}"
         >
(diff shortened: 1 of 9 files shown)


src/TextInput/TextInput.svelte
```diff
@@ -120,7 +120,8 @@
           class:bx--visually-hidden="{hideLabel}"
           class:bx--label--disabled="{disabled}"
           class:bx--label--inline="{inline}"
-          class="{inline && !!size && `bx--label--inline--${size}`}"
```
metonym, line 123: The `inline` condition isn't necessary since this element is already contained in an `inline` if block.

PR #1329, opened 2022-06-02
breaking(dropdown): remove useless `inline` prop

metonym:
`Dropdown` supports an inline variant by setting `type="inline"`. The exported `inline` prop doesn't control this value. This PR removes the redundant `inline` prop.

Although this is technically a breaking change, it should have no user impact as `<Dropdown inline ... />` would not have worked anyway.

A broader discussion pertains to prop naming in general. Some components have `inline` as a boolean, while others have it as an enum for the `type` prop. I actually do like "inline" over a two value enum (e.g., "default" | "inline"). Also, the `type` prop name is problematic as it can clash with the native attribute. A future PR could explore making prop names follow one convention.


metonym:
This should be categorized as a fix, as the `inline` prop did not work in the first place.

PR #1337, opened 2022-06-05
feat(data-table): support programmatic sorting

metonym:
This PR adds the `sortKey` and `sortDirection` props to allow programmatic sorting.

This also allows the user to set a "default" sorted table without sorting the `rows` manually.

**Usage**

By default, the table is not sorted by a specific key. The `sortKey` value must be a valid `key` specified in the `headers` object.

Possible values for `sortDirection` include `"none"` or `"ascending"` or `"descending"`.

Setting `sortKey` to `null` and `sortDirection` to `"none"` should reset the table rows to their initial order.

```svelte
<DataTable
  sortable
  sortKey="name"
  sortDirection="descending"
  headers="{headers}"
  rows="{rows}"
/>
```

**Notes**

This refactored `DataTable` and `TableHeader` to avoid using a Svelte store to maintain state. Instead of passing a store down using `setContext`, sorting-related props are passed down directly.

As a result, this caused a minor breaking change where `disableSorting` in `TableHeader` has been removed.


metonym:
Released in [v0.65.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.65.0)

chilligerchief:
This feature changed the sorting behavior, is this intended?
Previously:
If you click on a column, you initially sort ascending, clicking again cylces to descending and none.
For example:
Click Name -> Sort by Name ASC
Click Port -> Sort by Port ASC
Click Port -> Sort by Port DESC
Click Name -> Sort by Name ASC


Now:
If you click on a column, the sortDirection of the previous sorted column gets cycled.
For example:
Click name -> Sort by Name ASC
Click Port -> Sort By Port **DESC**
Click Name-> No Sorting

I think the old behavior is more intuitive.

metonym:
@chilligerchief This is definitely an unintended regression. Tracking in #1341.

PR #1339, opened 2022-06-06
Simplify open modal tracking code

brunnerh:
- Class application:
  - Can be simplified using `toggle`.
  - Can be extracted, so it does not need to be duplicated across components.
- On destroy cleanup can be extracted and unified with open/close logic.
- `ComposedModal` ID can be removed again.

Addresses comments on [`22f93ee`](https://github.com/carbon-design-system/carbon-components-svelte/commit/22f93ee675c04e96a3119fdd76eaa273614891cd)

### DIFF
--- a/src/ComposedModal/ComposedModal.svelte
+++ b/src/ComposedModal/ComposedModal.svelte
@@ -96,11 +96,7 @@
       dispatch("open");
     }
 
-    if ($modalsOpen.length > 0) {
-      document.body.classList.add("bx--body--with-modal-open");
-    } else {
-      document.body.classList.remove("bx--body--with-modal-open");
-    }
+    document.body.classList.toggle("bx--body--with-modal-open", $modalsOpen.length > 0);
   });
 
   $: if (open) {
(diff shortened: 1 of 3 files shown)


brunnerh:
@metonym Simplifying the code for the store to just a count was not very robust, so this still tracks more (an "open" store instead of IDs), but the code is now in one location and should be overall more simple.

Used this for testing, the `mount` is used to check if the destroy logic behaves as expected:
```svelte
<script lang="ts">
	import { Modal, Button } from 'carbon-components-svelte';

	let mount = true;
	let open = false;
	let open2 = false;
</script>

<Button on:click={() => open = true}>Open outer</Button>
<br />
<label>Mount <input type=checkbox bind:checked={mount}/></label>

{#if mount}
	<Modal modalHeading="Outer" bind:open passiveModal>
		<Button on:click={() => open2 = true}>
			Open Inner
		</Button>
		<br />
		<label>Mount <input type=checkbox bind:checked={mount}/></label>
	</Modal>

	<Modal modalHeading="Inner" bind:open={open2} passiveModal>
		<label>Mount <input type=checkbox bind:checked={mount}/></label>
	</Modal>
{/if}

<div style:height="150vh"/> <!-- Forces scroll -->
```

metonym (approved):
💯

Thanks for DRYing this up.

PR #1344, opened 2022-06-12
fix(time-picker): use native binding for `value` prop

metonym:
Similar to #1343, #1169, #1173, the input should use native binding instead of manually managing state in order to avoid #935.


metonym:
Fixed in [v0.65.2](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.65.2).

PR #1346, opened 2022-06-12
Fix reactivity of thKeys property for sorting headers

### LINKED ISSUE #1345: DataTable sorting not working for headers added after component creation
If a DataTable has its headers modified after initialization, most things will update reactively, but the sorting for any newly added headers will not function.

This is because the sort process depends on the header key being present in [`$theKeys` store](https://github.com/carbon-design-system/carbon-components-svelte/blob/b21718ccfd772ce0fb44a6935590c352430cea29/src/DataTable/DataTable.svelte#L359). `$thKeys` is a [derived store based on a `headerItems` variable](https://github.com/carbon-design-system/carbon-components-svelte/blob/b21718ccfd772ce0fb44a6935590c352430cea29/src/DataTable/DataTable.svelte#L149).

However... the `headerItems` variable is unused besides this, so `$thKeys` will never update after the first initialization. The first initialization is actually based on the `headers` variable, so the sorting will work correctly if the header being sorted is present in the `headers` array from the beginning of the component creation. This is why sorting generally works.

Some of our codebase has wrapped the DataTable such that the headers are not created until a network request is complete, so this is preventing sorting from working for our scenario.

I believe the intention is to have `thKeys` be reactive based on changes to `headers`, so I have created a pull request with this fix. The fix is working for me, but if there is another more desirable way to fix this issue, I'm fine with another direction being taken.

Thanks for the hard work on the project!

jorgebv:
Fixes sorting for headers added after component creation. Fixes #1345

### DIFF
--- a/src/DataTable/DataTable.svelte
+++ b/src/DataTable/DataTable.svelte
@@ -144,13 +144,10 @@
   };
   const dispatch = createEventDispatcher();
   const batchSelectedIds = writable(false);
-  const headerItems = writable([]);
   const tableRows = writable(rows);
-  const thKeys = derived(headerItems, () =>
-    headers
-      .map(({ key }) => ({ key, id: key }))
-      .reduce((a, c) => ({ ...a, [c.key]: c.id }), {})
-  );
+  $: thKeys = headers
+    .map(({ key }) => ({ key, id: key }))
+    .reduce((a, c) => ({ ...a, [c.key]: c.id }), {});
   const resolvePath = (object, path) => {
     if (path in object) return object[path];
     return path
@@ -356,7 +353,7 @@
                     sortKey === header.key ? sortDirection : 'none';
                   sortDirection = sortDirectionMap[currentSortDirection];
                   sortKey =
-                    sortDirection === 'none' ? null : $thKeys[header.key];
+                    sortDirection === 'none' ? null : thKeys[header.key];
                   dispatch('click:header', { header, sortDirection });
                 }
               }}"


src/DataTable/DataTable.svelte
```diff
@@ -144,13 +144,10 @@
...
   const batchSelectedIds = writable(false);
-  const headerItems = writable([]);
   const tableRows = writable(rows);
-  const thKeys = derived(headerItems, () =>
-    headers
-      .map(({ key }) => ({ key, id: key }))
-      .reduce((a, c) => ({ ...a, [c.key]: c.id }), {})
-  );
+  $: thKeys = headers
+    .map(({ key }) => ({ key, id: key }))
```
metonym, line 149: Since we're here, let's refactor this to remove the `.map` since we use the `key` as the `id`.

```js
$: thKeys = headers.reduce((a, c) => ({ ...a, [c.key]: c.key }), {});
```

metonym (commented):
Nice catch – really appreciate the PR and clear explanation of the issue.

src/DataTable/DataTable.svelte
jorgebv: Thanks, it makes sense and works fine in my local testing.

metonym (approved):

metonym:
Fixed in [v0.65.2](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.65.2). Thanks again!

PR #1354, opened 2022-06-18
feat(slider): add `fullWidth` prop

metonym:
The `bx--slider` class in the `Slider` component has max-width of `40rem`. It would be nice to set it to have a width of 100%.

Because `$$restProps` is spread to the top-level `div` element in `Slider`, the only way to override this rule is to use global styles:

```css
:global(.bx--slider-container) {
  width: 100%;
}

:global(.bx--slider) {
  max-width: none;
}
```

This PR adds a `fullWidth` prop that sets a full width style using inline style attributes.

Pictured below compares the `Slider` with and without the `fullWidth` property set.

```svelte
<Slider fullWidth />
<Slider />
```

Screen Shot 2022-06-18 at 6 21 28 AM

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -3468,25 +3468,26 @@ None.
 
 ### Props
 
-| Prop name      | Required | Kind             | Reactive | Type                                    | Default value                                    | Description                                |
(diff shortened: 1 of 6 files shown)


metonym:
Released in [v0.66.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.66.0).

PR #1355, opened 2022-06-18
feat(select): support `selected` prop number type

metonym:
This PR enables numbers to be used as values with the `Select` component. As a result, we can also remove logic from `Pagination` needed to coerce the bound `pageSize` and `page` values.

```svelte
<Select
  on:change={(e) => {
    console.log("change", e.detail, typeof e.detail);
  }}
>
  <SelectItem value={0} />
  <SelectItem value={1} />
  <SelectItem value={2} />
  <SelectItem value={3} />
  <SelectItem value={4} />
</Select>
```


metonym:
Released in [v0.66.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.66.0).

brunnerh:
(This is probably a breaking change when using TypeScript.)

PR #1368, opened 2022-06-23
fix(side-nav-menu): expanded `SideNavMenu` should not have a max-height 

metonym:
The Carbon styles apply a max-height of `93.75rem` to the `SideNavMenu`. Although this seems to be a deliberate design choice, I don't think a cap should be set.

This overrides the max-height to be `none` if the menu is expanded.


metonym:

PR #1369, opened 2022-06-23
fix(side-nav-menu): chevron icon should not be focusable

metonym:
The chevron icon in the `SideNavMenu` should not be focusable.

See the recording below for the current behavior:

https://user-images.githubusercontent.com/10718366/175302933-bbe4e8b0-2db6-44ba-9f2f-12d98b12d864.mov


metonym:

PR #1371, opened 2022-06-25
fix(password-input): support invalid state when used in a `FluidForm`

metonym:
Currently, a `PasswordInput` in a `FluidForm` has a broken invalid state.

This fixes the invalid state to align with the upstream implementation.

---

### Before

Screen Shot 2022-06-25 at 6 59 35 AM

### After

Screen Shot 2022-06-25 at 6 58 22 AM


metonym:

PR #1373, opened 2022-06-26
fix(side-nav-menu): remove title attribute from menu icon

metonym:
The chevron icon in the UI Shell `SideNavMenu` icon should not have a title attribute.

Ref: https://github.com/carbon-design-system/carbon/blob/f17b8147b7e66325bbee6bcc6a6180d9aa5ec4d5/packages/react/src/components/UIShell/SideNavMenu.js#L71


metonym:

PR #1406, opened 2022-07-19
fix(side-nav): Remove content left margin when side nav collapsed

### LINKED ISSUE #1145: Hamburger menu adds extra margin on left side
I added a hamburger menu and it's created an extra margin on the left side. My UI now looks weird because the Rows>Tiles (that populate the Content) have a larger margin on the left side compared to the right side (even when menu is not open). (i.e. the margins are unequal)

Is this intended behaviour? If so, then shouldn't it be changed.

Screen Shot 2022-03-05 at 12 16 15 pm

Screen Shot 2022-03-05 at 12 16 35 pm

willbicks:
As described in #1145, when a SideNav is used outside a header, a 3rem margin is applied to the site Content block, even when the side navigation component is collapsed. 

image

This was manually overwritten with a CSS styling rule in the docs site. The purpose of this PR is to bring that change to the component so that it is applied by default, ensuring there is no extraneous left margin on smaller screens.

Fixes #1145

### DIFF
--- a/docs/src/global.css
+++ b/docs/src/global.css
@@ -246,7 +246,6 @@ main.bx--content {
 
 @media (max-width: 1056px) {
   .bx--side-nav ~ .bx--content {
-    margin-left: 0;
     padding-left: 1rem;
     padding-right: 1rem;
   }
--- a/src/UIShell/SideNav.svelte
+++ b/src/UIShell/SideNav.svelte
@@ -74,3 +74,9 @@
 >
   <slot />
 </nav>
+
+<style>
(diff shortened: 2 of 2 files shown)


metonym (commented):
Thanks for opening a PR.

Implementation-wise, I'd prefer avoiding the `global` directive as it's more of an escape hatch.

Instead, I propose that the `SideNav` expanded/collapsed state to the UI Shell [`navStore.js`](https://github.com/carbon-design-system/carbon-components-svelte/blob/master/src/UIShell/navStore.js) and using that value to control a style attribute in `Content.svelte`.

PR #1411, opened 2022-07-20
fix(combobox): filterable and custom label (#1405)

### LINKED ISSUE #1405: Bug in filterable ComboBox: "undefined" instead of custom label
The filterable ComboBox using `shouldFilterItem` doesn't fully support the custom label set by `itemToString` when displaying the selected item in the input field and displays `undefined` instead:

Illustration of the issue

I traced it down to lines like the following (there are multiple occurrences):

https://github.com/carbon-design-system/carbon-components-svelte/blob/f8e400f2edd971f41dacfbc21bfff4ee059e5163/src/ComboBox/ComboBox.svelte#L297

In my opinion, this line should make use of the `itemToString`, e.g., `value = itemToString(filteredItems[highlightedIndex]);` similar to this line:
https://github.com/carbon-design-system/carbon-components-svelte/blob/f8e400f2edd971f41dacfbc21bfff4ee059e5163/src/ComboBox/ComboBox.svelte#L394

Should I create a pull request for this?

hffmr:
Fixes #1405

The ComboBox does not display the custom label set with `itemToString` in the input after a selection, when combined with the filterable option with `shouldFilterItem`.

This is now fixed, so that all direct accesses to the default attribute `text` are replaced with accessing `itemToString` instead.

Also adds the specific use case of combining filter and custom string to the documentation.

### DIFF
--- a/docs/src/pages/components/ComboBox.svx
+++ b/docs/src/pages/components/ComboBox.svx
@@ -54,6 +54,12 @@ Specify `focus: false` in the method options to avoid re-focusing the input.
 
 <FileSource src="/framed/ComboBox/FilterableComboBox" />
 
+### Filterable with custom label
+
+Combine a custom label `itemToString` with the filterable option, e.g., when working with internationalization.
+
+<FileSource src="/framed/ComboBox/FilterableComboBoxCustomLabel" />
+
 ### Top direction
 
 Set `direction` to `"top"` for the combobox dropdown menu to appear above the input.
--- a/docs/src/pages/components/ComboBox.svx
+++ b/docs/src/pages/components/ComboBox.svx
@@ -56,7 +56,7 @@ Specify `focus: false` in the method options to avoid re-focusing the input.
 
 ### Filterable with custom label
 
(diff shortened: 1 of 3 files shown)


metonym (commented):
Excellent work – and thank you for going the extra mile and adding a practical example.

docs/src/pages/components/ComboBox.svx
```diff
@@ -54,6 +54,12 @@ Specify `focus: false` in the method options to avoid re-focusing the input.
 
 <FileSource src="/framed/ComboBox/FilterableComboBox" />
 
+### Filterable with custom label
+
+Combine a custom label `itemToString` with the filterable option, e.g., when working with internationalization.
```
metonym, line 59: Nit: brevity

```suggestion
Combine a custom label `itemToString` with the filterable option (e.g., internationalization).
```

hffmr: @metonym Agreed, thanks!

metonym (approved):

PR #1418, opened 2022-07-29
fix: guard against invalid accesses

wilriker:
Current check does only check for presence of document and assumes the full tree below exists but that is not always the case and can lead to exceptions.

### DIFF
--- a/src/Modal/modalStore.js
+++ b/src/Modal/modalStore.js
@@ -31,6 +31,6 @@ export const trackModal = (openStore) =>
   });
 
 modalsOpen.subscribe((openCount) => {
-  if (typeof document !== "undefined")
+  if (typeof document !== "undefined" && document.body && document.body.classList && typeof document.body.classList.toggle === "function")
     document.body.classList.toggle("bx--body--with-modal-open", openCount > 0);
 });


metonym:
In your case, is the code throwing an exception because your DOM doesn't have a `body` tag?

wilriker:
Exactly, `document.body` is `null` in my case.

I am in slow transition from a crazy proprietary custom web framework towards clean Svelte but that's the state I have to deal with. I am currently stuck on carbon-components 0.64.3 since starting with 0.65.0 I can no longer load the pages and this PR fixes it - or rather it works around it.

metonym:
What does your `index.html` or markup look like? I'm interested in a repro so that I can test this.

When I try to forgo the `body` tag, I get the "[Uncaught Error: 'target' is a required option - Svelte](https://stackoverflow.com/questions/64847693/uncaught-error-target-is-a-required-option-svelte)" error.

If I defer loading of the `script` or wrap it in a `div`, a `body` tag is produced automatically in the DOM.

wilriker:
The `index.html` does have a `body` and if I check DevTools I can both see a `body` tag as well as access `document.body` (and stuff below that).

Maybe the reason here is that this is not (yet) a full Svelte application but we only inject Svelte components created via `new ComponentName` at some points in the existing app. I still have no clue why this would lead to the situation... Oh wait...

I just found the issue: to be able to load Svelte components we added a `<script>` tag to our `index.html` that loads the output of rollup. But we did so inside the `<head>` tag - and at the time this was evaluated the `body` tag did not exist yet. I moved this `<script>` tag now towards the end of the `body` tag and tada, it works. Sorry for the noise, did not expect this to be the reason.

PR #1419, opened 2022-08-03
fix(ComboBox): fix an issue, when selection via typing does not work as expected, resolves #1222

### LINKED ISSUE #1222: ComboBox selection via typing only does not work as expected
E.g. taking the [default example from the docs](https://carbon-components-svelte.onrender.com/components/ComboBox#default):

```svelte
<script>
  import { ComboBox } from "carbon-components-svelte";
</script>

<ComboBox
  titleText="Contact"
  items={[
    { id: "0", text: "Slack" },
    { id: "1", text: "Email" },
    { id: "2", text: "Fax" },
  ]}
/>
```

If one types `Email` and presses Enter, the component will select `Slack` instead. Highlighting and selecting different items only works via arrow keys, which defeats the point of having a text-based input.

The [React implementation](https://react.carbondesignsystem.com/?path=/story/components-combobox--combobox-story) gets this somewhat right:
- When e.g. `Option 2` is typed, the corresponding option is highlighted
- When Enter is pressed the highlighted option is selected

Ideally it should highlight `Option 1` as soon as `Option` is typed, which does not happen. It also hijacks the Home and End keys, which I do not approve of, as those can be useful for editing the text in the field.

Tested on Windows 10 in Chrome and Firefox.

patoi:
resolves #1222

### DIFF
--- a/src/ComboBox/ComboBox.svelte
+++ b/src/ComboBox/ComboBox.svelte
@@ -308,7 +308,7 @@
                 // typed value has matched or fallback to first enabled item
                 open = false;
                 selectedItem = matchedItem;
-                value = itemToString(selectedItem.text);
+                value = itemToString(selectedItem);
                 selectedId = selectedItem.id;
               }
             }
--- a/src/ComboBox/ComboBox.svelte
+++ b/src/ComboBox/ComboBox.svelte
@@ -205,7 +205,9 @@
   $: menuId = `menu-${id}`;
   $: comboId = `combo-${id}`;
   $: highlightedId = items[highlightedIndex] ? items[highlightedIndex].id : 0;
-  $: filteredItems = items.filter((item) => shouldFilterItem(item, value));
+  $: /** @type Array<ComboBoxItem> */ filteredItems = items.filter((item) =>
+    shouldFilterItem(item, value)
+  );
 </script>
 
 <svelte:window
@@ -287,7 +289,6 @@
         on:keydown|stopPropagation="{({ key }) => {
           if (key === 'Enter') {
             open = !open;
-
             if (
               highlightedIndex > -1 &&
               filteredItems[highlightedIndex]?.id !== selectedId
@@ -299,11 +300,33 @@
                 selectedId = filteredItems[highlightedIndex].id;
               }
             } else {
-              open = false;
-              if (filteredItems[0]) {
-                value = itemToString(filteredItems[0]);
-                selectedItem = filteredItems[0];
-                selectedId = filteredItems[0].id;
+              // searching typed value in text list with lowercase
+              /** @type Array<ComboBoxItem> */
+              const listOfItems = filteredItems.map((e) => {
+                return {
+                  id: e.id,
+                  text: e.text?.toLowerCase(),
+                  disabled: e.disabled,
+                };
+              });
+              const matchedItem = listOfItems.find(
+                (e) => e.text === value?.toLowerCase()
+              );
+              if (matchedItem) {
+                // typed value has matched
+                open = false;
+                selectedItem = filteredItems.find(
+                  (e) => e.id === matchedItem.id
+                );
+                value = selectedItem.text;
+                selectedId = selectedItem.id;
+              } else {
+                open = false;
+                if (filteredItems[0]) {
+                  value = itemToString(filteredItems[0]);
+                  selectedItem = filteredItems[0];
+                  selectedId = filteredItems[0].id;
+                }
               }
             }
             highlightedIndex = -1;
--- a/src/ComboBox/ComboBox.svelte
+++ b/src/ComboBox/ComboBox.svelte
@@ -299,25 +299,17 @@
               }
             } else {
               // searching typed value in text list with lowercase
-              const matchedItem = filteredItems.find(
-                (e) =>
-                  e.text.toLowerCase() === value?.toLowerCase() && !e.disabled
-              );
+              const matchedItem =
+                filteredItems.find(
+                  (e) =>
+                    e.text.toLowerCase() === value?.toLowerCase() && !e.disabled
+                ) ?? filteredItems.find((e) => !e.disabled);
               if (matchedItem) {
-                // typed value has matched
+                // typed value has matched or fallback to first enabled item
                 open = false;
                 selectedItem = matchedItem;
                 value = selectedItem.text;
                 selectedId = selectedItem.id;
-              } else {
-                open = false;
-                // find first enabled item
-                const matchedItem = filteredItems.find((e) => !e.disabled);
-                if (matchedItem) {
-                  value = itemToString(matchedItem);
-                  selectedItem = matchedItem;
-                  selectedId = matchedItem.id;
-                }
               }
             }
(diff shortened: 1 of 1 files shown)


patoi:
Bug:

https://user-images.githubusercontent.com/1846548/182596543-5809f505-9078-45c6-8cd9-1591993531be.mp4

Fix: it is not case sensitive.

https://user-images.githubusercontent.com/1846548/182596593-13d95ccb-8395-4557-b733-a10d5366e1bf.mp4

Handling unknown value: select the first one from the list.

https://user-images.githubusercontent.com/1846548/182596691-78c167a9-38e6-44b2-8199-b43ae8510175.mp4

src/ComboBox/ComboBox.svelte
```diff
@@ -205,7 +205,9 @@
   $: menuId = `menu-${id}`;
   $: comboId = `combo-${id}`;
   $: highlightedId = items[highlightedIndex] ? items[highlightedIndex].id : 0;
-  $: filteredItems = items.filter((item) => shouldFilterItem(item, value));
+  $: /** @type Array<ComboBoxItem> */ filteredItems = items.filter((item) =>
```
brunnerh, line 208: Shouldn't the type already be inferred correctly based on the type of `items`?

src/ComboBox/ComboBox.svelte
```diff
@@ -299,11 +300,33 @@
...
+              });
+              const matchedItem = listOfItems.find(
+                (e) => e.text === value?.toLowerCase()
+              );
+              if (matchedItem) {
+                // typed value has matched
+                open = false;
+                selectedItem = filteredItems.find(
+                  (e) => e.id === matchedItem.id
+                );
```
brunnerh, line 320: `listOfItems` seems unnecessary.
- `e.text?.toLowerCase()` could be done directly in a `find` on filteredItems
- The additional `filteredItems.find` would not be needed and `matchedItem` could be assigned directly to `selectedItem`
- Also, the `disabled` is disregarded when determining matching items. Those probably should not be considered potential matches?

src/ComboBox/ComboBox.svelte
```diff
@@ -299,11 +300,33 @@
...
+                value = selectedItem.text;
+                selectedId = selectedItem.id;
+              } else {
+                open = false;
+                if (filteredItems[0]) {
+                  value = itemToString(filteredItems[0]);
+                  selectedItem = filteredItems[0];
+                  selectedId = filteredItems[0].id;
+                }
```
brunnerh, line 329: Would try to avoid duplicating the index.
Also, this probably should `find` the first non-disabled item instead of just picking index `0`.

brunnerh (changes requested):
I am not a maintainer, but would have a few suggestions.

metonym:
Thank you for opening a PR, and thank you @brunnerh for initiating the review – I agree with your comments.

src/ComboBox/ComboBox.svelte
metonym: Nice catch – items that are `disabled` should not be matched.

src/ComboBox/ComboBox.svelte
patoi: I respect that it takes time to write such a helpful reply. I will fix it.

src/ComboBox/ComboBox.svelte
patoi: Very good comments, thank you!

src/ComboBox/ComboBox.svelte
patoi: OK, I think that's a good suggestion. This will be a separate commit.

patoi:
@brunnerh @metonym Thanks for the quick and good suggestions. Based on these, I modified the code.

patoi:
Selection via typing only find the enabled item, and first enabled item will be selected:

https://user-images.githubusercontent.com/1846548/182781397-d0112977-50fc-4e64-befe-e8561aaa36ad.mp4

src/ComboBox/ComboBox.svelte
```diff
@@ -299,11 +298,26 @@
...
+                selectedId = selectedItem.id;
+              } else {
+                open = false;
+                // find first enabled item
+                const matchedItem = filteredItems.find((e) => !e.disabled);
+                if (matchedItem) {
+                  value = itemToString(matchedItem);
+                  selectedItem = matchedItem;
+                  selectedId = matchedItem.id;
+                }
```
brunnerh, line 320: This looks like it could be simplified further, the state updates look like they should be identical.

There is a discrepancy with the `value`:
```js
// if
value = selectedItem.text;
// else
value = itemToString(matchedItem);
```

Maybe set `matchedItem` in way that directly falls back to the first item, then the `else` could maybe be omitted?

Something like

```js
const matchedItem =
  filteredItems.find((e) => !e.disabled && e.text.toLowerCase() === value?.toLowerCase()) ??
  filteredItems.find((e) => !e.disabled);
```

(Though `prettier` will probably make this look ugly 😅)

You could also pre-filter on the `disabled`, since this is required in any case. Not sure if this would negatively impact performance for large lists, though. (Ideally one would use a [generator function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator) so the list is not iterated yet, but their ergonomics are currently not that great. [This proposal](https://github.com/tc39/proposal-iterator-helpers) would fix that.)

brunnerh (commented):
Thank you for your efforts.
Noticed one more thing...

src/ComboBox/ComboBox.svelte
patoi: OK, I think it can be done. Thanks!

brunnerh: Isn't `itemToString` the correct version of setting the `value`?
(That is what was used prior in the code.)

patoi: Hm. Good question. Based on the code, it seems that itemToString is useful, if the text is empty. On the other hand, according to the definition of the type, "text" cannot be empty.
```javascript
@typedef {{ id: ComboBoxItemId; text: string; disabled?: boolean; }} ComboBoxItem
```
I will modify it according to your suggestion.

src/ComboBox/ComboBox.svelte
```diff
@@ -299,11 +298,18 @@
...
+              const matchedItem =
+                filteredItems.find(
+                  (e) =>
+                    e.text.toLowerCase() === value?.toLowerCase() && !e.disabled
+                ) ?? filteredItems.find((e) => !e.disabled);
+              if (matchedItem) {
+                // typed value has matched or fallback to first enabled item
+                open = false;
+                selectedItem = matchedItem;
+                value = itemToString(selectedItem.text);
```
brunnerh, line 311: Argument should be the entire item, not just `text`.

patoi: Thanks, I fixed!

brunnerh (approved):

metonym (approved):
Excellent – thank you both.

PR #1434, opened 2022-08-10
Add Switch Role to Toggle Component for Accessibility

bertybot:
ARIA Authoring Guide recommends adding the role "switch" to on/off components like the toggle.
- [ARIA Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/switch/)
- [Documentation For Switch Role](https://w3c.github.io/aria/#switch)
- [Example of Checkbox Switch](https://www.w3.org/WAI/ARIA/apg/example-index/switch/switch-checkbox.html)

### DIFF
--- a/src/Toggle/Toggle.svelte
+++ b/src/Toggle/Toggle.svelte
@@ -54,6 +54,7 @@
   on:mouseleave
 >
   <input
+    role="switch"
     type="checkbox"
     class:bx--toggle-input="{true}"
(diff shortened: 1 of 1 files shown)


metonym (approved):
Good catch – thanks!

metonym:
Fixed in [v0.67.8](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.67.8)

PR #1436, opened 2022-08-10
fix(inline-checkbox): remove `aria-checked="mixed"` if indeterminate

metonym:
Ref: https://github.com/carbon-design-system/carbon/pull/11908

This implements a backported change to the `InlineCheckbox` component, which avoids setting `aria-checked="mixed"` if the checkbox is indeterminate.

Within the library itself, `InlineCheckbox` is only used in a selectable `DataTable`.


metonym:
Fixed in [v0.67.8](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.67.8)

PR #1440, opened 2022-08-11
feat(ToolbarBatchActions): Allow showing the toolbar via override

### LINKED ISSUE #1438: Pressing Cancel on Datatable toolbarbatchactions sets selectedRowIds to empty list
When using the DataTable along with a toolbar + toolbarbatchactions, the datatable may be initialized with a set of selectedRowIds. For example, a data store of already selected "load balancers" is rendered. 

When batchSelection is enabled and a user makes changes to the pre-loaded selection and decides to "cancel",  the initially selectedRowIds are cleared.

Since the user clicked 'cancel' to negate his changes, the selectedRowIds should probably be restored to the selection prior to any changes and not simply an empty list.

Since there is no way to bind to the cancel button's on:click from the toolbarbatchactions.svelte , the cancel button becomes a "clear selection" button instead of a true "cancel". 

I would recommend exposing the on:click event of the cancel button in the toolbarbatchactions component file so that the behavior can be developer defined.  

Update:  A demo of this behavior can be seen here: https://carbon-components-svelte.onrender.com/framed/DataTable/DataTableBatchSelectionToolbar?theme=white

naegelin:
Fixes #1438.

Show the toolbar even if there are no selected toolbars.

Maintains current behavior but allows for an over-ride to ensure the toolbarbatchactions bar does not close when the user has unselected all items.

Example Useage:
```
<Toolbar>
    <ToolbarContent>
        <ToolbarBatchActions show={true}>
         ....
        </ToolbarBatchActions>
    </ToolbarContent>
</Toolbar>
```

### DIFF
--- a/src/DataTable/ToolbarBatchActions.svelte
+++ b/src/DataTable/ToolbarBatchActions.svelte
@@ -5,13 +5,19 @@
    */
   export let formatTotalSelected = (totalSelected) =>
     `${totalSelected} item${totalSelected === 1 ? "" : "s"} selected`;
-
+  
+  /** 
+   * Set to `true` to show the ToolbarBatchActions regardless of row selection
+   * @type {boolean}
+   */
+  export let show = false;
+  
   import { onMount, getContext } from "svelte";
   import Button from "../Button/Button.svelte";
 
   let batchSelectedIds = [];
 
-  $: showActions = batchSelectedIds.length > 0;
+  $: showActions = batchSelectedIds.length > 0 || show;
 
   const ctx = getContext("DataTable");
   const unsubscribe = ctx.batchSelectedIds.subscribe((value) => {


src/DataTable/ToolbarBatchActions.svelte
```diff
@@ -5,13 +5,19 @@
    */
   export let formatTotalSelected = (totalSelected) =>
     `${totalSelected} item${totalSelected === 1 ? "" : "s"} selected`;
-
+  
+  /** 
+   * Set to `true` to show the ToolbarBatchActions regardless of row selection
+   * @type {boolean}
+   */
+  export let show = false;
```
metonym, line 13: Let's rename this to `active`.

src/DataTable/ToolbarBatchActions.svelte
```diff
@@ -5,13 +5,19 @@
    */
   export let formatTotalSelected = (totalSelected) =>
     `${totalSelected} item${totalSelected === 1 ? "" : "s"} selected`;
-
+  
+  /** 
+   * Set to `true` to show the ToolbarBatchActions regardless of row selection
```
metonym, line 10: Nit: brevity

```suggestion
   * Set to `true` to show the toolbar regardless of row selection
```

src/DataTable/ToolbarBatchActions.svelte
naegelin: sure thing.

naegelin:
@metonym renamed show to active as requested and updated docs.

metonym (commented):
LGTM

metonym (approved):

PR #1441, opened 2022-08-11
feat(ToolbarBatchActions): Override cancel button

### LINKED ISSUE #1438: Pressing Cancel on Datatable toolbarbatchactions sets selectedRowIds to empty list

naegelin:
Fixes #1438.

Give the user the ability to override the functionality of the cancel button by passing in their own function. If no override is provided, the existing functionality of reseting selected row ids is preserved.

Example implementation:

```
<ToolbarBatchActions cancel={()=>{console.log("User clicked cancel!")}}>
 ...
</ToolbarBatchActions>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- ## Unreleased -->
 
+## [0.67.9](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.67.9) - 2022-08-11
+
+**Fixes**
+
+- `NotificationActionButton` types should extend `Button` props
+
 ## [0.67.8](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.67.8) - 2022-08-10
 
 **Fixes**
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -4422,9 +4422,10 @@ None.
 
 ### Props
 
-| Prop name           | Required | Kind             | Reactive | Type                                           | Default value                                                                                       | Description                            |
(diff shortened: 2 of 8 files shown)


metonym (commented):
I support this feature. Implementation-wise, I think that dispatching a cancelable "cancel" event would be more elegant. It would also have the benefit of listening to the cancel event in the first place, instead of just overriding the default.

Svelte v3.48 supports cancelable events, which suits this use case very well. The `InlineNotification` component is one example that employs this API.

```svelte
<Button
  on:click={() => {
    const shouldContinue = dispatch("cancel", null, { cancelable: true });

    if (shouldContinue) {
      ctx.resetSelectedRowIds();
    }
  }}
/>
```

Usage:

```svelte
<ToolbarBatchActions on:cancel={(e) => { e.preventDefault(); } />
```

naegelin:
Agree @metonym  cancelable events is the way to go. Updated the PR including docs.

metonym (approved):
Excellent – looks like this PR needs rebasing now that #1440 is merged. Overall, I'm happy with this approach as it also enables the consumer to listen to the `cancel` event in the first place.

naegelin:
@metonym should be rebased now and good to go

PR #1442, opened 2022-08-11
fix(notification): add extends to button

piperswe:
This way, props from Button can be used in NotificationActionButton and
still typecheck correctly.

### DIFF
--- a/src/Notification/NotificationActionButton.svelte
+++ b/src/Notification/NotificationActionButton.svelte
@@ -1,4 +1,6 @@
 <script>
+  /** @extends {"../Button/Button.svelte"} ButtonProps */
+
   import Button from "../Button/Button.svelte";
 </script>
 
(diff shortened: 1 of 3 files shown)


src/Notification/NotificationActionButton.svelte
```diff
@@ -1,5 +1,7 @@
 <script>
   import Button from "../Button/Button.svelte";
+  
+  /** @extends {"../Button/Button.svelte"} ButtonProps */
```
metonym, line 4: This is great – thanks for the fix.

For consistency, could the comment be moved to [top of the script block](https://github.com/carbon-design-system/carbon-components-svelte/blob/ab5eff5205587a88b1682ec254e17184c991c010/src/Accordion/Accordion.svelte#L2)?

piperswe: moved!

metonym (approved):

metonym:
Fixed in [v0.67.9](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.67.9). Thanks again!

PR #1444, opened 2022-08-14
feat(toast-notification): add `fullWidth` prop

metonym:
By design, the `ToastNotification` has a hardcoded width of `18rem`. It would be useful to allow it to span the full width (`100%`). A simple use case would be having an increased maximum width for toast notifications displayed in the top right corner.

The `Slider` component added a [similar prop](#1354) that overrides the default width value.

```svelte
<div style="max-width: 24rem">
  <ToastNotification fullWidth />
</div>
```


metonym:
Released in [v0.68.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.68.0)

Issue #1449, opened 2022-08-16
Feature: When Datatable is selectable, emit select event on checkbox or radio button activity


naegelin:
Example Scenario:

The user sees a DataTable with multi-selection checkboxes . The user checks two checkboxes from two table rows. The developer wishes to react to these checks but is unable to.

See data table demo: https://carbon-components-svelte.onrender.com/framed/DataTable/DataTableBatchSelectionToolbar?theme=white

Currently, only the selected row ids are updated when users select table rows using batch selection. This makes it difficult to build functionality / reactivity based on the user's actions.

Proposed solution:
Similar to on:click:row--expand , surface a on:click:row--select event on the DataTable and dispatch the row selection status when a row is selected/unselected as follows:

`dispatch('click:row--select', { row, selected: <selection status> });`

metonym:
Released in [v0.69.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.69.0)

PR #1450, opened 2022-08-16
emit select event on checkbox or radio button activity

naegelin:
Proposed solution to #1449

Implementation follows existing `click:row--expand ` pattern. 

No backwards-compatibility issues AFIAK.


metonym (approved):
This is a cut-and-dry change. Nice work.

PR #1453, opened 2022-08-17
feat(data-table): dispatch `on:click:header--select` event

metonym:
Follow-up to #1450

#1450 added a dispatched event `click:row--select` for selectable rows.

This PR adds an additional event when using a batch selectable `DataTable`, where clicking the checkbox in the header will either select or deselect all rows. A similar event would be `click:header--expand` for a batch expandable `DataTable`.

Compared to `click:row--select`, one difference in the `e.detail` API is that the checkbox state can be `indeterminate`.

```ts
// indeterminate
0 < n < rows.length - 1
```

**Sample usage**

```svelte
<DataTable
  batchSelection
  on:click:header--select={(e) => {
    console.log("header--select", e.detail.selected); // boolean
    console.log("header--select", e.detail.indeterminate); // boolean
  }}
/>
```


metonym:
Released in [v0.69.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.69.0)

PR #1461, opened 2022-08-22
fix(date-picker): set `role`, `aria-label` attributes on calendar container

metonym:
Ref: https://github.com/carbon-design-system/carbon/blob/6b57099f46a734c7642b77093ff937e23f79be40/packages/react/src/components/DatePicker/DatePicker.js#L385-L390


metonym:
Fixed in [v0.70.1(https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.70.1)

PR #1484, opened 2022-09-07
fix(ComboBox): handling TAB behaviour

patoi:
**Problem:**
Typing a valid value into ComboBox (e.g: Fax) and press TAB: value and selectId will be empty.

https://user-images.githubusercontent.com/1846548/188866242-f14a4e40-962a-4067-86b1-9a007b01eb67.mp4

**Fix:**
If typed value has matched, then selectId and value will be set correctly.

https://user-images.githubusercontent.com/1846548/188866300-5b3c7ea6-3a5e-45e5-8619-396dc3fc8d26.mp4

**Change in default behaviour:** if typed value has not matched, then ComboBox selectId/value remains unchanged. So far, the first value has been selected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 .DS_Store
 .vscode
 .idea
+yarn-error.log
--- a/docs/src/pages/framed/ComboBox/FilterableComboBox.svelte
+++ b/docs/src/pages/framed/ComboBox/FilterableComboBox.svelte
@@ -1,6 +1,7 @@
 <script>
   import { ComboBox } from "carbon-components-svelte";
 
+  let selectedId
   function shouldFilterItem(item, value) {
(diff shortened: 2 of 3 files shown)


patoi:
@metonym Could you review this enhancement? Thanks!

metonym:
Thanks for the PR, but I'm not sure if this matches expectations of tabbing behavior.

A more common autocompletion pattern is pressing "Tab" to complete a partial input value, but not actually selecting it. Pressing "Enter" should actually select the matched item. Although I'm curious if you could point me to examples that do follow this pattern.

Regardless, it would be nice if the consumer could override the "Tab" behavior and implement custom logic.

patoi:
Thank you for checking! I accept what you wrote. I'll see what I can do.

PR #1495, opened 2022-09-23
fix(button): resolve visual regression in `Pagination`, `PaginationNav`

### LINKED ISSUE #1476: danger ghost icon Button is too wide
For some reason, the icon inside an icon-only `"danger-ghost"` button has an extra `margin-left: 0.5rem`.

Here's a `primary` button beside a (hovered) `danger-ghost`:

image

metonym:
#1477 removed the "bx--btn__icon" class for icon-only buttons to fix #1476.

However, it caused a visual regression in the next/previous buttons in `Pagination` and `PaginationNav`, which require the "bx--btn__icon" class.

The solution is to use an inline style attribute to manually unset the left margin instead of removing the class altogether. This is the simplest solution because it does not change the existing component API (like adding a prop/slot or using context).

---

### Before

Before

### After

After

### DIFF
--- a/src/Button/Button.svelte
+++ b/src/Button/Button.svelte
@@ -175,7 +175,8 @@
     <slot /><svelte:component
       this="{icon}"
       aria-hidden="true"
-      class="{hasIconOnly ? '' : 'bx--btn__icon'}"
+      class="bx--btn__icon"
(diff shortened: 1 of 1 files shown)


metonym:
Fixed in [v0.70.9](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.70.9)

PR #1502, opened 2022-09-30
fix transitions delaying navigation

### LINKED ISSUE #1501: transitions causing jittery navigation
The `<Header>` component transition plays when navigating to a different page where it doesn't exist, causing it to persist on screen during navigation.

https://user-images.githubusercontent.com/54401897/193199481-85d143cd-b530-46ec-a6d6-6f31d82509a9.mp4

Notice how the next page's content has loaded beneath the current page because the transition has to play.

This can be solved by localising the transition of components.
https://svelte.dev/docs#template-syntax-element-directives-transition-fn-transition-events

The issue is that when parent blocks are destroyed (navigating to a different page), the component still plays its transition.
However, localising the transition causes it to play only when the component is specifically destroyed (menu closes).
Therefore, each component should have the `local` suffix added to it such as:
```diff
- <nav transition:fade>
+ <nav transition:fade|local>
```

s3812497:
fixes #1501

Adds `local` to transitions to prevent them from playing between page navigations.

https://svelte.dev/docs#template-syntax-element-directives-transition-fn-transition-events
> Local transitions only play when the block they belong to is created or destroyed, not when parent blocks are created or destroyed.

### DIFF
--- a/src/ImageLoader/ImageLoader.svelte
+++ b/src/ImageLoader/ImageLoader.svelte
@@ -86,7 +86,7 @@
       style="width: 100%;{$$restProps.style}"
       src="{src}"
       alt="{alt}"
-      transition:fade="{{ duration: fadeIn ? fast02 : 0 }}"
+      transition:fade|local="{{ duration: fadeIn ? fast02 : 0 }}"
     />
   {/if}
   {#if error}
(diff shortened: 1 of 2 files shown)


metonym (approved):

metonym:
Thanks – fixed in [v0.70.10](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.70.10).

Issue #1510, opened 2022-10-04
`NumberInput`: unintuitive interaction between `allowEmpty` and `min` props


malinowskip:
Currently, if the `allowEmpty` option is set to `true` and the `min` prop is set to a positive integer, but the value is empty (`null`), the component displays an error. I think the intuitive behavior in this case would be for the `allowEmpty` prop to take precedence, i.e. the component should tolerate a `null` value.

metonym:
Fixed in [v0.70.11](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.70.11)

PR #1511, opened 2022-10-04
fix(number-input): fix interaction between `allowEmpty` and `min`

malinowskip:
Proposal to fix https://github.com/carbon-design-system/carbon-components-svelte/issues/1510


metonym (approved):
Agreed. Thanks!

PR #1518, opened 2022-10-12
feat!: use 1-indexing for PaginationNav

### LINKED ISSUE #1513: Consistent page indexing between Pagination and PaginationNav
## Issue

I noticed the [`<Pagination />`](https://carbon-components-svelte.onrender.com/components/Pagination) and [`<PaginationNav />`](https://carbon-components-svelte.onrender.com/components/PaginationNav) components both have a `page` prop, except they have different indexing:

- `Pagination`'s `page` prop is 1-indexed
- `PaginationNav`'s `page` prop is 0-indexed

Would it be a good idea to have both of these `page` props be consistent? 

## Proposal

I would elect for 1-indexed counting since it's more indicative of the active and visible page number, and would line up with potential `url.searchParams` if used such as `?page=1`. Any JS arrays or database rows that are 0-indexed could have some `+1` math added to them. It would help for applications that make use of both `Pagination` and `PaginationNav` without having to maintain off-by-one logic between the two. Either way, this would be a breaking change if implemented.

Thanks for maintaining.

theetrain:
Closes #1513 

Changes include:

- BREAKING CHANGE (PaginationNav): use 1-indexing
  closes #1513 
- fix(PaginationNav): make 'change' event emit after user interaction instead of reactive event (explanation below)

I removed the event dispatcher in `<PaginationNav>` for `change` called in `afterUpdate` since it would dispatch on mount when it should only dispatch upon a user's interaction. When adding a change handler to `<PaginationNav>`, I would have to add a mount flag as a workaround. This treatment is similar to the one I provided in #1497.

With the changes in this PR I can remove my workaround in my change handler, allowing me to trust the `change` event would fire after user input only:

```diff
<script>
  import {
    PaginationNav
  } from 'carbon-components-svelte'
- import { onMount } from 'svelte'
  import { goto } from '$app/navigation'
  import { page } from '$app/stores'

  /** @type {import('./$types').PageServerData} */
  export let data

- let mounted = false

  // ...

  /**
   * changePage via pagination dropdown or prev/next buttons
   * @param e
   */
  const changePage = async e => {
-   if (!mounted) return
    const url = new URL($page.url)
    url.searchParams.set('page', e.detail.page)

    await goto(`?${url.searchParams.toString()}`)
  }

- onMount(() => mounted = true)
</script>

<!-- ...other html -->

<PaginationNav
  page={listPage}
  total={Math.ceil(data.itemCount / pageSize)}
  shown={5}
  on:change={changePage}
/>
```

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -2694,7 +2694,7 @@ None.
 
 | Prop name    | Required | Kind             | Reactive | Type                 | Default value                | Description                               |
 | :----------- | :------- | :--------------- | :------- | -------------------- | ---------------------------- | ----------------------------------------- |
-| page         | No       | <code>let</code> | Yes      | <code>number</code>  | <code>0</code>               | Specify the current page index            |
+| page         | No       | <code>let</code> | Yes      | <code>number</code>  | <code>1</code>               | Specify the current page index            |
 | total        | No       | <code>let</code> | No       | <code>number</code>  | <code>10</code>              | Specify the total number of pages         |
 | shown        | No       | <code>let</code> | No       | <code>number</code>  | <code>10</code>              | Specify the total number of pages to show |
 | loop         | No       | <code>let</code> | No       | <code>boolean</code> | <code>false</code>           | Set to `true` to loop the navigation      |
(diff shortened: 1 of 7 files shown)


metonym (approved):
Excellent – this works great. I agree that the `change` event should only fire on a distinct change event and not after every update.

theetrain:
Released in [v0.71.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.71.0)

PR #1524, opened 2022-10-15
Tabindex should be a number not a string

ibilux:
We all know that JavaScript is weird !
But we are using Typescript. So tabindex should be a number not a string.

### DIFF
--- a/src/Button/Button.svelte
+++ b/src/Button/Button.svelte
@@ -69,7 +69,7 @@
   export let href = undefined;
 
   /** Specify the tabindex */
-  export let tabindex = "0";
+  export let tabindex = 0;
 
   /** Specify the `type` attribute for the button element */
   export let type = "button";
--- a/src/DataTable/ToolbarSearch.svelte
+++ b/src/DataTable/ToolbarSearch.svelte
@@ -35,7 +35,7 @@
   export let filteredRowIds = [];
 
   /** Specify the tabindex */
-  export let tabindex = "0";
+  export let tabindex = 0;
 
   /**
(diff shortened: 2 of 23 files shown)


metonym:
Interestingly, Svelte Language Tools seems to tolerate strings and numbers for `tabindex`, but not empty strings.

I'm marking this as a TypeScript-related pain point, as it does seem that the number type is more closely associated with `tabindex` (especially given the "index" part).

Since both strings/numbers are tolerated, a non-breaking approach would be to type all `tabindex` props as `string | number`. On the other hand, only allowing numbers would be a trivial breakage, and it does seem reasonable that most users would assume it's a number. This library does not collect telemetry so I can only guess.

Thoughts @theetrain?

Screen Shot 2022-12-10 at 10 24 41 AM

ibilux:
@metonym I usually use `<script lang="ts">` when editing/updating `carbon-components-svelte` source code to get a strict syntactical superset.
But, as you said it may be hard for users to think of `tabindex` as a `number` without `UX` view.

theetrain:
I would support the `string | number` type since both plain HTML and Svelte support it. While testing, I noticed the svelte language server only properly checks `tabindex` values when they're assigned as strings; though this may be an upstream issue.

```html
<!-- a11y warning is thrown -->
<button tabindex="2">Hello</button>

<!-- a11y warning is NOT thrown -->
<button tabindex={2}>Hello</button>
```

Interestingly, doubles such as `3.9` are also supported by HTML, but that's contrived and hopefully never used.

PR #1528, opened 2022-10-19
use 'sm' instead of 'small'

ibilux:
Why not use `sm` instead of `small` ?
It's better for semantics  with other components, or there is something I'm missing ?


joseph-ismailyan:
Does this introduce a backwards compatibility issue for people who've already used `size="small"`?

PR #1546, opened 2022-10-30
fix(overflow-menu): dispatch close event for all cases

### LINKED ISSUE #1541: OverflowMenu don't always dispatch close event.
When you click outside the menu it only change `open` to `false`. It should also dispatch the `close` event.

metonym:
Fixes #1541


metonym:
Fixed in [v0.70.13](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.70.13).

PR #1555, opened 2022-11-20
Fix `step` floating point issue in `NumberInput`

### LINKED ISSUE #233: NumberInput and floating point values
I was trying to use the NumberInput component and I found out that using the up&down buttons instead of the arrow keys on my keyboard (thus triggering the custom function that changes the value) causes the classic approximation problem of the floating point binary representation.

For example, setting the step at 0.01 (which cannot be expressed correctly in base 2) and starting from 0 gives the following sequence of values displayed in the input field:
0.01, 0.02, 0.03, 0.04, 0.05, 0.060000000000000005, 0.07, 0.08, 0.09, 0.09999999999999999, ..., and so on.

Possible solutions:
- use a library that handles these problems
or
- replace the following code (in NumberInput.svelte, line 149, function updateValue):
```
const nextValue = (value += direction * step);
```
with
```
const stepDecimalPlaces = step % 1 === 0 ? 0 : step.toString().split('.')[1].length;
const nextValue = stepDecimalPlaces === 0
    ? value + step * direction
    : Math.round(value * Math.pow(10, stepDecimalPlaces) + step * Math.pow(10, stepDecimalPlaces) * direction) / Math.pow(10, stepDecimalPlaces);
```

### LINKED ISSUE #486: NumberInputs floating point error with small step
If I do:
```
<NumberInput
type="number"
min={0}
max={1}
value={0.01}
step={0.001}
invalidText="Number must be between 0 and 1"
/>
```
When you increment the value a few times it starts to diverge from multiples of 0.001

### LINKED ISSUE #1554: NumberInput step problem.
If you put a `step={0.1}` and a let's say a `value={8.5}` and you press the `+` or `-` steppers multiple time, the step will jump of 0.999999 instead of 0.1.
image

~Another issue, if you resize, the number input will disapear in the background...? Maybe paddings?~
image

Instead of having the floating point problem with javascript, you could use the built-in `stepUp` and `stepDown` element built-in functions.
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/stepUp

Here's an exemple : https://jsfiddle.net/L105/rj7zm35o/3/

jqlio18:
Fixes #233, fixes #486, fixes #1554

Use the built-in function of steppers then assign the value.

JS has a bug with floating point which result in a visual bug where input has many decimals. 
There's 2 ways to fix this : 

1. Converting `float` to an `int` (by multiplying by 10 x number of decimals) then make the equation then reconvert to a floating value.
2. Use built-in functions to step up and down.

I prefer the second approach which is simpler.

Also, with this method we don't need to test the `min` and `max` values since `stepUp` and `stepDown` built-in functions do this internally. We do need to assign `value` to the input value because Svelte won't know that the value changed.

I also added a step test and updated the docs to add the step function.

For the margin I understand that because of the error Icon we can't reduce the padding, I'll update the issue.

### DIFF
--- a/docs/src/pages/components/NumberInput.svx
+++ b/docs/src/pages/components/NumberInput.svx
@@ -23,6 +23,10 @@ See [#no-value](#no-value) to allow for an empty value.
 
 <NumberInput min="{4}" max="{20}" value="{4}" invalidText="Number must be between 4 and 20." helperText="Clusters provisioned in your region" label="Clusters (4 min, 20 max)" />
 
+## With step value
+
+<NumberInput value="{1}" helperText="Step of 0.1" step={0.1} label="Clusters" />
+
 ## No value
 
 Set `allowEmpty` to `true` to allow for no value.
--- a/src/NumberInput/NumberInput.svelte
+++ b/src/NumberInput/NumberInput.svelte
@@ -113,16 +113,13 @@
 
   const dispatch = createEventDispatcher();
 
-  function updateValue(direction) {
-    const nextValue = (value += direction * step);
-
-    if (nextValue < min) {
-      value = min;
-    } else if (nextValue > max) {
-      value = max;
+  function updateValue(isIncrementing) {
+    if (isIncrementing) {
+      ref.stepUp();
     } else {
-      value = nextValue;
+      ref.stepDown();
     }
+    value = ref.value;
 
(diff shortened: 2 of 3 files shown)


metonym (approved):
This is brilliant.

PR #1559, opened 2022-11-21
Fix 1556 Toggle events fire before `toggled` is changed

jqlio18:
`Toggle` events `change`, `keyup` and `click` will stay with the old value instead of the new when firing. By changing the order of  execution of `change` and `keyup`, it seems to fix the problem. `click` stayed the same because the user may wants to execute something before the `toggled` change status.


metonym:
Thank you. Fixed in [v0.70.13](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.70.13).

PR #1560, opened 2022-11-21
feat(progress-bar): add `status` prop

jqlio18:
ProgressBar was missing the `status` prop according to spec.

Status default is `active` and have `finished` and `error` state. When setting a `finished` or `error` state, the bar is full, use the variables in CSS to set the color and add the icon at the end of the label.

I don't have access to SCSS files so I created a style element in the end. Please tell me if I did the right thing.


theetrain (commented):
Overall, this is great work and follows the spec closely!

1. For the styles, I see that classes were added for progress bar in [carbon-components@10.57.0](https://github.com/carbon-design-system/carbon/releases/tag/v10.57.0). If you upgrade to the latest v10 version of `carbon-components`, those `*--finished` and `*--error` classes should become available. It's best to use the carbon-provided classes since anything involving tokens or colours should respond to theme changes.

2. Lastly, I recommend adjusting the error or success icon to remain on the top-right while label text wraps. 

Here's how your changes currently appear on very small viewport widths:

image

Your label text wraps (as it should 👍 ) when the parent wrapper gets very narrow. This is more accommodating for languages that take up a lot of space and for users who set their browser font sizes to something large.

Carbon React uses text truncation, which I hope to avoid for the above reasons:

image

metonym:
Thanks for taking the time to add these states as well as examples.

Agreed with @theetrain that styles should be used from `carbon-components`. The main reason is to support the white/g10/g80/g90/g100 CSS files, not just `all.css` which uses CSS variables.

1. Upgrade `carbon-components` to 10.57
2. Run "yarn build:css"

A [previous PR](https://github.com/carbon-design-system/carbon-components-svelte/pull/1213) as an example.

jqlio18:
I agree with you. I'm more comfortable now working on the project and know how it populates `CSS`. I will work on this tonight and fix this commit.

jqlio18:
My approach for this commit is to use a variable instead of assigning the transform in the style attribute. This way prevent the override of css rules of carbon. Everything seems fine now :)
image

metonym (changes requested):
This is shaping up nicely. Several things:

1. Needs to use `carbon-components` to version >=10.57 for the status styles.

```
yarn add -D carbon-components
yarn build:css
```

2. I'd prefer no added styles, which adds more specificity and JS

3. Needs to run `yarn build:docs` to update the type definitions

jqlio18:
I have an idea to avoid using the `style` tag. I'll post a PR and update the docs.

jqlio18:
Here's the one I think :)

src/ProgressBar/ProgressBar.svelte
```diff
@@ -32,6 +38,16 @@
   /** Set an id for the progress bar element */
   export let id = "ccs-" + Math.random().toString(36);
 
+  import CheckmarkFilled from "../icons/CheckmarkFilled.svelte";
+  import ErrorFilled from "../icons/ErrorFilled.svelte";
+
+  let isActiveAndDeterminate = false;
```
theetrain, line 44: It appears this isn't being used; you may remove it.

src/ProgressBar/ProgressBar.svelte
```diff
@@ -55,6 +73,12 @@
...
     </slot>
+    {#if status === "error" || status === "finished"}
+      <svelte:component
+        this="{statusIcons[status]}"
+        class="bx--progress-bar__status-icon"
+      />
+    {/if}
   </label>
   <div
     role="progressbar"
```
theetrain, line 84: Please add `aria-busy` if you can, even though it's out of scope:

```html
<div
  role="progressbar"
  aria-busy="{status === 'active'}"
  ...
>
```

docs/src/pages/components/ProgressBar.svx
```diff
@@ -21,6 +21,18 @@ Specify a `value` for the progress bar to be determinate.
 
 <ProgressBar value={40} labelText="Upload status" helperText="40 MB of 100 MB" />
 
+## Finished status
+
+Specify `status="finished"` for the progress bar.
+
+<ProgressBar value={100} status="finished" labelText="Upload status" helperText="100 MB of 100 MB" />
```
theetrain, line 28: I recommend updating these labels:

- `labelText="Upload file"`
- `helperText="Upload complete"`

Reference: https://carbondesignsystem.com/components/progress-bar/accessibility#provide-status-for-indeterminate-and-errored-progress-bars

docs/src/pages/components/ProgressBar.svx
```diff
@@ -21,6 +21,18 @@ Specify a `value` for the progress bar to be determinate.
...
+
+## Error status
+
+Specify `status="error"` for the progress bar.
+
+<ProgressBar value={40} status="error" labelText="Upload status" helperText="40 MB of 100 MB" />
```
theetrain, line 34: I recommend updating these lables:

- `labelText="Upload file"`
- `helperText="Invalid file format"`

src/ProgressBar/ProgressBar.svelte
```diff
@@ -32,6 +38,16 @@
...
+  import CheckmarkFilled from "../icons/CheckmarkFilled.svelte";
+  import ErrorFilled from "../icons/ErrorFilled.svelte";
+
+  let isActiveAndDeterminate = false;
+
+  const statusIcons = {
+    error: ErrorFilled,
+    finished: CheckmarkFilled,
+  };
+
```
theetrain, line 50: Below here, please set `capped` to `0` when there is an error. Since it already uses 2 ternaries, maybe use a switch or reactive block to improve readability:

```js
let capped
$: {
  if (status === 'error' || value < 0) {
    capped = 0
  } else if (value > max) {
    capped = max
  } else {
    capped = value
  }
}
```

docs/src/pages/components/ProgressBar.svx
```diff
@@ -45,6 +57,10 @@ The inline variant visually hides the `helperText`.
 
 <ProgressBar kind="indented" value={40} labelText="Upload status" helperText="40 MB of 100 MB" />
 
+## Indented status variant
+
+<ProgressBar kind="indented" status="finished" value={40} labelText="Upload status" helperText="100 MB of 100 MB" />
```
theetrain, line 62: I recommend updating these labels:

- `labelText="Upload file"`
- `helperText="Upload complete"`

theetrain (changes requested):
Overall this is very functional! Tested on Chrome and Safari + VoiceOver. In addition to my comments below, please perform the following:

1. Rebase your branch against the latest `master`
2. Install carbon-components v10.57.0 with `yarn add -D carbon-components@10.57.0` and commit package.json and yarn.lock
3. Run `yarn build:css` and `yarn build:docs` and commit changes

metonym (approved):
I'd like to get this merged; I went ahead and did another review, and made some changes:

- finished/error statuses cannot be indeterminate fdd57f02ba9317472cbb6c5ca9d4ec6c48bfcedf
- re-generated the CSS files to include the new classes 3c551803a3550753b3609700459c8dfdedfd2eb6
- re-ran `yarn build:docs` so that the types are updated 3daca763d6ce47624b371f5dd5723be13da8b857
- asserted the `status` prop in `ProgressBar.test.svelte` f2a0fb03c6e183813dec4ebce02065f79d563539

I checked the docs and the examples look great to me.

Finally, I did some local testing with the various states:

Screen Shot 2023-02-18 at 4 59 05 PM

theetrain (commented):
Visually, this looks complete. I have a couple of unresolved comments above that ~~I recommend adding~~ I'll try adding now to get this merged:

1. Add `aria-busy` to indicate something is in progress:
    ```html
    <div
      role="progressbar"
      aria-busy="{status === 'active'}"
      ...
    >
    ```
   Reference: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role#best_practices
2. Automatically set `capped` to `0` when the `status` is `"error"`, that way the user won't have to be mindful of it, and it saves us from documenting this best practice.
    ```js
    let capped
    $: {
      if (status === 'error' || value < 0) {
        capped = 0
      } else if (value > max) {
        capped = max
      } else {
        capped = value
      }
    }
    ```

theetrain (approved):

theetrain:
This is now released in [v0.73.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.73.0).

PR #1566, opened 2022-11-29
fix `TreeNode` type so `children` works recursively

DetachHead:


metonym (changes requested):
Thanks for taking the time to fix this.

The TypeScript definitions are auto-generated and should not be directly edited. The types are generated from JSDoc comments in the Svelte source code.

Could you instead apply the change to the [`@typedef`](https://github.com/carbon-design-system/carbon-components-svelte/blob/ae3445158315f7f834ae7f4be75bfd79aea5a299/src/TreeView/TreeView.svelte#L4) and run "yarn build:docs"?

DetachHead:
@metonym done, though i couldn't get yarn to work on my PC so i'm hoping my edits match what `yarn build:docs` would've done

on a related note, how come the source code is written in js? would it make more sense to have the source code written in typescript and use [`svelte-package`](https://kit.svelte.dev/docs/packaging) to distribute type-stripped `.svelte` files and `.d.ts` files?

metonym:
> though i couldn't get yarn to work on my PC so i'm hoping my edits match what yarn build:docs would've done

It's looking good. I'll run `yarn build:docs` so the other auto-generated docs are included in this PR.

PR #1569, opened 2022-11-29
fix(Dropdown): adjust select event and add arrow key interactions

### LINKED ISSUE #1475: Change selected dropdown item without opening dropdown list by using up and down arrows

theetrain:
Changes:

1. Dispatch `select` event only upon user interaction
    * Currently, when using Dropdown with `on:select`, this event always dispatches on initial load, which is undesirable when I depend on that event for side-effects such as navigation.
    * With these changes, it will only dispatch upon a user interaction. It's a similar dilemma to the one I had raised in #1497.

2. Allow arrow keys to open menu and highlight items, closes #1475 
    * Matches interaction behaviours in the main [Carbon Dropdown Live Demo](https://carbondesignsystem.com/components/dropdown/usage/#live-demo) also described in [Dropdown Interactions](https://carbondesignsystem.com/components/dropdown/usage/#interactions).
    * Does not fully address #1475 since the ask was to keep the Dropdown collapsed while selecting items, like an HTML `<select>`. Perhaps a design request can be made in the main Carbon repository. I also left the "slide-down transition upon opening" out of scope.

I'm in particular need for the fix in (1). Your time is appreciated! 😃

### DIFF
--- a/src/Dropdown/Dropdown.svelte
+++ b/src/Dropdown/Dropdown.svelte
@@ -112,6 +112,12 @@
 
   let highlightedIndex = -1;
 
+  $: inline = type === "inline";
+  $: selectedItem = items.find((item) => item.id === selectedId);
+  $: if (!open) {
+    highlightedIndex = -1;
+  }
+
   function change(dir) {
     let index = highlightedIndex + dir;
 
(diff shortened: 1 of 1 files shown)


metonym (approved):
Awesome fix – two birds one stone

PR #1570, opened 2022-12-01
Fix popover click outside handler when clicking siblings

vizath:
When clicking on a sibling of the popover component, the `click:outside` event was not firing, because the sibling doesn't contain the popover's ref.

Also added a usage example with a button that binds the `open` prop, since it has a small behavior to think into account when clicking on the button.


vizath:
Thanks!

PR #1577, opened 2022-12-08
fix(multi-select): avoid runtime error if `items` is empty

metonym:
Related #1545

A runtime error is thrown when focusing the input and using arrow keys of an empty `MultiSelect`.

```svelte
<MultiSelect items={[]} />
```


metonym:
Fixed in [v0.70.13](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.70.13).

PR #1581, opened 2022-12-09
chore(types)!: remove `sveltekit:` attributes from types

metonym:
[`sveltekit:` attributes were removed](https://github.com/sveltejs/kit/blob/20ed756c822615564326e300b4e44d0a4060a214/packages/kit/CHANGELOG.md#100-next454) from SvelteKit almost half a year ago.

This upgrades `sveld` and removes the attributes from the types.


theetrain (approved):
Looks good to me!

PR #1582, opened 2022-12-10
docs: link source code to folder for multiple components

metonym:
Currently, the "Source code" link at the top of a component docs navigates to a single component.

This is confusing since many components actually contain related components. One example is `ContentSwitcher`, which comprises of `ContentSwitcher` and `Switch`.

The expected behavior is that it should navigate to the folder (e.g., `src/ContentSwitcher`).

This PR does the following:

- link to folder if doc contains multiple components (e.g., Breadcrumbs, DataTable etc.)
- remove `source` from frontmatter (the metadata is already available in the Sveld auto-generated docs)
  - additionally, the current implementation defaults to the [`page.title` for the source code path](https://github.com/carbon-design-system/carbon-components-svelte/compare/docs-source-code?expand=1#diff-d41ad8b8ac387e8248685fc8e6fcc261dc5ca5b05c71a7d515c28372b4af880bL47), which may be flaky

I've tested:

- `ContentSwitcher` source code links to the folder
- `CodeSnippet` source code links to the `.svelte` file

---

Screen Shot 2022-12-10 at 10 03 20 AM

### DIFF
--- a/docs/src/layouts/ComponentLayout.svelte
+++ b/docs/src/layouts/ComponentLayout.svelte
@@ -22,7 +22,6 @@
 
   export let component = $page.title;
   export let components = [component];
-  export let source = "";
   export let unreleased = false;
   export let unstable = false;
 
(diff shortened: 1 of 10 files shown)


theetrain (approved):
I tested a few single-file sources and multi-file sources and the changes work well 👍

Issue #1583, opened 2022-12-11
NumberInput on error missing icons and CSS


jqlio18:
`NumberInput` has some visual problems

1. The right border is not showing when in error state.
2. The icons are not showing both in error and warning state.
3. When hovering a stepper, the error outline is below the stepper.

State right now :
![image](https://user-images.githubusercontent.com/3661575/206887449-0c752cfe-a802-4ff0-8abc-ad57a783aa0f.png)

What it should be : 
![image](https://user-images.githubusercontent.com/3661575/206887482-d9839045-688f-4c09-bb84-15c360598898.png)

Fixes : 

1. The right broder is a missing `css` in the last button wich should be :
```
.bx--number input[type=number][data-invalid]~.bx--number__controls .bx--number__control-btn.up-icon:after {
    background-color: var(--cds-support-error,#da1e28);
}
```
2. The icons are just not added, I'll post a Pull Request
3.  Steppers are missing this `css` : 
```
.bx--number input[type=number][data-invalid]:not(:focus)~.bx--number__controls .bx--number__control-btn:hover {
    outline: 2px solid var(--cds-support-error,#da1e28); 
    outline-offset: -2px;
}
```

metonym:
Not able to repro on the docs site: https://carbon-components-svelte.onrender.com/components/NumberInput#invalid-state

Could you provide sample code?

jqlio18:
You won't see the problem in the `invalid` state of the `NumberInput`. To replicate the issue, go to : https://carbon-components-svelte.onrender.com/components/NumberInput#default and put it at empty state. The input will have the `error` outline but not the icon.

metonym:
Yep, this is a bug. Looks like `invalid` is not reactive.

jqlio18:
I'm writing a fix for that. I got some problem with VS code since yesterday. I should be able to post a PR soon.

jqlio18:
I'm navigating in the code and the `allowEmpty` attribute is just wrong. It should be `required` attribute instead which is a valid state of an input. It validates exactly the same thing.

metonym:
Should be fixed now.

<img width="971" height="250" alt="Image" src="https://github.com/user-attachments/assets/91ea83da-288a-4bd9-bbb9-08ba30cbc650" />

PR #1584, opened 2022-12-12
NumberInput : Fix error states

jqlio18:
1. Breaking change : Change `allowEmpty` attribute to `required`. `allowEmpty` makes no sense, it should always be `required` to match base `HTML input` atrributes. 
2. The input was in `invalid state` with a red outline and the validity of the input would stay `valid` (only if empty). This patch fix this.

** NOTE TO DISCUSS : This bug was caused by the ambiguity between `invalid` and `error` variables. I think that we need to remove the `export invalid` and let it be local only. If OP wants to validate the input, we can export the `pattern` or change the `setCustomValidity` with the `ref`. Since this PR check the validity of the input on `input` and `change`, the `setCustomValidity` will set the error.

### DIFF
--- a/src/NumberInput/NumberInput.svelte
+++ b/src/NumberInput/NumberInput.svelte
@@ -39,8 +39,8 @@
   /** Set to `true` for the input to be read-only */
   export let readonly = false;
 
-  /** Set to `true` to allow for an empty value */
-  export let allowEmpty = false;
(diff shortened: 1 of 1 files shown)


metonym:
https://github.com/carbon-design-system/carbon-components-svelte/issues/1583#issuecomment-3829229154

PR #1585, opened 2022-12-14
SideNav should overlay Content in UIShell

### LINKED ISSUE #1463: SideNav should overlay Content in UIShell
When observing the UIShell on a small-width display, then expanding a collapsed SideNav, the SideNav squeezes the Content container. The SideNav should overlay the Content container without impacting the Content's layout.

To reproduce the issue:

1. Go to the [React UIShell in Storybook](https://react.carbondesignsystem.com/iframe.html?args=&id=components-ui-shell--header-base-w-navigation-actions-and-side-nav&viewMode=story)
2. Make browser width 400px
3. Open SideNav
4. Observe SideNav overlays Content
5. Go to the [Svelte UIShell demo](https://carbon-components-svelte.onrender.com/framed/UIShell/HeaderSwitcher?theme=white) and repeat steps 2-4
6. Observe SideNav squeezes Content

Here's the issue in screenshots:

React:

Svelte:

This is a possible duplicate of #516 

Thanks for maintaining!

jqlio18:
Fixes #1463

The expended class should not be applied when the `sideNav` is below the breaking point. I added the `bx--sidenav-fixed` class instead, it seems to works perfectly.

This PR also fixes the hack of `bx--content` that was set previously.

### DIFF
--- a/src/UIShell/Content.svelte
+++ b/src/UIShell/Content.svelte
@@ -1,24 +1,8 @@
 <script>
   /** Specify the id for the main element */
   export let id = "main-content";
-
-  import { isSideNavCollapsed, isSideNavRail } from "./navStore";
-
-  /**
-   * By default, the `SideNav` applies a left margin of `3rem` to `Content`
-   * if it's a sibling component (e.g., .bx--side-nav ~ .bx--content).
-   *
-   * We manually unset the left margin if the `SideNav`
-   * is collapsed and if it's not the `rail` variant.
-   */
-  $: unsetLeftMargin = $isSideNavCollapsed && !$isSideNavRail;
 </script>
 
-<main
-  id="{id}"
-  class:bx--content="{true}"
-  {...$$restProps}
-  style="{unsetLeftMargin ? 'margin-left: 0;' : ''} {$$restProps.style}"
->
+<main id="{id}" class:bx--content="{true}" {...$$restProps}>
   <slot />
 </main>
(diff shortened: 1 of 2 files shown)


src/UIShell/Content.svelte
```diff
@@ -11,14 +11,14 @@
...
-  $: unsetLeftMargin = $isSideNavCollapsed && !$isSideNavRail;
+  $: unsetLeftMargin = !$isSideNavRail && !$isSideNavExpanded;
 </script>
 
 <main
   id="{id}"
   class:bx--content="{true}"
+  style="{unsetLeftMargin ? 'margin-left: 0' : ''}"
   {...$$restProps}
-  style="{unsetLeftMargin ? 'margin-left: 0;' : ''} {$$restProps.style}"
```
metonym, line 21: I don't think that this should be changed since `$$restProps.style` will override the attribute.

metonym:
When testing this locally, the side nav is not appearing when toggling the hamburger menu.

It seems to retain the `bx--side-nav--collapsed` class even when toggled.

Repro

1. `cd docs; yarn dev`
2. Visit `http://127.0.0.1:5173/framed/UIShell/HeaderSwitcher`


https://user-images.githubusercontent.com/10718366/208265955-2e355051-826a-4bca-8038-5615c3fd07d5.mov

jqlio18:
You are right my approach for this problem was wrong. I changed my strategy and took the same approach as React. The main strategy is to take advantage of the `Grid` and it's `offset` propriety to the main content.  I added the same wrappers to the main content as React too

Source : https://v7-react.carbondesignsystem.com/iframe.html?id=components-ui-shell--header-base-w-navigation-actions-and-side-nav&args=&viewMode=story

metonym:
Fixed via #2521

PR #1591, opened 2022-12-16
feat(notification)!: remove iconDescription props for InlineNotification and ToastNotification

theetrain:
Closes #672 
Closes #1563 

The InlineNotification and ToastNotification components get revamped label treatment for their icon, to support a11y and i18n.

- BREAKING: remove `iconDescription` prop
- Feat: add `statusIconDescription` prop
- Feat: add `closeButtonDescription` prop

I'm also piloting some handy documentation links: "detailed usage" and "see also".

image

PR #1592, opened 2022-12-17
Update docs

gregorw:
It seems that in #1518 some docs were not properly generated.

This is the change that results from running `yarn build:docs`.

### DIFF
--- a/docs/src/COMPONENT_API.json
+++ b/docs/src/COMPONENT_API.json
@@ -8511,7 +8511,8 @@
         {
           "type": "dispatched",
           "name": "change",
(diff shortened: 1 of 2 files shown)


theetrain (approved):
Thanks for putting in the effort for this. The docs weren't generated since I had added the JSDoc comment for this component before we upgraded to sveld v0.18 that generates `@event` docs. I also have it generated in #1591, but can easily rebase.

theetrain:
Released in [v0.71.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.71.0). Thank you!

PR #1596, opened 2022-12-21
Fix `Dropdown` click outside

### LINKED ISSUE #1595: `Dropdown` click outside `iframe` doesn't close
Click outside doesn't work outside iFrame.

https://user-images.githubusercontent.com/3661575/208811784-57fa6dc0-d1cd-4891-b206-5593ec425acd.mp4

jqlio18:
Fixes #1595

Instead of registering the click on the `window` element, we can simply use the `blur` method on the `bx--list-box__field` button. This fix make it works in every situation even in `iframe`.

### DIFF
--- a/src/Dropdown/Dropdown.svelte
+++ b/src/Dropdown/Dropdown.svelte
@@ -281,7 +281,7 @@
             active="{selectedId === item.id}"
             highlighted="{highlightedIndex === i}"
             disabled="{item.disabled}"
-            on:click="{(e) => {
+            on:mousedown="{(e) => {
               if (item.disabled) {
                 e.stopPropagation();
                 return;
--- a/src/Dropdown/Dropdown.svelte
+++ b/src/Dropdown/Dropdown.svelte
@@ -150,14 +150,6 @@
   };
 </script>
 
-<svelte:window
-  on:click="{({ target }) => {
-    if (open && ref && !ref.contains(target)) {
-      open = false;
-    }
-  }}"
-/>
-
 <div
   class:bx--dropdown__wrapper="{true}"
   class:bx--list-box__wrapper="{true}"
@@ -182,10 +174,11 @@
     size="{size}"
     name="{name}"
     aria-label="{$$props['aria-label']}"
-    class="bx--dropdown {direction === 'top' && 'bx--list-box--up'} {invalid &&
-      'bx--dropdown--invalid'} {!invalid &&
-      warn &&
-      'bx--dropdown--warning'} {open && 'bx--dropdown--open'}
+    class="bx--dropdown 
+      {direction === 'top' && 'bx--list-box--up'} 
+      {invalid && 'bx--dropdown--invalid'} 
+      {!invalid && warn && 'bx--dropdown--warning'} 
+      {open && 'bx--dropdown--open'}
       {size === 'sm' && 'bx--dropdown--sm'}
       {size === 'xl' && 'bx--dropdown--xl'}
       {inline && 'bx--dropdown--inline'}
@@ -233,7 +226,6 @@
             open = false;
           }
         } else if (key === 'Tab') {
-          open = false;
           ref.blur();
         } else if (key === 'ArrowDown') {
           if (!open) open = true;
@@ -245,6 +237,7 @@
           open = false;
         }
       }}"
+      on:blur="{() => (open = false)}"
       on:keyup="{(e) => {
         const { key } = e;
         if ([' '].includes(key)) {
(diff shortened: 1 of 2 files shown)


src/Dropdown/Dropdown.svelte
```diff
@@ -245,6 +237,7 @@
           open = false;
         }
       }}"
+      on:blur="{() => (open = false)}"
```
theetrain, line 240: The flaw with adding a blur handler is it will run before a user can select an option. Instead, we can add an event listener to the parent window during `onMount`:

```html
<script>

  // ...

  const pageClickHandler = ({ target }) => {
    if (open && ref && !ref.contains(target)) {
      open = false;
    }
  };

  onMount(() => {
    if (parent) {
      parent.addEventListener("click", pageClickHandler);
    }
  });
</script>

<svelte:window on:click="{pageClickHandler}" />
```

theetrain (changes requested):
Thanks for your effort; see comments below.

src/Dropdown/Dropdown.svelte
jqlio18: I don't think we need to avoid blur completly. You proposition will work if there's only one `iframe` but will fail with mutpliple. To make the code works, we can use the `mousedown` event instead of the `click` [6b84b51](https://github.com/carbon-design-system/carbon-components-svelte/pull/1596/commits/6b84b51238ad3afcb96926dfad925b3ca87d9d3b)

If you want to use `parent` I can make it works with a loop for parents.

theetrain: `on:mousedown` is an interesting solution as it works on both touch and pointer devices. However, this will cause Dropdown items to become selected when the user clicks down instead of `mouseup`, which may be unexpected behaviour to the user.

It seems `on:blur` works against us. The only alternative I see is listening for `on:blur`, but also listen for `mousedown` and set a flag if the user is currently clicking on a list item, though I fear a race condition may occur. I got something that works:

```html
<script>
  // ...

  let active = false;
</script>

<button
  on:blur="{() => {
    if (active) return;
    open = false;
  }}"
>
<!-- -->
</button>

<!-- -->
<ListBoxMenuItem
  on:mousedown="{() => {
    active = true;
  }}"
  on:click="{(e) => {
    // ... assign values
    active = false;
  }}"
/>
```

It keeps the Dropdown 'active' when there's a mousedown. Try that out and let me know what you think. Any thoughts @metonym?

Note, changing code in ListBoxMenuItem will impact consumers MultiSelect and Combobox; but we can work on those once we agree on an interaction data flow.

jqlio18: I made some tests and `onblur` won't work because of the `label` which would cause more problems. I think we'll need to stick with `svelte:window` origin. I'll post a commit with your  `pageClickHandler`.

jqlio18: Sorry for all the commits. I'm still getting used to `git`. Someday I'll understand everything.

theetrain: What sort of issues wth `label` did you run into? Also don't worry about making so many commits, it's often good practice. 😃

jqlio18: The `blur` event is set to the button so if the `label` is clicked, `blur` will be called and will flash the `ListItemMenu`.

jqlio18: @theetrain I added `onDestroy` event to remove the the `eventListener` in case of the component get destroyed. We don't want to overpopulate the `eventListeners` :^)

theetrain (approved):
This looks good to me! Tested on Chrome.

I'll have it released within the next few days. Combobox needs the same treatment; feel free to open a separate PR for that, otherwise we can backlog it.

src/Dropdown/Dropdown.svelte
```diff
@@ -148,15 +148,27 @@
...
     }
-  }}"
-/>
+  };
+
+  onMount(() => {
+    if (parent) {
+      parent.addEventListener("click", pageClickHandler);
+    }
+  });
```
metonym, line 162: Non-blocking: Why not use the return function in `onMount` instead of `onDestroy`?

```ts
onMount(() => {
  if (parent) {
    parent.addEventListener("click", pageClickHandler);
  }

  return () => {
    // do something
  }
});
```

theetrain: I forgot about this feature! I guess this could simplify things assuming there's no difference between `onDestroy` and this implied `unmount`. Either way I agree it's optional.

jqlio18: Never saw the docs about the return function! It's great! I think the `onDestroy` adds more readability but anyway both are good.

theetrain:
This has been released in [v0.72.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.72.0).

Issue #1616, opened 2023-01-07
Typescript improvements


theetrain:
As advised in https://github.com/carbon-design-system/carbon-components-svelte/discussions/1614#discussioncomment-4618257

- [ ] Switch from `svelte.JSX` to `svelte/elements`
- [ ] Decide whether to use `svelte-package` or augment `sveld`

[`svelte-package`](https://kit.svelte.dev/docs/packaging) allows us to write components with `<script lang=ts>` and ship `carbon-components-svelte` with `.svelte` and `.d.ts` pairs. Sveld is likely still needed to generate documentation.

SimpleProgrammingAU:
I tried to use `<script lang=ts>` for `TextInput` and assigned a type to one of the prop variables.

Is there some more setup that needs to happen before TypeScript can be adopted?

```
5:50:05 pm [vite] Internal server error: C:/code/carbon-components-svelte/src/TextInput/TextInput.svelte:10:20 Unexpected token

- Did you forget to add a TypeScript preprocessor? See https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/preprocess.md for more information.
  Plugin: vite-plugin-svelte
  File: C:/code/carbon-components-svelte/src/TextInput/TextInput.svelte:10:20
    8 |     * Set to "char" to enable display the character counter or "word" to display the word count.
    9 |     */
   10 |    export let counter: "char" | "word" | undefined = undefined;
                             ^
   11 |
   12 |    /** Set to `true` to disable the input */
```

theetrain:
Using `lang="ts"` is out of scope for the time being. We may eventually use it after closing #1871 since `@sveltejs/package` can convert `.svelte` files containing `lang="ts"` into respective `.svelte` and `.svelte.d.ts` files; allowing users to consume `carbon-components-svelte`, with type safety, and only optionally needing `vitePreprocess`. 

Please use JSDoc for now, as Sveld can build declarations files and documentation for us.

Issue #1605, opened 2022-12-31
Replace `as` prop with `svelte:element`


metonym:
`svelte:element` was added to Svelte in [3.47](https://github.com/sveltejs/svelte/blob/master/CHANGELOG.md#3470).

Currently, four components expose an `as` prop that allows the consumer to customize the tag. This prop was added years ago long before `svelte:element` was supported. Now that it exists, we can instead use `svelte:element` and allow the consumer to customize the tag element.

Components that expose an `as` prop:

- Button
- Column
- Row
- Grid

theetrain:
Is there a reason why `Button` allows element overrides with `as`? I cannot imagine a situation where you would want anything except for `<button>`, and I couldn't find a rationale in /carbon repository.

metonym:
@theetrain Same. IMO, the `as` prop from `Button` should be removed.

PR #1628, opened 2023-01-15
feat!: remove `as` prop in favor of `svelte:element`

metonym:
This was developed initially to address #1605. It removes the `as` prop, which was added before `svelte:element` was supported in Svelte. The purpose is to allow the consumer to customize the element tag.

However, this should not be merged since it has a flaw.

The issue is that `$$restProps` are spread to the element; the tag attributes in the generated types should be generic to match the tag. This is a limitation of sveld, which does not support generics in props.

Perhaps another reason to switch to TypeScript + `@sveltejs/package` (#1616)


theetrain:
Thanks for exploring this. Usage of `$$restProps` seems essential in order for users to apply standard attributes to a component's canonical element. I'm also exploring svelte-package in #1629 and results look promising so far.

Since `$$props` and `$$restProps` are discouraged, maybe we should use a custom prop such as `buttonAttributes` for `<Button />` and apply its respective HTML element type to that prop. At least it'll look more proper, can be documented, and hopefully can be optimized in the final Svelte output.

metonym:
> maybe we should use a custom prop such as `buttonAttributes` for `<Button />` and apply its respective HTML element type to that prop. At least it'll look more proper, can be documented, and hopefully can be optimized in the final Svelte output.

That's a very elegant solution. +1

PR #1631, opened 2023-01-16
fix(tree-view)!: remove `expanded` property from `TreeNode` interface

### LINKED ISSUE #1630: TreeView ignores expanded property in children object
Docs on TreeView [here](https://carbon-components-svelte.onrender.com/components/TreeView#with-icons) provide an example with `expanded: true` property to expand the tree element - this element is ignored by TreeView and the child is not expanded.

This behaviour can be seen directly in the example in the docs itself:
image

metonym:
Fixes #1630

**Changes**

- breaking: remove `expanded` from the `TreeNode` interface
- fix: remove unused `expanded` prop from internal `TreeViewNodeList`
- docs: remove `expanded` as a property usage

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -4661,7 +4661,6 @@ export interface TreeNode {
   text: any;
   icon?: typeof import("svelte").SvelteComponent;
   disabled?: boolean;
-  expanded?: boolean;
   children?: TreeNode[];
 }
 ```
--- a/docs/src/COMPONENT_API.json
+++ b/docs/src/COMPONENT_API.json
@@ -14522,9 +14522,9 @@
           "ts": "type TreeNodeId = string | number"
         },
         {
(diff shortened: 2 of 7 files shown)


theetrain (changes requested):
Component changes look good, but some documentation changes are needed:

1. Remove `expanded` property in default example paragraph

   image

2. Update the `TreeNode` interface, removing the `expanded?` optional key:

   https://github.com/carbon-design-system/carbon-components-svelte/blob/7701c012a571a7cf7f6180ca144dcc75b36b2154/src/TreeView/TreeView.svelte#L4

metonym:
Added the suggested changes; best reviewed by commit.

theetrain (approved):
This looks good. I consider this a breaking change due to the type changes being reflected on the docs. I don't consider removing the `expanded` prop from `TreeViewNodeList.svelte` as a breaking change since it's internal-facing; users won't directly engage with that component.

theetrain:
This has been released in [v0.72.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.72.0).

Issue #1638, opened 2023-01-25
`parent` is undefined during SSR


ptrxyz:
In the dropdown component, `parent` is referenced (https://github.com/carbon-design-system/carbon-components-svelte/blob/afa692c9e0db235d0e54975281fa89f831109077/src/Dropdown/Dropdown.svelte#L158-L168) in the onMount and onDestroy functions, however this is not defined during SSR in SvelteKit. 

A possible fix that would work in Svelte as well as SvelteKit would be to check for `globalThis.window`:

- during SSR it is `globalThis.window` is undefined
- in the browser `globalThis.window` is defined

So this would work:

```js
  let parent = globalThis.window ?? undefined;

  onMount(() => {
    if (parent) {
      parent.addEventListener("click", pageClickHandler);
    }
  });

  onDestroy(() => {
    if (parent) {
      parent.removeEventListener("click", pageClickHandler);
    }
  });
```

Alternatively checking if `parent` is defined would work in case `globalThis` is not an option (yet Svelte uses `globalThis` since version 3, so it should be save). This one will always work I guess:

```js
if (typeof parent !== "undefined" && parent) {
```

ptrxyz:
Minimal example: https://github.com/ptrxyz/carbon-bug-testing/

theetrain:
I reproduced this as well; the solution is to return a function in `onMount` since `onDestroy` does in fact run during SSR. (@metonym, you were right all along).

As discussed on Discord, @ptrxyz will provide a PR.

theetrain:
This is now released in v0.72.1

Resolved by #1639

metonym:
Thank you both for the quick fix and release.

PR #1639, opened 2023-01-25
Fix for #1638

ptrxyz:
TIL `onDestroy` hook is also run during SSR, but returning a destruction function from `onMount` should do the trick.


theetrain (approved):
Tested with linked packages locally, and it works!

Thank you!

theetrain:
This is now released in v0.72.1

PR #1642, opened 2023-01-29
Carbon v11 Button

### LINKED ISSUE #1641: Button v11

theetrain:
Closes #1641 

- forward focus and blur events
- provide button role for custom elements using 'as'
- BREAKING CHANGE: danger-tertiary and danger-ghost props were renamed to danger--tertiary and danger--ghost (two dashes instead of one)

Follow-ups:

- ComposedModal vertically centres Button copy, but it should be resolved once the Modal adopts v11 SCSS


theetrain:
Todo

- render SCSS as global; it wasn't designed to be scoped and won't work with ButtonSet
- fix pipeline in `feat/carbon-v11`

theetrain:
I decided to keep the `as` prop for now; I'll wait for a response in the flagship repo in https://github.com/carbon-design-system/carbon/issues/13045

At least I added `role="button"` that can be overridden with `$$restProps` so that users of `as` won't ship inaccessible custom buttons.

Tests won't pass for the time-being; we'll need to refactor imports, and svelte-package needs to pick up `typedef` items in JSDoc.

metonym (approved):
Awesome, the minimal diffs made this easy to review

PR #1646, opened 2023-01-31
fix(dropdown): dispatch correct `selectedItem` in `select` event

### LINKED ISSUE #1645: DropDown "selectedItem" is the last selection, not the current
Unsure if this is expected behaviour and the wording is just misleading.  But if you print `console.log(e.detail)` when selecting an element in a DropDown component, the `selectedId` is correct and properly reflects the active selection.  However `selectedItem` appears to be what you _last_ had selected (i.e. the selection you moved away from).

metonym:
Fixes #1645

The `select` event dispatched in `Dropdown` uses the previous `selectedItem` value. One solution is to use the `tick` method to wait for all state changes to be applied.

### DIFF
--- a/src/Dropdown/Dropdown.svelte
+++ b/src/Dropdown/Dropdown.svelte
@@ -146,7 +146,10 @@
   }
 
   const dispatchSelect = () => {
-    dispatch("select", { selectedId, selectedItem });
+    dispatch("select", {
+      selectedId,
+      selectedItem: items.find((item) => item.id === selectedId),
+    });
   };
 
   const pageClickHandler = ({ target }) => {
@@ -164,7 +167,7 @@
       if (parent) {
         parent.removeEventListener("click", pageClickHandler);
       }
-    }
+    };
   });
 </script>
 


src/Dropdown/Dropdown.svelte
```diff
@@ -145,7 +145,8 @@
     highlightedIndex = index;
   }
 
-  const dispatchSelect = () => {
+  const dispatchSelect = async () => {
+    await tick();
     dispatch("select", { selectedId, selectedItem });
   };
```
theetrain, line 151: Overall this works and you're welcome to ship it. Thanks for catching this miss by me.

Would you prefer to remove the `tick` and return the `selectedItem` directly from the `items` array? Although the `tick` is a negligibly short time, I fear it may cause a race condition somehow.

```suggestion
  const dispatchSelect = async () => {
    dispatch("select", {
      selectedId,
      selectedItem: items.find((item) => item.id === selectedId).text,
    });
  };
```

theetrain (approved):

src/Dropdown/Dropdown.svelte
metonym: Amended the commit to use this approach.

theetrain (approved):

theetrain:
This is now available in [v0.72.2](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.72.2).

PR #1651, opened 2023-02-06
docs(button): add example of disabled, icon-only button

metonym:
Inspired by this [message from Discord](https://discord.com/channels/689212587170201628/756226301475684543/1070663479645720636).

> Hi — is there a disabled state for Icon only buttons?  I can see the disabled state for standard buttons and buttons with icon, but can't find the disabled behaviour for Icon only buttons.

I think it would be helpful to showcase what a disabled, icon-only button looks like.

This PR also changes "Disabled button" to "Disabled state" to be consistent with other docs pages.

---

Disabled state


theetrain (approved):
Thanks for the quick win!

Issue #940, opened 2021-12-18
Filterable dropdown triggers focus twice when clicking on arrow icon


alkanna:
Everything's in the title :)
You can easily reproduce it on the filterable example here : https://carbon-components-svelte.onrender.com/components/MultiSelect

DerekRoy:
Also affected by this. To illustrate the issue further: 

1. Click on MultiSelect arrow with filter enabled the event occurs twice flashing the dropdown content, then hiding it.  
2. Clicking anywhere outside the error brings up the multiselect drop down fine.

metonym:
Hot fixed in [v0.50.4](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.50.4)

PR #1654, opened 2023-02-08
fix: remove pointer-events style from multiselect dropdown

### LINKED ISSUE #1647: Filterable Multiselect Pointer Events
As can be in the [filterable multiselect example from the docs site](https://carbon-components-svelte.onrender.com/components/MultiSelect#filterable), hovering over the expander icon retains the text input cursor instead of switching to a pointer cursor.

Viewing the [comparable React component example](https://carbondesignsystem.com/components/dropdown/code/), the React library does have a pointer cursor when hovering on the icon. This is consistent with the other visually similar components such as ComboBox.

This seems to primarily be caused by the setting of pointer-event to none on this line: https://github.com/carbon-design-system/carbon-components-svelte/blob/57b6ea68b53368240b07faff80a8a8157fae0786/src/MultiSelect/MultiSelect.svelte#L469

In my local testing I see no harm from removing this style, but it seems so intentionally put there that I worry I am missing something. Do we know why pointer-events were sometimes disabled for the MultiSelect icon? Other components like the Svelte ComboBox that have similar structures and styles don't have this style. Is the line just a bug or is there something deeper lurking behind it?

I could generate a PR to remove the line and restore the pointer cursor but am hesitant since I feel like I'm missing something.

jorgebv:
Resolves #1647 
This makes the component more consistent with other components that have the dropdown, and also more consistent with the flagship react library.

### DIFF
--- a/src/MultiSelect/MultiSelect.svelte
+++ b/src/MultiSelect/MultiSelect.svelte
@@ -466,7 +466,6 @@
           />
         {/if}
         <ListBoxMenuIcon
-          style="pointer-events: {open ? 'auto' : 'none'}"
           on:click="{(e) => {
             e.stopPropagation();
             open = !open;


metonym:
Adding some context here – looking at the Git blame, this hotfix was added in https://github.com/carbon-design-system/carbon-components-svelte/pull/954 to fix an odd behavior where [clicking the icon](https://github.com/carbon-design-system/carbon-components-svelte/issues/940) for a filterable multiselect would double trigger the open behavior.

Agreed that it should be more consistent – is it possible to resolve the previous bug without the pointer-events workaround?

jorgebv:
Thanks for the context; I should have thought to look at the blame myself. I will look at the issue again with this context in mind as you suggest.

jorgebv:
@metonym I found that the primary cause of #940 was the focus handler of the ListBoxField, which would open the MultiSelect. The pointer events of the ListBoxMenuIcon would then immediately close it.

The flagship react MultiSelect doesn't seem to open on focus, so I removed this behavior from the svelte. Maybe this is a breaking change.

Removing the focus on open caused typing to get lost in the MultiSelect, so I added on:input to the input handler, similar to ComboBox and other components. The on:input handler I have implemented for this MultiSelect is a little basic compared to the ComboBox because I didn't quite see what the purpose of everything the ComboBox is doing on:input. Maybe the on:input handler in MultiSelect is insufficient and needs to be more similar to ComboBox. Feedback welcome.

I also added a little bit of styling for when the input is focused, to get the highlighting of the filterable MultiSelect which was missing before.

This became a little bigger than I wanted it to and a little outside my comfort zone. Maybe some of my approach is undesirable. Please let me know if you see any problems with the overall approach.

metonym:
Thank you for submitting the PR. I investigated this issue again. This issue was already resolved in #2231 (merged Oct 17, 2025).

The `pointer-events: none` style was originally added in #954 to fix a double-trigger bug (#940) where clicking the chevron icon in filterable MultiSelect would fire handlers twice.

PR #2231 completely refactored the filterable MultiSelect to address accessibility violations (#2172). As part of this refactor, the DOM structure was reorganized, which naturally resolved the double-trigger issue without needing `pointer-events: none`. The style was removed in commit c4c335ae.

Issue #1621, opened 2023-01-10
Standardize props and events


theetrain:
We have several props and dispatched event names that are reused across components. Let's improve their consistency in naming and intent.

## Prop names

WIP

Props that have different intent follow different aspirational rules:

- Flags: props that toggle a feature via `true` or `false` values are represented as plain adjectives without an `is` prefix. Examples: `expandable`, `selectable`, `selected`. Prefer to have them default to `false` for easy truthy flagging. i.e. absense of `selectable` implies `selectable == false`, but the presence of `selected` toggles to `true`.

## Dispatched events

WIP

## Forwarded events

WIP

## `$$restProps`

See https://github.com/carbon-design-system/carbon-components-svelte/issues/1064#issuecomment-1375954124

metonym:
Linking this: https://github.com/carbon-design-system/carbon-components-svelte/discussions/1420

brunnerh:
Also related:
- #1325 
- #534

metonym:
I agree that we should only forward useful events.


A lot of components have a wrapping `div` in which many events are forwarded to. We should not forward events in these cases as it generates a lot of boilerplate code. A user can always wrap the component with a `div` and listen to events as they bubble up.

Furthermore, it can be confusing if we forward events to different elements.

Example:

```svelte
<div
    on:click
    on:mouseover
    on:mouseenter
    on:mouseleave
>
   <input on:input />
</div>
```

theetrain:
@metonym I added "Prop ordering" above. 

On another note, I noticed `<Button>` has a `skeleton` prop whereas other components like `<TextInput>` has a separate `<TextInputSkeleton>` component. Should we prefer props or components? I'd say components are better for tree-shaking. The user experience comparison would be:

As props:

```html
<form method="POST" action="?/deleteProfile" use:enhance>
  <Button
    type="submit"
    kind="tertiary"
    size="lg"
    id="abc123"
    skeleton={!ready}
  />
    Delete Profile
  </Button>
</form>
```

As components:

```html
{#if ready}
  <form method="POST" action="?/deleteProfile" use:enhance>
    <Button
      type="submit"
      kind="tertiary"
      size="lg"
      id="abc123"
      skeleton={!ready}
    />
      Delete Profile
    </Button>
  </form>
{:else}
  <ButtonSkeleton size="lg" />
{/if}
```

metonym:
[...]

PR #1655, opened 2023-02-09
fix(types): add missing `$$restProps` for `Checkbox`, `Filename`, `FluidForm`

metonym:
A user [reported in Discord](https://discord.com/channels/689212587170201628/1048235099583418448/1073080988957151343) that `$$restProps` types in `Checkbox` are missing from the type definitions.

https://github.com/carbon-design-system/carbon-components-svelte/issues/1621 proposes that we remove `$$restProps` for both clarity and performance reasons. This PR is a hotfix that addresses the existing bugs pre-v1.

The solution is to manually add a `@restProps` tag that sveld uses to generate types from.

### DIFF
--- a/docs/src/COMPONENT_API.json
+++ b/docs/src/COMPONENT_API.json
@@ -924,7 +924,7 @@
         { "type": "forwarded", "name": "blur", "element": "input" }
       ],
       "typedefs": [],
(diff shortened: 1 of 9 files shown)


theetrain (approved):

theetrain:
Now available in [v0.72.3](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.72.3)

PR #1666, opened 2023-02-25
fix: `readonly` should override `error` behavior in `TextInput`, `NumberInput`

### LINKED ISSUE #1664: Text input: icons overlap when `readonly` and `invalid`
If a text input is `readonly` and `invalid` at the same time, the icons overlap:

```tsx
<TextInput
  name="mybox"
  readonly
  invalid={true}
/>
```

will result in this:

image

(see the icons on the right overlapping)

metonym:
Fixes #1664

This primarily fixes a bug where setting both `readonly` and `invalid` in `TextInput` would display two icons. The solution is to separate the `readonly` state from the `invalid/warn` states. This PR applies a similar logic to `NumberInput`, although icons will not clash in that component.

Furthermore, the `readonly` state will override the `error` state. That is, if both `readonly` and `error` are set, the `readonly` behavior will prevail.

The diff is best reviewed with [whitespace hidden](https://github.com/carbon-design-system/carbon-components-svelte/pull/1666/files?diff=split&w=1).

See a before/after:

```svelte
<TextInput
  labelText="TextInput (readonly + invalid)"
  value="Text"
  readonly
  invalid
  invalidText="Invalid text"
/>

<FluidForm>
  <TextInput
    labelText="Fluid TextInput (readonly + invalid)"
    value="Text"
    readonly
    invalid
    invalidText="Invalid text"
  />
</FluidForm>

<NumberInput
  label="NumberInput (readonly + invalid)"
  value={0}
  readonly
  invalid
  invalidText="Invalid text"
/>
```

Before/After

### DIFF
--- a/src/NumberInput/NumberInput.svelte
+++ b/src/NumberInput/NumberInput.svelte
@@ -128,7 +128,7 @@
   $: incrementLabel = translateWithId("increment");
   $: decrementLabel = translateWithId("decrement");
   $: error =
-    invalid ||
+    (invalid && !readonly) ||
     (!allowEmpty && value == null) ||
     value > max ||
     (typeof value === "number" && value < min);
@@ -191,8 +191,8 @@
         type="number"
         pattern="[0-9]*"
         aria-describedby="{errorId}"
-        data-invalid="{invalid || undefined}"
-        aria-invalid="{invalid || undefined}"
+        data-invalid="{(error) || undefined}"
+        aria-invalid="{(error) || undefined}"
         aria-label="{label ? undefined : ariaLabel}"
         disabled="{disabled}"
         id="{id}"
--- a/src/NumberInput/NumberInput.svelte
+++ b/src/NumberInput/NumberInput.svelte
@@ -211,16 +211,17 @@
         on:blur
         on:paste
       />
-      {#if invalid}
-        <WarningFilled class="bx--number__invalid" />
-      {/if}
-      {#if !invalid && warn}
-        <WarningAltFilled
-          class="bx--number__invalid bx--number__invalid--warning"
-        />
-      {/if}
       {#if readonly}
         <EditOff class="bx--text-input__readonly-icon" />
+      {:else}
+        {#if invalid}
+          <WarningFilled class="bx--number__invalid" />
+        {/if}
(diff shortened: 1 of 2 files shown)


theetrain (commented):
Since the React library hides the `invalid` red border when `readonly` is set; should we do the same? I can't imagine there being something that's both invalid and readonly. Plus having it this way sets `aria-invalid="true"`, but there's nothing the user can do to resolve it since it's read-only.

metonym:
Nice catch. I added a new commit that does not mark `readonly` inputs as invalid, removing both the `aria-invalid` attribute and red outline.

I also tested with the fluid `TextInput`. Now – even if both `readonly` and `invalid` are set, the `readonly` behavior will prevail:

After

metonym:
I've updated my original comment with the new before and after screenshot.

theetrain (approved):
Looks good! During testing I found some non-regression issues and put them in separate tickets.

If that sounds good to you, I deploy this within the next few days (unless you merge and deploy before I do).

See: #1667 #1668

theetrain:
Fixed in [v0.73.2](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.73.2).

Addeuz:
This change breaks how we use the TextInput component, why would an input that is `readonly` not be able to be invalid? That is kind of weird behaviour. I see the clash in icons is weird. But you could either hide the `readonly` icon when the input is invalid or not show the invalid icon, but still show the red border and `invalidText`.

Now I have to hack the input to be readonly manually. Since we rely on a `readonly` input being invalid to show to a user that the data we read from a user submitted file is invalid. 

So, I believe a mix between readonly and invalid should be possible?

theetrain:
Hi @Addeuz 

According to the flagship designs, it's not possible to have `readonly` and `invalidText` at the same time. You can see in the React demo here: https://react.carbondesignsystem.com/?path=/story/components-textinput--playground&args=invalid:!true;readOnly:!true

I assume this is deliberate since an error message cannot be remediated by the user since the read-only text input cannot be edited. If you believe this design choice should be reassessed, I suggest opening an issue in the flagship repository: https://github.com/carbon-design-system/carbon

PR #1691, opened 2023-03-16
fix text overflow when progressIndicator space-equally

willy1920:
when progressIndicator enable space-equally text became overflow
before fix
before
after
after


brunnerh:
Issue like this are supposed to be solved by the styles provided by the flagship package `carbon-components`, not inline styles.
Maybe some classes are just missing or there is an issue with the element hierarchy.

willy1920:
Hi! after checking upstream, I just found out this fix requires carbon v11. Thank you.

PR #1696, opened 2023-03-21
fix(slider): no label is rendered when the labelText property is not specified.

### LINKED ISSUE #1682: The label element will be rendered in the Slider component even if the title parameter is not passed
https://github.com/carbon-design-system/carbon-components-svelte/blob/d2e471bbecbbaf24cad8fe1589e6363354a35471/src/Slider/Slider.svelte#L142

Because the label element had a 0.5 rem marginBottom, I was unable to align the Slider with the other form controls.

image

siaikin:
label elements are not rendered when no `labelText` or `labelText` slot is passed.

Close #1682 

#### Before
image

#### After
image

### DIFF
--- a/src/Slider/Slider.svelte
+++ b/src/Slider/Slider.svelte
@@ -139,16 +139,18 @@
   on:mouseenter
   on:mouseleave
 >
-  <label
-    for="{id}"
-    id="{labelId}"
-    class:bx--label="{true}"
-    class:bx--label--disabled="{disabled}"
-  >
-    <slot name="labelText">
-      {labelText}
-    </slot>
-  </label>
+  {#if labelText || $$slots.labelText}
+    <label
+      for="{id}"
+      id="{labelId}"
+      class:bx--label="{true}"
+      class:bx--label--disabled="{disabled}"
+    >
+      <slot name="labelText">
+        {labelText}
+      </slot>
+    </label>
+  {/if}
   <div
     class:bx--slider-container="{true}"
     style="{fullWidth ? 'width: 100%' : undefined}"


metonym (commented):
I think this is a necessary fix. However, for accessibility, I would suggest adding a prop `hideLabel` to visually hide the label instead of not rendering it altogether. This is a pattern used in many of the input components. See [Dropdown](https://github.com/carbon-design-system/carbon-components-svelte/blob/4960504ace3a92a131b08762ac4fcbc8d979340c/src/Dropdown/Dropdown.svelte#L189).

```svelte
class:bx--visually-hidden="{hideLabel}"
```

theetrain:
Keep in mind the flagship library doesn't allow Slider to hide its label https://react.carbondesignsystem.com/?path=/story/components-slider--playground

I don't mind the addition since it's already supported in other inputs with the same heights.

siaikin:
I noticed that this was a common behavior even in flagship libraries (referring to `marginBottom` with `0.5rem`) and I had to ask them if this was intentional. I hope to hear from them soon XD.

> Keep in mind the flagship library doesn't allow Slider to hide its label https://react.carbondesignsystem.com/?path=/story/components-slider--playground
> 
> I don't mind the addition since it's already supported in other inputs with the same heights.

siaikin:
> I think this is a necessary fix. However, for accessibility, I would suggest adding a prop `hideLabel` to visually hide the label instead of not rendering it altogether. This is a pattern used in many of the input components. See [Dropdown](https://github.com/carbon-design-system/carbon-components-svelte/blob/4960504ace3a92a131b08762ac4fcbc8d979340c/src/Dropdown/Dropdown.svelte#L189).
> 
> ```svelte
> class:bx--visually-hidden="{hideLabel}"
> ```

thanks for you suggestions. I think what needs to be fixed is to remove the `marginBottom` property with `0.5rem` on the label. Instead of removing the entire label.
but because of https://github.com/carbon-design-system/carbon-components-svelte/pull/1696#issuecomment-1478863383 , I will decide whether to continue this pr after communication.

siaikin:
> I think this is a necessary fix. However, for accessibility, I would suggest adding a prop `hideLabel` to visually hide the label instead of not rendering it altogether. This is a pattern used in many of the input components. See [Dropdown](https://github.com/carbon-design-system/carbon-components-svelte/blob/4960504ace3a92a131b08762ac4fcbc8d979340c/src/Dropdown/Dropdown.svelte#L189).
> 
> ```svelte
> class:bx--visually-hidden="{hideLabel}"
> ```

thanks for you suggestions. I think what needs to be fixed is to remove the `marginBottom` property with `0.5rem` on the label. Instead of removing the entire label.
but because of https://github.com/carbon-design-system/carbon-components-svelte/pull/1696#issuecomment-1478863383 , I will decide whether to continue this pr after communication.

siaikin:
> Keep in mind the flagship library doesn't allow Slider to hide its label https://react.carbondesignsystem.com/?path=/story/components-slider--playground
> 
> I don't mind the addition since it's already supported in other inputs with the same heights.

I noticed that this was a common behavior even in flagship libraries (referring to `marginBottom` with `0.5rem`) and I had to ask them if this was intentional. I hope to hear from them soon XD.

PR #1700, opened 2023-03-24
docs(theme): replace custom `Theme` component with packaged version

gregorw:
The custom `Theme` component in docs is not necessary. We can remove it. Instead, let’s use the packaged version of the same component. As the saying goes: “Eat your own dog food.”


docs/src/pages/_layout.svelte
```diff
@@ -38,7 +39,7 @@
 <!-- routify:options bundle=true -->
 <svelte:window bind:innerWidth />
 
-<Theme persist>
+<Theme persist bind:theme={$theme}>
```
metonym, line 42: Could you run `yarn lint` here? I think it will make a style change.

metonym (commented):
Nice catch. LGTM pending the lint.

docs/src/pages/_layout.svelte
gregorw: Sure.

metonym (approved):

Issue #1649, opened 2023-02-01
BUG on docs page


dsm-72:
Please view this page
https://carbon-components-svelte.onrender.com/components/UIShell

Notice that the right side bar is off the screen.
Not sure if I should use the UI library if the UI library breaks.

brunnerh:
I have no idea what you are referring to.
Please at least describe what you mean and ideally add a screenshot + browser version.

dsm-72:
@brunnerh 
<img width="1435" alt="Screenshot 2023-02-04 at 9 18 14 AM" src="https://user-images.githubusercontent.com/100158155/216772552-7ea1c938-3afc-4f35-b187-74e70402241d.png">

dsm-72:
Note this holds for Firefox too!
![Screenshot 2023-02-04 at 09-50-59 UIShell](https://user-images.githubusercontent.com/100158155/216773912-3d91e536-91b5-4e06-ac44-8fbc191a90eb.png)

brunnerh:
This only happens on "smaller" screens which is why I did not see it.
Some page content is not shrinking correctly (which has nothing to do with the component library itself, it's just the docs).

dsm-72:
@brunnerh strange. Can the docs be fixed then? As "smaller" is like full-size laptop (14") and it is only on this page, suggesting it is the UIShell component taking up too much horizontal space (e.g. using width:vh)

brunnerh:
Well, I work on 27" screens (of course the resolution and system/browser scale settings are more important than the physical size).
Does not look like this is isolated; the [`Accordion`](https://carbon-components-svelte.onrender.com/components/Accordion) page also overflows for me if I reduce the width.

Should be fixable by adjusting the layout.

theetrain:
I've noted the bug and we'll fix it in time. Since responsive design impacts usability, I agree it should be fixed.

chipnetics:
I can see similar behaviour on https://carbon-components-svelte.onrender.com/components/OrderedList as well, the numbering is outside the bounding box. That said, this probably isn't a component that gets much use anyhow....

PR #1706, opened 2023-03-26
Fix horizontal scrollbars on small screens

gregorw:
This patch addresses #1649 to fix documentation layout. In fact there are multiple issues with the current layout:

1. Flex layout might have an issue with overflowing content. See [CodePen](https://codepen.io/gregorw/full/gOdqreo).
2. The props table at the bottom of each page doesn’t overflow and stretches the layout.

This minimal change fixes code snippets in such a way that they are wrapped on small screens, since `overflow-x: scroll` doesn’t seem to work with the flexbox layout used. See code pen for problem description and solution A. 

**Before**

image

**After**

image

This fixes horizontal scrollbars on all pages that don’t have other issues such as a large non-overflowing props table at the bottom. It also makes code snippets in the props table break a bit more.

**Before**

image

**After**

image

The prop table at the bottom needs a separate PR. Possible solutions are that we get overflow/scroll to work or that we change to the mobile table of contents earlier. Maybe using a non-flexbox layout would solve some of the problems, too. See solution B of code pen above.

### DIFF
--- a/docs/src/components/InlineSnippet.svelte
+++ b/docs/src/components/InlineSnippet.svelte
@@ -6,7 +6,12 @@
 </script>
 
 <div>
-  <CodeSnippet code="{code}" type="inline" copy="{(text) => copy(text)}" />
+  <CodeSnippet
+    code="{code}"
+    type="inline"
+    copy="{(text) => copy(text)}"
+    class="code-override"
+  />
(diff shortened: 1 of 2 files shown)


gregorw:
I have found a better way…

PR #1707, opened 2023-03-26
docs: fix layout for smaller screens

### LINKED ISSUE #1649: BUG on docs page

gregorw:
This patch fixes #1649. The documentation layout is currently broken. The fix adds scroll behavior to:

- Code snippets
- Component props table

Like this the layout is no longer overstretched.

**Before**

image

image

**After**

image

image

This is a better fix than #1706 which is inspired by https://defensivecss.dev/tip/flexbox-min-content-size.

### DIFF
--- a/docs/src/global.css
+++ b/docs/src/global.css
@@ -33,6 +33,11 @@ html[theme="g90"] .code-override {
   overflow-x: auto;
 }
 
+/* Addig this to the layout grid fixes overstretching. */
+.fix-overflow {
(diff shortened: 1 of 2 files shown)


metonym (approved):
Brilliant – thank you.

PR #1710, opened 2023-03-27
docs: fix redundant css class

gregorw:
The same class is defined twice.


metonym (approved):

theetrain:
Now available in [v0.74.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.74.0)

PR #1732, opened 2023-05-03
Resolve `a11y` warnings from Svelte version 3.58

metonym:
Now that Svelte version [3.58](https://github.com/sveltejs/svelte/blob/e45a1e05a341e95d30ddb097bc144d8d6f5b3573/CHANGELOG.md#3580) has been released, a new slew of static a11y warnings needs to be addressed.

To repro the warnings, run the following at the root of this repository on `master`:

```sh
yarn add -D svelte; yarn build:lib;
# ... a11y warnings emitted
```

This PR is best reviewed by commit because it actually contains a fix.

For each warning, I referenced the upstream Carbon React implementation. There are a couple of cases where the a11y warning can be ignored. There is one fix for `ListBoxMenuItem`. For the `tabIndex` -> `tabindex` capitalization, I don't think there is any user-facing change (Svelte Language Server does not seem to pick up `tabIndex`).

@theetrain If approved, should this PR *not* be squash merged? Will `standard-version` automatically pick up the fix for de921e63ec6bf6862f5e4f8cbb042cea939f8ec9?

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -22,7 +22,7 @@
     "prism-svelte": "^0.4.7",
     "prismjs": "^1.28.0",
     "remark-slug": "^6.0.0",
-    "svelte": "^3.57.0",
+    "svelte": "^3.58.0",
     "vite": "^3.0.9"
   },
   "routify": {
--- a/docs/yarn.lock
+++ b/docs/yarn.lock
@@ -165,7 +165,7 @@ bufferutil@^4.0.1:
     node-gyp-build "~3.7.0"
 
 carbon-components-svelte@../:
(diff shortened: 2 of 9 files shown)


theetrain (approved):
Looks good! I'll squash commits and will modify the generated changelog since it's still semi-automatic. 

If we want to fully automate releases then we'll have to be diligent about commit names, or adopt changesets.

theetrain:
Now available in [v0.74.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.74.0)

metonym:
Thank you!

PR #1733, opened 2023-05-06
feat(pagination-nav): add `tooltipPosition` prop

### LINKED ISSUE #1656: add `tooltipPosition` prop for PaginationNav component
bad case:
image

src:
```
node_modules/_carbon-components-svelte@0.72.2@carbon-components-svelte/src/PaginationNav/PaginationNav.svelte
```

add:
```svelte
<script>
    export let tooltipPosition = "bottom";
</script>

<Button
    kind="ghost"
    tooltipAlignment="center"
    tooltipPosition="{tooltipPosition}"
    iconDescription="{forwardText}"
    disabled="{!loop && page === total}"
/>
```

SimpleProgrammingAU:
Added the following options for `tooltipPosition` for the [`paginationNav`](https://github.com/carbon-design-system/carbon-components-svelte/blob/b9aaa3adcaf87e34ac3d45a2e39f1db92ebe1b33/src/PaginationNav/PaginationNav.svelte) component as requested in #1656:

- `"left"`
- `"right"`
- `"top"`
- `"bottom"`
- `"outside"`
- `"inside"`

Close #1656

This is my first pull request for this project. Lemme know if I've messed anything up 👍


src/PaginationNav/PaginationNav.svelte
```diff
@@ -76,7 +82,11 @@
       <Button
         kind="ghost"
         tooltipAlignment="center"
-        tooltipPosition="bottom"
+        tooltipPosition="{tooltipPosition === 'inside'
```
theetrain, line 85: Since this passes `tooltipPosition` to `Button`, perhaps all we need is
`tooltipPosition="{tooltipPosition}"`. I don't think `outside` or `inside` has
intuitive meaning, so we can leave out those options.

Everything else looks good 👍

src/PaginationNav/PaginationNav.svelte
```diff
@@ -147,7 +157,11 @@
...
```
theetrain, line 160: Same as above

src/PaginationNav/PaginationNav.svelte
SimpleProgrammingAU: I wasn't going to push back, but then I realised: without `inside` and `outside` there is no way to achieve that layout without adding an additional prop like `leftTooltipPosition` and `rightTooltipPosition` to the `PagionationNav` component.

I haven't actually used this component in production before, but I can see times where constrained layouts may force the need for the inside and outside options.

image

image

What do you think?

theetrain: Looking at this code:

```html
<Button
        tooltipPosition="{tooltipPosition === 'inside'
          ? 'right'
          : tooltipPosition === 'outside'
          ? 'left'
          : tooltipPosition}"
>
```

When `tooltipPosition` is `inside`, it will set `tooltipPosition="right"`.
When it is `outside`, then it sets `tooltipPosition="left"`. 

Since the underlying effect is `right` or `left`, I'm not sure how aliasing that with `outside` and `inside` can be used. 

>I can see times where constrained layouts may force the need for the inside and outside options.

Can you provide an example?

SimpleProgrammingAU: Did the images not come through in my previous message demoing "outside"?

If the dev passes in "left" **or** "right" they get both labels on the same side. There's no way for the dev to pass in "left" **and** "right".

src/PaginationNav/PaginationNav.svelte
```diff
@@ -23,6 +23,12 @@
   /** Specify the backward button text */
   export let backwardText = "Previous page";
 
+  /**
+   * Set the position of the tooltip relative to the icon
```
theetrain, line 27:
```suggestion
   * Set the position of the tooltip relative to the pagination buttons.
```

theetrain (commented):
@SimpleProgrammingAU I'd say this is good to go, depending on how you feel about this one prop definition change.

@metonym though the React component doesn't offer control over the PaginationNav's tooltip positioning, nor do the docs have strict guidelines on tooltip positioning, I feel this enhancement is good. The `inside` and `outside` options are interesting, and I suppose they can be included as well.

Outside example:

image
image

Inside example:

image
image

metonym:
@theetrain Makes sense to me!

@SimpleProgrammingAU Thank you for working on this 🙏

SimpleProgrammingAU:
Agree with documentation change; added commit. Should be ready for merge (?) 👍

theetrain (dismissed):
🚀

PR #1738, opened 2023-05-28
fix(expandable-tile): set tile height using resize observer

### LINKED ISSUE #1730: [ExpandableTile] height should accommodate large above-the-fold content
I have a simple expandable tile with lots of text in the above section, which is cut off:
image

When I expand the tile and collapse it again, it shows everything:
image

How can I achieve that the height from the second picture is always used even before interacting with it?
I know I might be able to use `tileMaxHeight`, but I don't know the height of the text beforehand, as they are different from text to text.

My tile:
```
<ExpandableTile>
	<div slot="above">
		<p><strong>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Curabitur porttitor, dolor quis rhoncus venenatis, turpis sapien auctor metus, at porttitor ex quam in justo. Duis eget tortor pretium, gravida sapien in, aliquam leo. Ut imperdiet aliquam dui, non laoreet purus varius vel. Nullam dictum magna leo, at vehicula est bibendum eu.</strong></p>
	</div>
</ExpandableTile>
```

metonym:
Fixes #1730

Currently, the `ExpandableTile` does not dynamically set the height of the above-the-fold content. In cases where the content is dynamic (a lot of text), this can result in clipped content.

This PR aims to fix this problem by using the Resize Observer to set the height of the tile, and adjust it if the height changes. It implements the [approach of the upstream React library](https://github.com/carbon-design-system/carbon/blob/97b6ca3e90fd9f95869792e4e6b27ea236cc7932/packages/react/src/components/Tile/Tile.js#L500-L509) almost verbatim.

In terms of behavior, this should not be a breaking change since you can still specify a custom tile height on the "above" and "below" slots.

**Changes**

- Fix `ExpandableTile` to use Resize Observer to set the tile max height
- Rework the docs to promote *not* setting a height; add an example for customizing the tile heights

---

## Example

In the following example, I am clicking a button that doubles the text in the above-the-fold tile. Note how the tile height adjusts itself dynamically.

https://github.com/carbon-design-system/carbon-components-svelte/assets/10718366/fa875e6e-124e-4eeb-a470-570c2b603fda

## Issue #1730 (before)

https://github.com/carbon-design-system/carbon-components-svelte/assets/10718366/33e0ca54-c519-4d50-8d47-7d63c2d2b9c0

## Issue #1730 (after)

https://github.com/carbon-design-system/carbon-components-svelte/assets/10718366/02dd432c-cc31-4314-aee4-2c2f411c81f3

### DIFF
--- a/docs/src/pages/components/ExpandableTile.svx
+++ b/docs/src/pages/components/ExpandableTile.svx
@@ -3,7 +3,29 @@
   import Preview from "../../components/Preview.svelte";
 </script>
 
-## Default (unexpanded)
+## Default
+
+This component uses a [resize observer](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) to determine the height of the above-the-fold content.
+
+It's unexpanded by default.
+
+<ExpandableTile>
+  <div slot="above">
+    <div>
+      Above the fold content here
+    </div>
+    <br />
+    <br />
+    <div>
+      More above the fold content
+    </div>
+  </div>
+  <div slot="below">Below the fold content here</div>
+</ExpandableTile>
+
+## Custom tile heights
+
+Set a custom height for the tiles on the "above" and "below" slots.
 
 <ExpandableTile>
   <div slot="above" style="height: 10rem">Above the fold content here</div>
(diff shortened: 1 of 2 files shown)


metonym:
Thanks for reviewing! 🙏

PR #1741, opened 2023-06-04
fix(types): allow `data-*` attributes for props forwarded to HTML elements

### LINKED ISSUE #1740: TypeScript: `data-*` attributes produce type errors using `svelte-check` version 3
Flagged in Discord: https://discord.com/channels/689212587170201628/1048235099583418448/threads/1112376518862188675

See https://github.com/sveltejs/language-tools/issues/1825

Many components forward `$$restProps` to the underlying HTML element. Data attributes (e.g., `data-test-id`) should be valid.

`svelte-check` version 3 applies a [new transform](https://github.com/sveltejs/language-tools/issues/1352) which marks `data-*` attributes as errors.

This library still uses `svelte-check` version 2; the first step would be to upgrade `svelte-check` to the latest version to reproduce this error. See #1739

Screen Shot 2023-06-04 at 8 07 10 AM

metonym:
Fixes issue #1740, related PR #1739

Best reviewed by commit.

I [patched `sveld`](https://github.com/carbon-design-system/sveld/releases/tag/v0.18.1) to allow `data-*` attributes to be passed to components that extend HTML elements.

As a result, `data-*` attributes should be valid, while other attributes not included in the HTML attributes or component props should still produce a type error:

Screen Shot 2023-06-04 at 9 52 05 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "rollup-plugin-terser": "^7.0.2",
     "sass": "^1.49.11",
     "standard-version": "^9.5.0",
-    "sveld": "^0.18.0",
+    "sveld": "^0.18.1",
     "svelte": "^3.58.0",
     "svelte-check": "^2.8.1",
     "typescript": "^4.7.4"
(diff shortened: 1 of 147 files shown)


metonym:

PR #1750, opened 2023-06-11
feat: make `titleText` slottable in `MultiSelect`, `ComboBox`

### LINKED ISSUE #1747: MultiSelect does not have a label/titleText slot
Most of the other components, labels text is defined as a slot, hence I can do something like this to add icons to labels or to make labels bold: 
```
<NumberInput value={0} >
  <p slot="label"><strong>Enter a number</strong> <Add /></p>
</NumberInput>

<Select>
  <SelectItem value="white" />
  <SelectItem value="g10" />
  <SelectItem value="g80" />
  <SelectItem value="g90" />
  <SelectItem value="g100" />
  <p slot="labelText"><strong>Select a carbon theme<strong> <Add /></p>
</Select>
```

but MultiSelect component label(titleText) is not available as a slot

sachinbhutani:
Fix #1747


metonym (approved):
Looks good – thanks for also applying this to the `ComboBox`.

One small change I made was to update the conditional wrapping the `label`, since the label is only rendered if `titleText` is truthy:

```diff
- {#if titleText}
+ {#if titleText || $$slots.titleText}
```

PR #1759, opened 2023-07-03
fix(date-picker): hide `helperText` if invalid or warning state

benperiton:
Hides the helperText if the field is invalid or warn


metonym (approved):
LGTM, thanks!

metonym:

PR #1764, opened 2023-07-11
feat(ui-shell): make `company` in `Header` slottable

p0lish:
I've noticed that the ui shell has a little glitch during load in our app.

image

The `Â` disappearing after load but it causes a glitch.

The root cause is an `&nbsp;` character in the company text span.
Maybe somebody else faced with this glitch too, so I created a small improvement for that.

### DIFF
--- a/src/UIShell/Header.svelte
+++ b/src/UIShell/Header.svelte
@@ -103,9 +103,3 @@
   </a>
   <slot />
 </header>
-
-<style>
-  .bx--header__name--prefix {
-    padding-right: 0.2rem;
-  }
-</style>
--- a/src/UIShell/Header.svelte
+++ b/src/UIShell/Header.svelte
@@ -94,13 +94,16 @@
     {...$$restProps}
     on:click
   >
-    {#if company}
-      <span class:bx--header__name--prefix="{true}">{company}</span>
+    {#if company || $$slots.company}
+      <span class:bx--header__name--prefix="{true}"
+        ><slot name="company">{company}&nbsp;</slot></span
+      >
     {/if}
     <slot name="platform">{platformName}</slot>
   </a>
   <slot />
 </header>
+
 <style>
   .bx--header__name--prefix {
     padding-right: 0.2rem;
--- a/src/UIShell/Header.svelte
+++ b/src/UIShell/Header.svelte
@@ -95,9 +95,14 @@
     on:click
   >
     {#if company}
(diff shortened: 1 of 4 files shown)


brunnerh:
That just looks like an encoding issue, did you set a/the correct `charset` in the `content-type` header or a `meta` tag?

metonym (changes requested):
WDYT about making the `company` prop slottable so the end user can override this behavior?

```svelte
{#if company || $$slots.company}
  <span class:bx--header__name--prefix="{true}"><slot name="company">{company}&nbsp;</slot></span>
{/if}
```

p0lish:
> WDYT about making the `company` prop slottable so the end user can override this behavior?
> 
> ```svelte
> {#if company || $$slots.company}
>   <span class:bx--header__name--prefix="{true}"><slot name="company">{company}&nbsp;</slot></span>
> {/if}
> ```

Hey, I think its a good idea. Thanks.

types/UIShell/Header.svelte.d.ts
```diff
@@ -29,6 +29,7 @@ export interface HeaderProps
 
   /**
    * Specify the company name
+   * Alternatively, use the named slot "company" (e.g., <span slot="company">...</span>)
```
brunnerh, line 32: Might want to enclose the HTML example in `` ` ``, as JS doc is commonly interpreted as Markdown.
Would also terminate line above with a `.`, potentially followed by two spaces, which indicates a manual line break.

You e.g. currently get this in VS Code:
image

src/UIShell/Header.svelte
```diff
@@ -94,10 +94,18 @@
...
+      <span class:bx--header__name--prefix="{true}"
+        ><slot name="company">{company}&nbsp;</slot></span
+      >
     {/if}
     <slot name="platform">{platformName}</slot>
   </a>
   <slot />
 </header>
+
+<style>
```
metonym, line 107: The style override can be removed

brunnerh (approved):

brunnerh:
Ahhh, the doc fixes should of course have been made directly in `Header.svelte`, they got reverted by the docs re-build.

metonym:
@brunnerh Nice catch – I can add that back in.

metonym:
Released in [v0.77.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.77.0)

PR #1777, opened 2023-07-23
feat(slider): add `hideLabel` prop

### LINKED ISSUE #1682: The label element will be rendered in the Slider component even if the title parameter is not passed

metonym:
Closes #1682, closes #1696

Based on #1696, but uses the approach to visually hide the label for accessibility.


metonym:

PR #1778, opened 2023-07-23
fix(file-uploader): do not render empty element if `labelTitle`, `labelDescription` not provided

### LINKED ISSUE #1775: FileUploader: Unset label/description cause blank space
There are two elements for the properties `labelTitle` and `labelDescription` which always exist, even if the corresponding property is not set. This can cause unnecessary blank space internally or in relation to other content.

image
image

Maybe the elements should be hidden or removed if the property is unset.

metonym:
Fixes #1775

In this case, the `labelTitle` and `labelDescription` are not accessible, so they can be conditionally rendered if truthy.

### DIFF
--- a/src/FileUploader/FileUploader.svelte
+++ b/src/FileUploader/FileUploader.svelte
@@ -103,18 +103,22 @@
   on:mouseenter
   on:mouseleave
 >
-  <p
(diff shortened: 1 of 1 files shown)


metonym:

PR #1779, opened 2023-07-23
fix(types): improve formatting of prop comments

metonym:
Pass at improving formatting of prop comments for display in IDEs.

---

Screenshot 2023-07-23 at 12 18 11 PM


metonym:

PR #1783, opened 2023-07-23
fix(loading)!: remove redundant `description` label

### LINKED ISSUE #1670: Loading and InlineLoading — find a better way to announce description
Currently on Safari + VoiceOver, the following code:

```html
<InlineLoading description="Finding products..." />
```

Announces:

> Active loading indicator. Active loading indicator. Active loading indicator.

A few ways we can improve this:

- toggle `aria-busy` depending on state
- have the SVG icon's `<title>` indicate the loading state in English, and can be overridden with a prop for every state
- ensure the `description` prop gets announced

metonym:
Fixes #1670

Revisited the `Loading` and `InlineLoading` components, using the upstream React implementation as the primary reference.

The `description` prop is already set on the `svg` title so the label is redundant. For `InlineLoading`, the icon title defaults to the `status` prop for the "error" and "completed" statuses.

**Breaking Changes**

- Change default `Loading` description text from "Active loading indicator" to "loading"
- Remove `id` prop from `Loading` since the label is removed

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -1895,12 +1895,12 @@ None.
 
 ### Props
 
-| Prop name       | Required | Kind             | Reactive | Type                                                                     | Default value          | Description                                                       |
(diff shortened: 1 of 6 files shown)


metonym:

PR #1781, opened 2023-07-23
fix(types): correctly type `kind` prop in `FileUploader`, `FileUploaderButton`

metonym:
The file uploader button is based on `Button`, which now supports more variants like "danger-tertiary" and "danger-ghost."

```diff
- "primary" | "secondary" | "tertiary" | "ghost" | "danger"
+ "primary" | "secondary" | "tertiary" | "ghost" | "danger" | "danger-tertiary" | "danger-ghost"
```

PR #1784, opened 2023-07-23
feat(file-uploader-button): support `danger-tertiary`, `danger-ghost` button variants

metonym:
#1781 fixed the `kind` prop type, but `FileUploaderButton` actually does not use the `Button`. Rather, it's an `input` element, so the classes must be manually applied.


metonym:

PR #1786, opened 2023-07-24
feat(file-uploader): add `size` prop to `FileUploaderButton`

metonym:
Adds the `size` prop to `FileUploaderButton`.

This change is backward-compatible as the default `size` value is `"small"`.


metonym:

PR #1794, opened 2023-08-20
fix(skeleton-text): fix rows reactivity in `paragraph` variant

### LINKED ISSUE #1793: SkeletonText lines carry over from other instances in some cases
https://github.com/carbon-design-system/carbon-components-svelte/blob/a1c7d9bd7c0de5b7321f745b8b85dee5682728aa/src/SkeletonText/SkeletonText.svelte#L16-L26

In svelte lists e.g.
```svelte
{#each list as item}
  <SkeletonText paragraph lines={2} />
{/each}
 ```
the number of lines can carry over from the previous instance in the list. That is, in this example, the first instance will have 2 elements in `rows`, the second will have 4 and so on. 

The reason is that `rows` is a reactive declaration, which will share its value with other members of the list. This behavior is a quirk of Svelte and not necessarily intuitive, but the number of lines is therefore not necessarily determined by the lines prop.  I recommend inserting a

`rows=[]`

just after line 19. 

Or possibly changing `$:rows = []` to `let rows = []` depending on expected use cases.

### PR DESCRIPTION
**Pull request by metonym**
Fixes #1793

This refactors `SkeletonText` to use a more declarative approach: the `Array.from` method in the markup instead of a for loop inside of a reactive statement.

### DIFF
--- a/src/SkeletonText/SkeletonText.svelte
+++ b/src/SkeletonText/SkeletonText.svelte
@@ -13,24 +13,21 @@
 
   const RANDOM = [0.973, 0.153, 0.567];
 
-  $: rows = [];
   $: widthNum = parseInt(width, 10);
   $: widthPx = width.includes("px");
-  $: if (paragraph) {
-    for (let i = 0; i < lines; i++) {
-      const min = widthPx ? widthNum - 75 : 0;
-      const max = widthPx ? widthNum : 75;
-      const rand = Math.floor(RANDOM[i % 3] * (max - min + 1)) + min + "px";
-      rows = [...rows, { width: widthPx ? rand : `calc(${width} - ${rand})` }];
-    }
-  }
 </script>
 
 <!-- svelte-ignore a11y-mouse-events-have-key-events -->
 {#if paragraph}
   <!-- svelte-ignore a11y-no-static-element-interactions -->
   <div {...$$restProps} on:click on:mouseover on:mouseenter on:mouseleave>
-    {#each rows as { width }}
+    {#each Array.from({ length: lines }).map((_, i) => {
+      const min = widthPx ? widthNum - 75 : 0;
+      const max = widthPx ? widthNum : 75;
+      const rand = Math.floor(RANDOM[i % 3] * (max - min + 1)) + min + "px";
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by theetrain — APPROVED** [review]
This works well, though I couldn't reproduce the issue reported in #1793.

I likely misinterpreted it. Here's my attempt: https://svelte.dev/repl/078ab3788d514d3fbf9724d933e39b9b?version=4.2.0

Side note, the scripts `build:docs` and `build:css` won't work until we switch to ESM; or were you able to run them and publish recent packages some other way?

**Comment by metonym** [comment]
> Side note, the scripts build:docs and build:css won't work until we switch to ESM; or were you able to run them and publish recent packages some other way?

Interesting – I'm still able to run these commands. Curious as to why they require ESM?

**Comment by theetrain** [comment]
Not sure either, here's my output

```
➜  carbon-components-svelte-legacy git:(fix-skeleton-text) yarn build:docs
yarn run v1.22.19
warning ../../../package.json: No license field
$ node scripts/build-docs
/.../carbon-components-svelte-legacy/scripts/build-docs.js:3
const { sveld } = require("sveld");
                  ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /.../sveld/lib/index.js from /.../carbon-components-svelte-legacy/scripts/build-docs.js not supported.
Instead change the require of index.js in /.../carbon-components-svelte-legacy/scripts/build-docs.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/.../carbon-components-svelte-legacy/scripts/build-docs.js:3:19) {
  code: 'ERR_REQUIRE_ESM'
}

Node.js v18.16.0
```

In package.json, `"type": "module"` or `"commonjs"` don't seem to influence my output.

metonym:
I'm using:

- Yarn 1.22.19
- Node 18.16.0

Is it possible that you were using a symlinked version of `sveld` when working on https://github.com/carbon-design-system/sveld/pull/108?

metonym:
I was also no able to reproduce it in later version of Svelte. However, I feel that refactoring to avoid the for loop inside the reactive statement is worth it.

Issue #1629, opened 2023-01-16
Carbon v11 style integration


theetrain:
This is a placeholder ticket to track the integration of Carbon v11 styles into carbon-components-svelte. Adopting Carbon v11 styles is our highest priority, but shipping components with `<style lang="sass">` is currently problematic with Vite as discussed in #1561.

I'm currently evaluating `@sveltejs/package` in the [`carbon-v11`](https://github.com/carbon-design-system/carbon-components-svelte/tree/feat/carbon-v11) branch in order to accomodate these aspirational goals:

- Write components using `<script lang="ts">` to distribute as JS Svelte components and companion `*.svelte.d.ts` files for TypeScript users
- Import Carbon v11 SCSS directly via `<style lang="sass">@use ...</style>` and rely on `@sveltejs/package` to compile styles into scoped CSS
- Ensure there's a way to theme components via CSS variables once SCSS compiles to scoped CSS in each component
- Test build performance on a plain SvelteKit app

If the above goals turn out to be favourable, we'll proceed to implement Carbon v11 styles into every component. Special thanks to @gregorw for early exploration in #1496.

theetrain:
@metonym some good progress so far (see notes above). 

Here's a TypeScript follow up to https://github.com/carbon-design-system/carbon-components-svelte/pull/1628#issuecomment-1383310189

```ts
import type { HTMLLabelAttributes } from "svelte/elements";

export let labelAttributes: HTMLLabelAttributes = {};
```

This will produce HTML types for the respective element. This will probably improve DX when adding props to components because you'll see the Carbon props as well as `labelAttributes` instead of carbon props mixed with every possible HTML attribute in alphabetical order. 

We can add `lang="ts"` gradually as we refactor components since JSDoc compiles fine as is. One thing I'd like to change is place `import` statements before `export` statements to follow ESM conventions.

Try this locally:
- checkout the `feat/carbon-v11` branch
- run `yarn` then `yarn build:package`
- Observe output in `package/TextInput/TextInput.svelte.d.ts`
- Observe built css in `package/Button/Button.svelte`

Once I'm finished evaluating and setting up the new project structure with `svelte-package`, I think we can keep `feat/carbon-v11` as a long-running branch and adopt v11 styles one at a time, making pull requests targeting `feat/carbon-v11`. I can add a task list above to see who's working on which components. 

Let me know if you have any feedback.

metonym:
@theetrain proof-of-concept LGTM

[...]

PR #1807, opened 2023-09-03
Breaking: Move `$$restProps` in `Select` from wrapper `<div>` to `<select>`

brunnerh:
This aligns the component with others like `TextInput`, `NumberInput` or `TextArea` and allows adding not explicitly defined attributes to the "main" element of the component. Unfortunately there is no real equivalent for `DropDown`, which is non-standard.

Maybe there could be a standardized property that exists on all inputs (like `ref`) for adding attributes on the root wrapper via spread, the main use case for me would be to apply layout classes to e.g. add margins.

Other changes are from formatting by `lint` task.
(Prettier removes manual Markdown line breaks (two trailing spaces) from JSDoc strings; turned that into paragraphs in `Header`.)


theetrain:
This is a good suggestion that is part of https://github.com/carbon-design-system/carbon-components-svelte/issues/1621 "pass-through" props. Maybe after #1629 is completed these props can be overhauled.

metonym:

PR #1810, opened 2023-09-27
feat(data-table): pass `row` to `display` function

Pierstoval:
This will allow `display` usage to use the whole object for use with computed values, or when needing multiple fields to show more than one element (like "item actions" that might need more than one field, for instance).

### DIFF
--- a/src/DataTable/DataTable.svelte
+++ b/src/DataTable/DataTable.svelte
@@ -2,12 +2,12 @@
   /**
    * @typedef {string} DataTableKey
    * @typedef {any} DataTableValue
-   * @typedef {{ key: DataTableKey; empty: boolean; display?: (item: Value) => DataTableValue; sort?: false | ((a: DataTableValue, b: DataTableValue) => (0 | -1 | 1)); columnMenu?: boolean; width?: string; minWidth?: string; }} DataTableEmptyHeader
-   * @typedef {{ key: DataTableKey; value: DataTableValue; display?: (item: Value) => DataTableValue; sort?: false | ((a: DataTableValue, b: DataTableValue) => (0 | -1 | 1)); columnMenu?: boolean; width?: string; minWidth?: string; }} DataTableNonEmptyHeader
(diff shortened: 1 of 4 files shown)


Pierstoval:
> Can you update the JSDoc types in `DataTable.svelte` ([Lines 5, 6, 10](https://github.com/carbon-design-system/carbon-components-svelte/blob/master/src/DataTable/DataTable.svelte#L5-L10)) and run `yarn build:docs`?

According to the docs, only line 10 needs update: the "display" method receives the row object only for cells, not headers, am I right?

Just rebased & pushed an update

metonym:
I believe that the `display` is reflected in the `DataTableNonEmptyHeader` and `DataTableEmptyHeader` interfaces.

When testing the types, I see this:

Screenshot 2023-09-28 at 8 53 59 AM

Repro: try using `row` in the type test for `DataTable`: https://github.com/carbon-design-system/carbon-components-svelte/blob/aae2dd42aa6b9b69fe7864564d23e8a823f6e340/tests/DataTable.test.svelte#L179

Pierstoval:
I don't see a reason to make this change on headers: `row` object should not be accessible when displaying headers 🤔

This means two things:

* The `row` object should be nullable for headers, and non-null for rows, I don't know how to do that
* The `display` function is overriden from the `headers` object, that's probably the cause of these typing issues:
    https://github.com/carbon-design-system/carbon-components-svelte/blob/aae2dd42aa6b9b69fe7864564d23e8a823f6e340/src/DataTable/DataTable.svelte#L200-L204

I think we should change the behavior for that.

I just pushed a fix to add `row?: ...` to header-related `display` functions, and kept `row: ...` for cell-related.

I'm not sure the rest of the PR is okay, feel free to make more comments 👌

src/DataTable/DataTable.svelte
```diff
@@ -200,7 +200,7 @@
     rows[row.id] = headerKeys.map((key, index) => ({
       key,
       value: resolvePath(row, key),
-      display: headers[index].display,
+      display: row.display ? row.display : headers[index].display,
```
metonym, line 203: I would prefer the original behavior, where the `display` is optionally controlled by the header (and this PR now passes the original `row` to `display`).

Pierstoval: Ok, I just realized something: the `display` function isn't actually used in the header part.

Using it for headers isn't really in the scope of this PR, I'll rollback this specific change and keep the original behavior 👌

metonym (approved):

Pierstoval:
Thank you @metonym !

metonym:

PR #1812, opened 2023-09-28
fix(file-uploader-button): clear value by setting to `""` instead of `null`

jperelli:
This fixes the following error I got:

> InvalidStateError: Failed to set the 'value' property on 'HTMLInputElement': This input element accepts a filename, which may only be programmatically set to the empty string

which made my tests in vitest fail.


metonym (approved):
LGTM, thanks!

metonym:

PR #1818, opened 2023-10-03
fix(tile-group): add `name` and `required` props

metonym:
Similar to #1037, fixes `TileGroup` to forward `name` / `required` props to its group of inputs.

### DIFF
--- a/src/Tile/RadioTile.svelte
+++ b/src/Tile/RadioTile.svelte
@@ -8,6 +8,9 @@
   /** Set to `true` to disable the tile */
   export let disabled = false;
 
+  /** Set to `true` to mark the field as required */
+  export let required = false;
+
   /** Specify the value of the radio input */
   export let value = "";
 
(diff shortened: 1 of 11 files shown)


src/Tile/RadioTile.svelte
```diff
@@ -23,10 +26,16 @@
   /** Specify a name attribute for the input */
   export let name = "";
```
brunnerh, line 27: Maybe also switch the default to `undefined` so the attribute is not added without value.

brunnerh:
Would obsolete #973 and address #972.

I removed `name` from the child component, but that is not really necessary.
Maybe then an explicit `required` should be added to `RadioTile` for consistency.

src/Tile/RadioTile.svelte
```diff
@@ -23,10 +26,16 @@
...
 
-  import { getContext } from "svelte";
+  import { getContext, readable } from "svelte";
```
brunnerh, line 29: That's from `'svelte/store'`

metonym:

PR #1819, opened 2023-10-03
fix(radio-button-group): strongly type dispatched change/select events

metonym:
Strongly type the `event.detail` as `string` for `RadioButtonGroup` and `TileGroup`. Currently, it defaults to `any` since the type is not explicitly annotated.

```svelte
<RadioButtonGroup
  legendText="Storage tier (disk)"
  selected="standard"
  on:change="{(e) => {
    console.log(e.detail); // string
  }}"
/>
```

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -2983,13 +2983,13 @@ None.
 
 ### Events
 
-| Event name | Type       | Detail |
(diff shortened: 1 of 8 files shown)


metonym:
Fixed in [v0.81.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.81.0)

Issue #1831, opened 2023-10-22
MultiSelect cannot clear selectedIds set in onMount()


keehun:
When setting an initial `selectedIds` within `onMount()` for a `MultiSelect`, the "clear" button does not work until the MultiSelect is "interacted with" in some way (show the items, type something into the box, etc).

Here is a repro: https://svelte.dev/repl/63e6f961f6b1483fab1eac28da08c011?version=4.2.2

I will submit a PR to fix this.

PR #1832, opened 2023-10-22
fix(multi-select): reset `selectedIds` when clearing selection

keehun:
Here I have a fix for issue #1831. I am unable to reproduce the issue (reproducible with `master` branch [here](https://svelte.dev/repl/63e6f961f6b1483fab1eac28da08c011?version=4.2.2)) after applying this patch.

If there is a better way to handle initial state set from `onMount()`, I would love to implement a better solution. Please let me know.

### DIFF
--- a/src/MultiSelect/MultiSelect.svelte
+++ b/src/MultiSelect/MultiSelect.svelte
@@ -393,6 +393,7 @@
           selectionCount="{checked.length}"
           on:clear
           on:clear="{() => {
+            selectedIds = [];
             sortedItems = sortedItems.map((item) => ({
               ...item,
               checked: false,


metonym (approved):
Thanks for providing a clear repro and description. LGTM

metonym:
Released in [v0.81.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.81.1)

keehun:
I noticed another issue around setting initial state in `onMount()` where the checked items don't show at the top of the list when `MultiSelect` is focused for the first time. I'll create a repro and open another ticket (and hopefully a fix) when I have a bit of time, but wanted to leave that here before I forgot about it. (This is mostly a note for myself.)

PR #1835, opened 2023-10-26
feat(multi-select): customize hidden inputs

### LINKED ISSUE #1742: MultiSelect not submiting payload inside a form
When submiting a form, the multi select values are not been serialised into the payload

image

```
<script>
    import {Button, Form, MultiSelect, TextInput} from "carbon-components-svelte";
    import {enhance} from '$app/forms';
</script>

<Form method="post">
    <TextInput name="text"></TextInput>
    <MultiSelect
            id="items"
            name="items"
            titleText="Contact"
            label="Select contact methods..."
            items={[
    { id: "0", text: "Slack" },
    { id: "1", text: "Email" },
    { id: "2", text: "Fax" },
  ]}
    />
    <Button type="submit">Submit</Button>
</Form>
```

theetrain:
Fixes #1742 

MultiSelect up til now wasn't rendering hidden inputs in order to be submittable within a `<form>`. With this enhancement:

1. ~~MultiSelect renders all items as `<input type="hidden" name="some-id" value="true" />` using `value="true"` for checked items and `value="false"` for unchecked items by default.~~
2. ~~New prop `selectedOnly` only renders hidden inputs for user-selected items.~~
3. ~~New prop `combineValues` renders a single hidden input with comma-separated values, or using a consumer-provided delimiter.~~
4. Prop `itemToInput` is enhanced to perform double duty for user-visible checkboxes as well as the hidden inputs, mainly to help override `name` and `value` attributes among the hidden inputs.
5. Use CSS to show and hide MultiSelect values so that their checkboxes are form-submittable

~~Feature (3) was a personal need in order to submit an array of values within a form, but I felt the default behaviour should be checkbox-like as per point (1) assuming that's what consumers of MultiSelect would expect.~~

See included documentation for more examples.


src/MultiSelect/MultiSelect.svelte
```diff
@@ -22,11 +22,26 @@
   export let itemToString = (item) => item.text || item.id;
 
   /**
-   * Override the item name, title, labelText passed to the checkbox input
-   * @type {(item: MultiSelectItem) => { name?: string; labelText?: any; title?: string; }}
+   * Override the item name, title, labelText, or value passed to the user-selectable checkbox input as well as the hidden inputs.
+   * @type {(item: MultiSelectItem) => { name?: string; labelText?: any; title?:
```
metonym, line 26: Nit: I think JSDoc parses this fine, but in my VSCode IDE the parsed `@type` is formatted strangely.

```suggestion
   * @type {(item: MultiSelectItem) => { name?: string; labelText?: any; title?: string; value?: string }}
```

Screenshot 2023-10-25 at 10 10 12 PM

src/MultiSelect/MultiSelect.svelte
```diff
@@ -22,11 +22,26 @@
...
+   * string; value?: string }}
    */
   export let itemToInput = (item) => {};
 
+  /**
+   * Set to `true` to only render selected options as hidden inputs for form submission.
+   * @type {boolean}
+   */
+  export let selectedOnly = false
```
metonym, line 35: Could you run Prettier to format this file? I think that the trailing semicolons will be applied.

metonym (approved):
LGTM – elegant fix and update to the docs.

metonym:
That approach seems valid.

I'll admit that for one-line style changes, I have a slight preference for using the `style:` directive.

metonym (approved):

PR #1840, opened 2023-11-05
fix(select-item): export `class` and `style` props

### LINKED ISSUE #1839: `SelectItem` properties `style`/`class` missing from types.
The values for `style`/`class` are pulled from `$$restProps` but the types only contain the defined props, leading to errors when using TypeScript.

https://github.com/carbon-design-system/carbon-components-svelte/blob/fee236194bbfc7aa9225fe694aa0c7fc098dd6dc/src/Select/SelectItem.svelte#L35-L43

Properties should probably either be explicit or all `$$restProps` should be spread along with a `/** @restProps {option} */` annotation.

metonym:
Fixes #1839 

`SelectItem` is missing type definitions for the `class` and `style` props. Even though this defines props on the component, this is not a breaking change as the behavior has always been supported.

### DIFF
--- a/src/Select/SelectItem.svelte
+++ b/src/Select/SelectItem.svelte
@@ -14,6 +14,20 @@
   /** Set to `true` to disable the option */
   export let disabled = false;
 
+  let className = undefined;
+
+  /**
+   * Specify the class of the `option` element
+   * @type {string}
+   */
+  export { className as class };
+
+  /**
+   * Specify the style of the `option` element
+   * @type {string}
+   */
+  export let style = undefined;
+
   import { getContext, onMount } from "svelte";
 
   const id = "ccs-" + Math.random().toString(36);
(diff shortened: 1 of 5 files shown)


src/Select/SelectItem.svelte
```diff
@@ -14,6 +14,14 @@
   /** Set to `true` to disable the option */
   export let disabled = false;
 
+  let className = "";
+
+  /** Specify the class of the `option` element */
+  export { className as class };
+
+  /** Specify the style of the `option` element */
+  export let style = "";
```
brunnerh, line 23: Would suggest `undefined` so the attribute is not added unnecessarily.
(Could also be done for `class`, but there already is a Carbon class on the element.

```suggestion
  /**
   * Specify the style of the `option` element
   * @type {string}
   */
  export let style = undefined;
```

metonym: Good catch – amended.

brunnerh (approved):

theetrain (approved):
Neat, I can ship this Monday evening, but if you get to it first then go ahead.

metonym:
Released in [v0.81.3](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.81.3)

PR #1846, opened 2023-11-18
fix(pagination-nav): set `button type="button"` to prevent form submission

metonym:
When `PaginationNav` is used inside a `form`, the form will be submitted when clicking a page button.

This is because the `button` lacks an explicit `type` attribute. If contained within a form, the default type is "submit."

```html
<form>
  <PaginationNav />
</form>
```


metonym:
Fixed in [v0.82.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.82.1)

PR #1849, opened 2023-11-20
docs: implement rudimentary full-text search

metonym:
This PR aims to improve the docs UX by adding client-side, full-text search using [minisearch](https://www.npmjs.com/package/minisearch).

The search index is extremely basic; it only searches the component name and example heading. More weight is given to the heading. A maximum of 10 results are shown.

For example: searching "bat" should yield `DataTable – batch selection`.

**Future Work**

- Example description text should also be indexed.

---

Screenshot 2023-11-20 at 1 19 24 PM

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -1,10 +1,11 @@
 {
   "private": true,
   "scripts": {
-    "dev": "run-p dev:*",
+    "dev": "yarn build:index-docs && run-p dev:*",
     "dev:routify": "cross-env NODE_ENV=development routify run",
     "dev:svite": "vite dev",
     "build": "run-s build:*",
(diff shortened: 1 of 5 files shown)


docs/package.json
```diff
@@ -1,10 +1,11 @@
 {
   "private": true,
   "scripts": {
-    "dev": "run-p dev:*",
+    "dev": "yarn build:index-docs && run-p dev:*",
     "dev:routify": "cross-env NODE_ENV=development routify run",
     "dev:svite": "vite dev",
     "build": "run-s build:*",
+    "build:index-docs": "node scripts/index-docs.js",
```
metonym, line 8: I don't think the generated search index should be checked into source control; as a result, we need to build the index before dev/build modes.

PR #1850, opened 2023-11-20
fix(header-search): vertically center button icons

metonym:
The search and clear button icons in the `UIShell` `HeaderSearch` component are currently misaligned. They should be vertically centered.


metonym:
Fixed in [v0.82.2](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.82.2)

PR #1851, opened 2023-11-20
fix(header-search): "Escape" should clear search query

metonym:
Pressing "Escape" in `HeaderSearch` should reset the search query without deactivating the search bar.


metonym:
Fixed in [v0.82.2](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.82.2)

PR #1853, opened 2023-11-21
fix(header-search): "Escape" should close empty search bar

metonym:
Follow-up to #1851.

"Escape" should reset the query but keep the search bar focused *if* the value is not empty.

For an empty value, "Escape" should blur and deactivate the search input. This is consistent with search inputs in general.


metonym:
Fixed in [v0.82.3](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.82.3)

PR #1855, opened 2023-11-21
fix(header-search): blur input when deactivating

metonym:
Fast-follow to [#1853](https://github.com/carbon-design-system/carbon-components-svelte/issues/1853).

"Escape" should deactivate the search input if no query is present. However, after deactivating, the input retains focus. It should also be blurred.


metonym:
Fixed in [v0.82.4](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.82.4)

Issue #1863, opened 2023-12-14
no exports condition for svelte


hslee2008:
[vite-plugin-svelte] WARNING: The following packages have a svelte field in their package.json but no exports condition for svelte.

carbon-components-svelte@0.82.4

Please see https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md#missing-exports-condition for details.

metonym:
Re-opening. Noticing a bug where types cannot be resolved. Working on a fix.

<img width="903" alt="Screenshot 2023-12-16 at 5 30 15 PM" src="https://github.com/carbon-design-system/carbon-components-svelte/assets/10718366/c5f1eb95-189e-470c-84a3-cd6b39f49550">

metonym:
Type-related errors should be fixed in [v0.82.6](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.82.6)

PR #1864, opened 2023-12-14
chore: add `exports` field to package.json

hslee2008:
#1863

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,11 @@
   "license": "Apache-2.0",
   "description": "Svelte implementation of the Carbon Design System",
   "svelte": "./src/index.js",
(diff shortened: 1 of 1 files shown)


metonym:
Thanks for this. In my testing, subpaths for `css` and `src` need to included as well.

Otherwise, these imports would fail when Vite attempts to prebundle the library:

```ts
import "carbon-components-svelte/css/all.css";
import Button from "carbon-components-svelte/src/Button/Button.svelte";
```

metonym:
Fixed in [v0.82.5](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.82.5)

PR #1868, opened 2023-12-17
fix(radio-button): allow `value` type to be a number

### LINKED ISSUE #1867: RadioButton `value` type should allow numbers
### Discussed in https://github.com/carbon-design-system/carbon-components-svelte/discussions/1866

<div type='discussions-op-text'>

<sup>Originally posted by **hslee2008** December 16, 2023</sup>
Currently, carbon-components-svelte only allows type of `string | undefined` in value (RadioButton).

image

But there are many cases where the value needs to be numbers.

I mean, putting in numbers does work but the types only say `string` is allowed.</div>

metonym:
Fixes #1867 

Currently, the RadioButton `value` is typed to only allow strings. This was inadvertence since the type definitions for the library are auto-generated. If no explicit type is annotated, the type generator will use its default type.

This PR fixes loosens the `value` type to allow strings and numbers. I'd prefer not to make this an `any` type as I believe there is value in strictness.

Without this change, the following code would run fine but produce type errors.

```svelte
<script lang="ts">
  import { RadioButtonGroup, RadioButton } from "carbon-components-svelte";

  const plans = [1, 2, 3];

  let plan = plans[0];

  $: console.log("Selected plan", plan, typeof plan);
</script>

<RadioButtonGroup
  bind:selected={plan}
  on:change={(e) => {
    const selected = e.detail;
    console.log("on:change", selected, typeof selected);
  }}
>
  {#each plans as value (value)}
    <RadioButton {value} />
  {/each}
</RadioButtonGroup>

```

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -2937,7 +2937,7 @@ None.
 | :------------ | :------- | :--------------- | :------- | ----------------------------------------- | ------------------------------------------------ | --------------------------------------------------- |
 | ref           | No       | <code>let</code> | Yes      | <code>null &#124; HTMLInputElement</code> | <code>null</code>                                | Obtain a reference to the input HTML element        |
 | checked       | No       | <code>let</code> | Yes      | <code>boolean</code>                      | <code>false</code>                               | Set to `true` to check the radio button             |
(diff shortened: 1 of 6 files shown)


metonym:
Fixed in [v0.82.7](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.82.7)

PR #1870, opened 2023-12-22
Removing use of reserved symbol $ in components

Sahasrara:
This is in anticipation of the release of Svelte 5

### DIFF
--- a/src/ContentSwitcher/Switch.svelte
+++ b/src/ContentSwitcher/Switch.svelte
@@ -23,8 +23,8 @@
 
   ctx.add({ id, text, selected });
 
-  const unsubscribe = ctx.currentId.subscribe(($) => {
-    selected = $ === id;
+  const unsubscribe = ctx.currentId.subscribe((currentIdVal) => {
+    selected = currentIdVal === id;
   });
 
   afterUpdate(() => {
--- a/src/ProgressIndicator/ProgressIndicator.svelte
+++ b/src/ProgressIndicator/ProgressIndicator.svelte
@@ -16,8 +16,8 @@
 
   const dispatch = createEventDispatcher();
   const steps = writable([]);
-  const stepsById = derived(steps, ($) =>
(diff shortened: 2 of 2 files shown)


src/ContentSwitcher/Switch.svelte
brunnerh, line 1: The code here strikes me as unnecessarily complicated.

The context, or just the store, could be destructured, then an auto-subscription instead of a manual one could be used.

```js
const { currentId, add, update, change } = getContext("ContentSwitcher");
add({ id, text, selected });

$: selected = $currentId === id;
```
```js
const { currentId, ...ctx } = getContext("ContentSwitcher");
ctx.add({ id, text, selected });

$: selected = $currentId === id;
```
(Unless there is a specific timing-related reason to not use a reactive statement here.)

src/ProgressIndicator/ProgressIndicator.svelte
```diff
@@ -16,8 +16,8 @@
 
   const dispatch = createEventDispatcher();
   const steps = writable([]);
-  const stepsById = derived(steps, ($) =>
-    $.reduce((a, c) => ({ ...a, [c.id]: c }), {})
+  const stepsById = derived(steps, (stepStore) =>
+    stepStore.reduce((a, c) => ({ ...a, [c.id]: c }), {})
```
brunnerh, line 20: `steps` is the store, this is the content.

Only `$` on its own is reserved, so in cases like this, I would go with the conventional `$steps` which is analogous to the name in auto-subscriptions.

metonym:
Hey, thanks for this PR.

This came up today regarding interoperability with Svelte 5. I've applied the changes in this commit to #1926, and assigned [commit co-authorship](https://github.com/carbon-design-system/carbon-components-svelte/pull/1926/commits/e5ea940da9643aa309cb5b3ea70b089c7dd1936a) to @Sahasrara.

Issue #461, opened 2020-12-11
UIShell components: rename "isOpen" prop to "open"


weaseldotro:
The `isOpen` prop should be renamed to `open` in the `UIShell` components: `HeaderAction`, `HamburgerMenu`, `SideNav`.

This would be a breaking change for apps currently using these components, but would improve consistency in prop naming with the rest of the library. There are only 3 components using this `isOpen`, while 11 other components use `open`.

PR #1876, opened 2024-01-02
UIShell components: rename "isOpen" prop to "open"

daviseidel:
Solving issue #461, it will bring more consistency, as now all components use `open` prop instead of `isOpen`


theetrain (commented):
Thank you, overall this works and helps with #1621 

When you have a moment, please update parts in `/docs` where UIShell is used with the `open` prop, and generated docs:

- [ ] Run `yarn build:docs` and `yarn build:lib`
- [ ] `/docs/src/pages/framed/UIShell/` examples
- [ ] `docs/src/pages/_layout.svelte`
- [ ] Anywhere else UIShell components are used

daviseidel:
Fixed all instances I could find. Tell me if I missed any.

PR #1878, opened 2024-01-02
fix(button): forward `on:focus` event

### LINKED ISSUE #1830: Button `focus()` enhancements
Currently, `Button.svelte` does not forward focus events. Adding `on:focus` to the element would solve this.

Also, to programmatically assign focus, access to the HTML element is required, but currently the Button component does not allow access. Perhaps this would work:

```
export let el: HTMLElement;

<button bind:this={el}
```

Which would allow `buttonInstance.el.focus()` in parent components.

I'm happy to generate a PR if these approaches are acceptable.

ryanbmarx:
Fixes #1830

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -402,6 +402,7 @@ export type BreakpointValue = 320 | 672 | 1056 | 1312 | 1584;
 | :--------- | :-------- | :----- |
 | click      | forwarded | --     |
 | focus      | forwarded | --     |
+| blur       | forwarded | --     |
 | mouseover  | forwarded | --     |
 | mouseenter | forwarded | --     |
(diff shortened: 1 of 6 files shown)


theetrain:
This sounds good to me. Please also update the documentation by running `yarn build:docs` and check in changed files.

ryanbmarx:
Committed updated docs

brunnerh:
`focus` is often used with `blur`, would add that as well.

ryanbmarx:
> `focus` is often used with `blur`, would add that as well.

Done.

metonym (approved):

metonym:
Thanks! Released in [v0.82.8](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.82.8)

PR #1880, opened 2024-01-03
Update checked for radio buttons without group.

### LINKED ISSUE #1879: RadioButton checked doesn't change
The checked property doesn't change if:

- You don't have a radio group
- You do have a radio group, but no "name" is specified.

brunnerh:
Also make radio button group `name` a required property (can change this again, if there is a good reason not to do this).

Unfortunately the `change` event does only fire for the newly selected radio button, not for the ones that are automatically unselected; this is why I used `document.getElementsByName` to find the associated ones.

Fixes #1879.

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -2967,7 +2967,7 @@ None.
 | selected      | No       | <code>let</code> | Yes      | <code>string &#124; number</code>           | <code>undefined</code>    | Set the selected radio button value                      |
 | disabled      | No       | <code>let</code> | No       | <code>boolean</code>                        | <code>false</code>        | Set to `true` to disable the radio buttons               |
 | required      | No       | <code>let</code> | No       | <code>boolean</code>                        | <code>undefined</code>    | Set to `true` to require the selection of a radio button |
-| name          | No       | <code>let</code> | No       | <code>string</code>                         | <code>undefined</code>    | Specify a name attribute for the radio button inputs     |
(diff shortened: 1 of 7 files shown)


brunnerh:
> Also make radio button group `name` a required property (can change this again, if there is a good reason not to do this).

The skeleton UIs might be a good reason 🤔.

```svelte
<RadioButtonGroup legendText="Storage tier (disk)" name="skeleton">
  <RadioButtonSkeleton />
  <RadioButtonSkeleton />
  <RadioButtonSkeleton />
</RadioButtonGroup>
```

metonym:
@brunnerh Happy to revisit this. In the context of implementing https://github.com/carbon-design-system/carbon-components-svelte/pull/2225, what does this PR do?

brunnerh:
It synchronizes the `checked` state across members of the implicit radio button group.
See [comments on your recent PR](https://github.com/carbon-design-system/carbon-components-svelte/pull/2225#issuecomment-3412080609).

metonym:
Why does it need to use the DOM to do this? Why can't RadioButtonGroup handle this?

brunnerh:
Because there is no `RadioButtonGroup`, this is specifically about usage without that component, e.g. because the radio buttons are far apart with other UI in-between. `RadioButtonGroup` does not just provide context, it also does layout (which arguably maybe it should not do in the first place).

metonym:
Thanks for your patience: thoughts on #2523? Would that work for you?

PR #1881, opened 2024-01-03
Initial pre-release - Carbon v11 styles

theetrain:
Closes #1636, closes #1709, closes #1886   
Relates to: #1872, #1629

This is the initial prerelease for carbon-components-svelte featuring:

- Carbon v11 styles (from `@carbon/styles@1.46.0`)
- borrowed adaptations from #1709 (thank you @gregorw)

## Questions

1. Should we support deprecated property values such as ContentSwitcher's `size="xl"`? I think we shouldn't.
2. Should default sizes be explicit strings such as `size="md"` instead of "size=undefined"? It may be better to use explicit values when writing a form generator, but they can always return `size=null` I suppose. `Search.svelte` supports `size="md"`; maybe this can be addressed as part of #1621

## Before merging todos

- [x] upgrade `@carbon/styles` to latest
- [x] remove DatePicker `size="xl"` option
- [x] test ExpandableTile docs
- [x] Test Toggle and ToggleSkeleton
- [x] Test TreeView docs
- [x] Restore Truncate component
- [x] Regenerate CSS files
   - [x] use `--cds` class prefix from flagship package instead of `--bx`
   - [x] create truncate classes from mixins
   - [x] include CSS in docs
- [x] Update Theme component to use new `data-` attributes
- [x] Ensure docs render nicely using v11 styles

Deferred todos:

- [ ] Components with glaring issues (to be resolved later)
   - [ ] CodeSnippet component
   - [ ] Black 'search' component
   - [ ] Notification CSS
   - [ ] Tile
- [ ] Test in a new local SvelteKit project to ensure modules can link
- [ ] Identify which components are 'v11 ready' and which aren't in #1629

## After merging todos

- [ ] "squash and merge" this PR onto the `next/v11-styles` branch (and subsequent contributions); use 'breaking changes' message below in commit body
- [ ] Rebase `next/carbon-v11` onto `master` to get latest fixes and features
- [ ] with `next/carbon-v11` checked out, deploy package as `1.0.0-alpha.0` in package.json (`v1.0.0-alpha.0` in git)
- [ ] Include install and contribution instructions in release notes; update related issues and discussions; (remember to mention using npm)
- [ ] Consider deploying docs based on changes in the `next/carbon-v11` branch

👀 The following can go into the squashed commit message:

## Breaking changes

Overall, this is a major style change the will impact the appearance and features of many components. Use caution when upgrading and test your applications.

### Components

- Button has new prop values for `size` and `kind`
[...]

PR #1886, opened 2024-01-11
chore: Use @carbon/styles

gregorw:
See upgrade guide here: https://carbondesignsystem.com/migrating/guide/develop

As proposed in https://github.com/carbon-design-system/carbon-components-svelte/pull/1881 we can now start adopting v11 styles and merge incremental changes and fixes onto `next/v11-styles` branch. I would propose to merge small steps onto this new branch and discuss separte topics individually. This allows many contributors to propose small changes.

The v11 style adoption should be held as minimal as possible since many components already have some notable changes in Carbon v11.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "flatpickr": "4.6.9"
   },
   "devDependencies": {
+    "@carbon/styles": "^1.47.0",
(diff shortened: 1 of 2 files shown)


gregorw:
It seems that we can take larger steps. See #1881

Issue #1888, opened 2024-01-13
TextInput v11


theetrain:
Spun off from #1629

- [x] Follow v11 styles for TextInput
- [x] Follow v11 styles for TextInputSkeleton
- [x] #1667
- [x] #1633 
- [x] #941 
- [x] #752 
- [x] #534 (relates to #1621)
- [x] (optional) #545

theetrain:
@SimpleProgrammingAU I captured as many outcomes as I could above. Thank you for contributing!

SimpleProgrammingAU:
@theetrain do you wanna add `PasswordInput` to this issue?

theetrain:
@SimpleProgrammingAU I think `PasswordInput` can be handled separately since it doesn't share co-dependencies with `TextInput`; this way PRs can be kept small.

theetrain:
This was released in [v1.0.0-next.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v1.0.0-next.1).

Issue #1667, opened 2023-02-26
[FluidForm] TextInput error icon is misplaced


theetrain:
When rendering an `invalid` TextInput, its icon is misaligned:

<img width="762" alt="image" src="https://user-images.githubusercontent.com/12798751/221391020-6db35498-6cc6-4d96-a85c-a2396c6a33c3.png">

This may potentially impact other inputs as well: PasswordInput, NumberInput, DateInput, TextArea.

Discovered in #1666

Issue #1633, opened 2023-01-17
[TextInput] helperText enhancements


theetrain:
- [ ] Use `aria-describedby` to read TextInput's `helperText`
- [ ] Provide slot for `helperText` to allow formatted messages

theetrain:
`aria-describedby` fix now available in [v0.73.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.73.1).

theetrain:
This fix will be included once v11 styles are shipped. See #1888.

Issue #941, opened 2021-12-18
Input validation issue on TextInput component


alkanna:
See REPL here : https://svelte.dev/repl/a122d2d5e47345099c0bb4a1712e5130?version=3.44.3

When erasing the input content (one character at a time), it's filled again with the original content

metonym:
Interesting, but I'm unable to repro in a non-REPL setting.

Exporting `value` in Input.svelte seems to fix it though: https://svelte.dev/repl/ff0ba6fcbc0347afb7695e8843d9831b?version=3.44.3

alkanna:
Interesting indeed... I guess we can close this then. I spent half na hour trying to find the issue inside the TextInput component... Guess it was my mistake.
I would be really interested in an explanation though if someone knows why this is happening.

metonym:
I would also be interested in an explanation. It seems that there is some reactivity issue where the `value="test"` prop passed down using `$$restProps` clashes with the `bind:value` in the input component. 

I've created another REPL that produces the same behavior when using a plain `input`: https://svelte.dev/repl/c58a66f954304c538e0ca02120e84b98?version=3.44.3

SimpleProgrammingAU:
Whenever I try any of the REPL links above, it gives me this error:
```
Unexpected character '@' (TreeViewNodeList.svelte:80:3)
```

I tried Firefox and a Chromium browser and both had the same error. Does anyone know what that's about?

PR #1889, opened 2024-01-14
TextInput v11 Styles

SimpleProgrammingAU:
Tasks completed and ready for review

- Adopts v11 styles for `TextInput` and `TextInputSkeleton` (#1888)
- Addresses `FluidForm` issues relating to `TextInput` (#1667)
- Added `slot`s for `helperText` (#1633), `invalidText`, `warnText` (#1621)
- Added `elementAttribute` objects to replace `$$restProps` (#1621) - is `Record<string, string>` the correct type for these?
- Property names checked (#534)

Tasks outstanding

- #752: I think this issue might be better resolved in `Select` - up for discussion
- #941: I couldn't open the REPL links without errors to see the issue
- #545: I haven't tried to tackle this one yet, will try to address it when I look at the other outstanding tasks above

I also alphabetised the prop list for legibility. I figured this was a good point to submit a PR since all v11 style stuff is done, just need to tidy up the loose ends with the other bugs.


brunnerh:
I would recommend doing things that cause a lot of non-functional differences, like sorting properties, in separate PRs. That makes it easier to review as there is less noise.

theetrain:
@brunnerh that's a good tip; I moved this outcome to its own ticket: #1891.

I'll try reviewing this PR as-is for now.

src/TextInput/TextInput.svelte
```diff
@@ -86,6 +114,8 @@
   /** @type {(e: Event) => void} */
   const onInput = (e) => {
     value = parse(e.target.value);
+    if (maxCount && value.length > maxCount) value = value.slice(0, maxCount);
+    if (counter && maxCount) updateCount();
     dispatch("input", value);
```
theetrain, line 119: Since other events were modified, I recommend forwarding `on:input` on the
`<input>` element and removing this dispatched `input`. It's a breaking change,
but one we'll get around to anyway.

src/TextInput/TextInput.svelte
```diff
@@ -94,6 +124,14 @@
     dispatch("change", parse(e.target.value));
```
theetrain, line 124: I think these `onChange` and `parse` methods should be removed. We cast
`number`-type inputs, but the React library doesn't. Maybe this can be handled
in userland, especially since `<input type="number" />` when submitted in a
`<form>` is read server-side as a string.

Demo: https://www.sveltelab.dev/csng8xsyb7bhy56

cc @metonym

src/TextInput/TextInput.svelte
```diff
@@ -125,38 +164,64 @@
           class:bx--label--disabled="{disabled}"
           class:bx--label--inline="{inline}"
           class:bx--label--inline--sm="{size === 'sm'}"
+          class:bx--label--inline--md="{size === 'md'}"
           class:bx--label--inline--lg="{size === 'lg' || size === 'xl'}"
```
theetrain, line 168: Please remove `xl` since it's no longer part of v11 styles.

src/TextInput/TextInput.svelte
```diff
@@ -5,78 +5,106 @@
...
-  /** Specify the invalid state text */
-  export let invalidText = "";
+  /** Set to `true` to mark the field as required */
+  export let required = false;
+
+  /**
+   * Set the size of the input
+   * @type {"sm" | "md" | "lg"}
+   */
+  export let size = undefined;
```
theetrain, line 74:
```suggestion
  export let size = 'md';
```

SimpleProgrammingAU:
> I would recommend doing things that cause a lot of non-functional differences, like sorting properties, in separate PRs. That makes it easier to review as there is less noise.

I thought the same thing after the fact. I sorted the properties because I was having so much difficulty finding them in the code as I was reviewing against the React component.

theetrain (changes requested):
Thanks for helping standardize some props and events while adopting v11 styles. 

Note to self:

### TextInput Breaking changes

- `on:mouseover`, `on:mouseenter`, and `on:mouseleave` events are no longer forwarded. Use `on:pointerup`, `on:pointerover`, `on:pointenter`, and `on:pointerleave` instead
- `size` no longer supports `xl` value

### TextInput new features

- Carbon v11 styles
- New `counter` and `maxCount` props to display a character count
- New slots: `helperText`, `invalidText`, `warnText`
- `light` prop is deprecated. Use the forthcoming `Layer` componenet instead

src/TextInput/TextInput.svelte
```diff
@@ -125,38 +164,64 @@
...
+      <label
+        for="{id}"
+        class:bx--label="{true}"
+        class:bx--visually-hidden="{hideLabel}"
+        class:bx--label--disabled="{disabled}"
+        class:bx--label--inline="{inline}"
+        class:bx--label--inline-sm="{inline && size === 'sm'}"
+        class:bx--label--inline-md="{inline && size === 'md'}"
+        class:bx--label--inline-lg="{inline &&
+          (size === 'lg' || size === 'xl')}"
```
theetrain, line 208: Please remove `xl` condition since it's no longer used.

src/TextInput/TextInput.svelte
```diff
@@ -195,16 +267,20 @@
...
+        bind:value="{value}"
         required="{required}"
         readonly="{readonly}"
         class:bx--text-input="{true}"
         class:bx--text-input--light="{light}"
         class:bx--text-input--invalid="{error}"
         class:bx--text-input--warning="{warn}"
         class:bx--text-input--sm="{size === 'sm'}"
+        class:bx--text-input--md="{size === 'md'}"
         class:bx--text-input--lg="{size === 'lg' || size === 'xl'}"
```
theetrain, line 279: Please remove `xl` condition

src/TextInput/TextInput.svelte
```diff
@@ -195,16 +267,20 @@
...
-        {...$$restProps}
+        class:bx--layout--size-sm="{size === 'sm'}"
+        class:bx--layout--size-md="{size === 'md'}"
+        class:bx--layout--size-lg="{size === 'lg' || size === 'xl'}"
```
theetrain, line 282: Please remove `xl` condition

theetrain:
Sorry I submitted my review prematurely. I was using VSCode to review and didn't realize 'request changes' would immediately submit. I'll follow up with the remainder of my review.

image

src/TextInput/TextInputSkeleton.svelte
```diff
@@ -1,17 +1,24 @@
 <script>
   /** Set to `true` to hide the label text */
   export let hideLabel = false;
+
+  /**
+   * Specify the div HTML attributes for the skeleton container
+   * @type {Record<string, string>}
```
theetrain, line 7:
```suggestion
   * @type {import('svelte/elements').HTMLDivAttributes}
```

docs/src/pages/components/TextInput.svx
```diff
@@ -9,47 +9,51 @@ components: ["TextInput", "TextInputSkeleton"]
...
+<TextInput size="lg" labelText="Username" placeholder="Enter username..." />
 
 ## Small size
 
-<TextInput size="sm" labelText="User name" placeholder="Enter user name..." />
+<TextInput size="sm" labelText="Username" placeholder="Enter username..." />
+
+## Character Counter
+
+<TextInput labelText="Username" placeholder="Enter username..." maxCount={32} counter="char" />
```
theetrain, line 45: Please adjust this if you change the `counter` prop.

theetrain (commented):
@SimpleProgrammingAU review done; apologies for the repeat reviews - VSCode's review UI is kinda unintuitive.

Last thing to note:

- [ ] As discussed in #1891, please do not reorder the prop names since they're grouped together in a logical fashion. 

Over to you!

src/TextInput/TextInput.svelte
```diff
@@ -5,78 +5,106 @@
...
-  export let size = undefined;
-
-  /**
-   * Specify the input value.
-   *
-   * `value` will be set to `null` if type="number"
-   * and the value is empty.
-   * @type {null | number | string}
+   * Set to "char" to enable display the character counter or "word" to display the word count.
+   * @type {"char" | "word"}
```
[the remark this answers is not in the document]
SimpleProgrammingAU: I was very confused for a second. I got it from `TextArea` which makes way more sense. I'll remove.

src/TextInput/TextInput.svelte
SimpleProgrammingAU: I'd agree that it would align better with the user expected type especially since there is a separate `NumberInput` component.

Lemme know if you want me to press ahead with that change.

SimpleProgrammingAU:
Thanks for the feedback. Back to you :)

src/TextInput/TextInput.svelte
```diff
@@ -1,50 +1,51 @@
 <script>
   /**
-   * @event {null | number | string} change
-   * @event {null | number | string} input
```
gregorw, line 4: Such changes will certainly break existing codebases. 

I think we should adopt v11 styles primarily while supporting existing codebases as much as possible.

Once v11 styles are merged to master we can then work towards feature parity with the canonical implementation (React). 

V11 styles will be even more delayed if we try to achieve multiple goals at the same time.

SimpleProgrammingAU: This is something that will happen eventually anyway. I'd make the case that given we're already making breaking changes, it may as well all come at once.

Most of the fixes to codebases will be relatively trivial (i.e. largely find & replace and copy-paste).

gregorw: I’m glad you’re pushing this project forward. We want the same thing. My worry is that once we open a PR from [next/v11-styles → master](https://github.com/carbon-design-system/carbon-components-svelte/compare/master...next/v11-styles) the reviewer, i.e. @metonym, will have a very hard time assessing all these to some extend unnecessary changes. What I’m saying is that we will have v11 styles on the `master` branch earlier if we keep the change to a minimum.

> This is something that will happen eventually anyway.

We could propose this very change to `master` then. It’s unrelated to v11 styles. Feature parity with the canonical implementation is a separate issue.

> I'd make the case that given we're already making breaking changes, it may as well all come at once.

1. I don’t think we should make breaking changes. At least I tried [not to make](https://github.com/carbon-design-system/carbon-components-svelte/pull/1709/files#diff-49e1f6991bcb2a3f21420920a5e8e3c3b9760c9d316c50d5390779ee018ddb69R16) any unnecessary breaking changes.
2. I don’t think making them all at once is a good idea because it makes reviewing and writing release notes very hard.

> Most of the fixes to codebases will be relatively trivial (i.e. largely find & replace and copy-paste).

Are you running a `carbon-components-svelte` app in production? We do. If we suddenly miss events on `TextInput` we cannot just run find & replace. It’ll break the application.

theetrain: I appreciate the discussion. I hope I have addressed Gregor's concerns in https://github.com/carbon-design-system/carbon-components-svelte/discussions/1872#discussioncomment-8185572 - let's continue there.

Issue #1893, opened 2024-01-15
Use `Button` component instead of native `button` in `HeaderGlobalAction`


spburtsev:
Currently, adding tooltips to the HeaderGlobalAction component is a bit cumbersome due to its reliance on the native button element. To improve this functionality, I propose using the Button component instead, as it natively supports tooltips and aligns with the approach taken in @carbon/react.

metonym:
Released in [v0.84.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.84.0).

[Example](https://svelte.carbondesignsystem.com/components/UIShell#header-with-utilities) in docs.

As simple as adding an `iconDescription` to complement the icon.

```svelte
<HeaderGlobalAction
  iconDescription="Settings"
  icon={SettingsAdjust}
/>

```

Adjust the tooltip alignment using the `tooltipAlignment` prop. Default value is `"center"`. Other values can be `"start"` or `"end"`.

```svelte
<HeaderGlobalAction
  iconDescription="Settings"
  icon={SettingsAdjust}
  tooltipAlignment="start"
/>

```

PR #1894, opened 2024-01-15
feat(ui-shell): support button tooltip in `HeaderGlobalAction`

spburtsev:
Breaking change : Use Button component in HeaderGlobalAction

** Issue: [#1893](https://github.com/carbon-design-system/carbon-components-svelte/issues/1893)

** Edit: an example of updated HeaderGlobalAction

image

### DIFF
--- a/src/UIShell/HeaderGlobalAction.svelte
+++ b/src/UIShell/HeaderGlobalAction.svelte
@@ -1,4 +1,8 @@
 <script>
+  /**
+   * @extends {"../Button/Button.svelte"} ButtonProps
+   */
+
   /** Set to `true` to use the active variant */
   export let isActive = false;
 
@@ -10,17 +14,18 @@
 
   /** Obtain a reference to the HTML button element */
   export let ref = null;
+
+  import Button from "../Button/Button.svelte";
+
+  $: buttonClass = [
+    "bx--header__action",
+    isActive && " bx--header__action--active",
+    $$restProps.class,
+  ]
+    .filter(Boolean)
+    .join(" ");
 </script>
 
-<button
-  type="button"
-  bind:this="{ref}"
-  class:bx--header__action="{true}"
-  class:bx--header__action--active="{isActive}"
-  {...$$restProps}
-  on:click
->
-  <slot>
-    <svelte:component this="{icon}" size="{20}" />
-  </slot>
-</button>
+<Button bind:this="{ref}" class="{buttonClass}" {...$$restProps} on:click>
+  <svelte:component this="{icon}" slot="icon" size="{20}" />
+</Button>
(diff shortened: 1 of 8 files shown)


types/UIShell/HeaderGlobalAction.svelte.d.ts
```diff
@@ -18,13 +18,11 @@ export interface HeaderGlobalActionProps extends ButtonProps {
    * Obtain a reference to the HTML button element
    * @default null
    */
-  ref?: null | HTMLButtonElement;
-
-  [key: `data-${string}`]: any;
+  ref?: undefined;
```
spburtsev, line 21: I am not sure why it was updated this way by `yarn build:docs`. Should I edit this file manually?

src/UIShell/HeaderGlobalAction.svelte
```diff
@@ -10,17 +14,18 @@
...
-  class:bx--header__action="{true}"
-  class:bx--header__action--active="{isActive}"
-  {...$$restProps}
-  on:click
->
-  <slot>
-    <svelte:component this="{icon}" size="{20}" />
-  </slot>
-</button>
+<Button bind:this="{ref}" class="{buttonClass}" {...$$restProps} on:click>
```
brunnerh, line 29: The order of the properties matters and probably causes issues.

`class:` directives work independently from a `class` property passed in via `$$restProps`.
But now that a component is used, I suspect that when a `class` is set, the spread will overwrite what was set before to `buttonClass`.

If that is the case, the spread should happen before setting class or the component should define a `class` property, so it no longer is part of `$$restProps`.

types/UIShell/HeaderGlobalAction.svelte.d.ts
spburtsev: I applied this suggestion, but the .d.ts file did not update properly until I specified the type in JSDoc explicitly

src/UIShell/HeaderGlobalAction.svelte
spburtsev: I did not find any differences depending on the order of these props, but just in case I put $$restProps before class.

metonym (approved):
I appreciate the both of you for working on this! Tried it myself and it looks great.

I made a couple of changes:

- 6ef0f0640c8cb4fcd944f2ccd707874a0e0709f9: Also apply the refactored icon slot logic to the link button variant (I noticed locally that the icon failed to render for link buttons)
- 321813f828dc25b65515ec0ce984f00d75c68452: Tweak the Header Utilities example to include three `HeaderGlobalAction` components to demo and test out the various tooltip positions

I'll get this out in the next minor version release.

metonym:
Released in [v0.84.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.84.0). Thanks again to both of you – really great to see.

[Example](https://svelte.carbondesignsystem.com/components/UIShell#header-with-utilities) in docs.

PR #1709, opened 2023-03-26
chore: use @carbon/styles (v11)

gregorw:
Reworking #1496. See https://carbon-components-svelte-v11.onrender.com for actual results.

This PR introduces v11 Carbon styles. As a first step, I would make sure that all existing components still work, before introducing new components or changing the technology stack in any other ways.

- [x] @carbon/styles
- [x] @ibm/plex

Here’s the list of components that need to be verified:

- [x] Accordion
- [x] AspectRatio
- [x] Breadcrumb
- [x] Breakpoint
- [x] Button
- [x] ButtonSet
- [x] Checkbox
- [x] ClickableTile
  - [ ] Light variant is not working yet. Uses layer tokens.
- [ ] CodeSnippet
  - [ ] Copy button is misaligned
- [x] ComboBox
- [x] ComposedModal
- [x] ContentSwitcher
- [ ] ContextMenu
- [x] CopyButton
  - [ ] Tooltip is misaligned
- [x] DataTable
  - [ ] `stickyHeader` has a bug in @carbon/styles most probably,  `rem(300px)` is not recognized by browser.
- [x] DatePicker
- [x] Dropdown
- [x] ExpandableTile
- [x] FileUploader
- [x] FluidForm
- [x] Form
  - [ ] Docs need `Stack` component for vertical spacing
- [x] Grid
  - [ ] Needs renaming to `FlexGrid` at some point to clear `Grid` for css grids.
- [x] ImageLoader
- [x] InlineLoading
- [x] InlineNotification
  - [ ] Docs need `Stack` component for vertical spacing
- [x] Link
- [x] Loading
- [x] LocalStorage
  - [ ] Dark mode toggle needs fixing. See `Theme`/`GlobalTheme`
- [x] Modal
- [x] MultiSelect
- [x] NumberInput
- [x] OrderedList
- [x] OverflowMenu
- [x] Pagination
- [x] PaginationNav
- [x] PasswordInput
- [x] Popover
- [x] ProgressBar
- [x] ProgressIndicator
- [x] RadioButton
- [x] RadioTile
- [x] RecursiveList
- [x] Search
- [x] Select
- [x] SelectableTile
- [x] SkeletonPlaceholder
- [x] SkeletonText
- [x] Slider
- [x] StructuredList
  - [ ] Selectable version needs double-checking
- [x] Tabs
- [x] Tag
  - [ ] Interactive needs fixing
- [x] TextArea
- [x] TextInput
- [ ] Theme
- [x] Tile
[...]

PR #1895, opened 2024-01-16
feat!: button v11

theetrain:
Closes: #1641  
For #1629

Todo

- [ ] Test with tooltip/popover, make appropriate changes
- [ ] Implement ButtonSet
- [ ] Implement ButtonSkeleton

Button changes

##BREAKING CHANGES

- `skeleton` prop has been removed. Use `ButtonSkeleton` component instead
- `size` prop has new values
- `kind` prop has new values
- `isSelected` is now `selected
- `on:mouseover`, `on:mouseenter`, and `on:mouseleave` forwarded events were replaced with `on:pointerover`, `on:pointerenter`, and `on:pointerleave`
- `as` no longer accepts boolean types

## Features

- New `2xl` size
- `expressive` styles now apply to all button sizes
- `as` accepts a string to specify a desired HTML element

## Fixes

- Exports all SCSS styles using official conventions
- All theme tokens are included in built CSS files


theetrain:
I'm accepting preliminary reviewed until the tasks above are complete.

gregorw (commented):
This is very hard to review due to the many changes.

Changes to `Button` using v11 can be [much smaller](https://github.com/carbon-design-system/carbon-components-svelte/pull/1709/files#diff-57a51bf4e52511ac70e0d9fe785b1a80b2d7bc115f6b180ea167b275bdf5e6e1) and they still [look good](https://carbon-components-svelte-v11.onrender.com/components/Button).

My 2 cents would be to separate v11 styles from feature parity. Otherwise it will be very hard to review the changes in `next/v11-styles` compared to `master`. We’re taking two steps at the same time and it will make reviewing, writing release notes and upgrading existing apps very very hard.

See also https://github.com/carbon-design-system/carbon-components-svelte/pull/1889#discussion_r1457998510.

If we really want to move the needle with v11 styles we should work on components such as `Theme`, `Layer`, `CodeSnippet`, maybe `Stack`, etc.

src/Button/Button.svelte
```diff
@@ -77,8 +69,14 @@
   /** Obtain a reference to the HTML element */
   export let ref = null;
 
+  /**
+   * Button, anchor, or div attributes
+   * @type {import('svelte/elements').HTMLAnchorAttributes |
+   * import('svelte/elements').HTMLButtonAttributes | import('svelte/elements').HTMLAttributes}
+   */
+  export let buttonAttributes = {};
```
brunnerh, line 77: I am not a fan of this, at least not in this form/for this component.
- It's unwieldy, you repeat "button" which is already part of the element, maybe it could just be `attributes` if it refers to the root element
- `Button` primarily is just a wrapper around a single element, I would prefer to pass any unknown properties automatically via `$$restProps`
- If an attributes property is used, there should at least also be `iconAttributes`.
- It is inconsistent, some attributes, like `disabled`, have their own property.

src/Button/Button.svelte
```diff
@@ -117,73 +106,79 @@
...
-    on:blur
-    on:mouseover
-    on:mouseenter
-    on:mouseleave
+<!-- svelte-ignore a11y-no-static-element-interactions -->
+<svelte:element
+  this="{as || (href && 'a') || 'button'}"
+  bind:this="{ref}"
+  type="{href && !disabled ? undefined : type}"
+  tab-index="{tabindex}"
```
brunnerh, line 121: That `-` should not be there.

src/Button/Button.svelte
theetrain: These are all good points. I think it's fair to revise #1621 accordingly.

> maybe it could just be `attributes` if it refers to the root element

I think `attributes` works. The main benefit of a dedicated prop over `$$restProps` is for more ergonomic autocomplete. Only component props would appear in the IDE, and element attributes would be contained within a dedicated prop such as `attributes`.

> I would prefer to pass any unknown properties automatically via `$$restProps`

Given the autocomplete trade-off above, I'm leaning towards `attributes`, but I'm also torn because this:

```html
<Button kind="secondary" type="submit" data-test-id="abc123">Save</button>
```

Looks nicer than this:

```html
<Button kind="secondary" type="submit" attributes={{ 'data-test-id': "abc123" }}>
  Save
</button>
```

> there should at least also be `iconAttributes`.

Agreed.

We have this note in our standardization ticket:

>HTML attributes are provided via Type declarations and should not be included as explicit props unless it has special treatment as part of a component.
>
>Appropriate: `disabled` for `<Button>`
>Inappropriate: `aria-label` for `<TextInput>`

I suppose `disabled` has special treatment depending on usage of `as` or `href`. For that reason, I think it deserves to be a prop.

Issue #1643, opened 2023-01-29
Slider on:input event


AZProductions:
Could there be a `on:input` event implemented in the Slider component. I'm running into problems where on:change gets fired when the user moves the slider, and when the slider value changes. I only want the event to fire when the user moves the slider.

I recreated a demo here:
https://svelte.dev/repl/10aa153589364fddbccb7ec765bde60e?version=3.55.1

Hope it's clear enough, thank you!

theetrain:
Thanks for pointing this out. Slider, like a few other components, are dispatching `change` reactively when it shouldn't. The problematic code is here:

https://github.com/carbon-design-system/carbon-components-svelte/blob/57b6ea68b53368240b07faff80a8a8157fae0786/src/Slider/Slider.svelte#L119-L121

We're looking to standardize events (see #1621) and it probably makes sense to forward the native `change` event if one exists, rather than dispatch it manually. Let us know if you're willing to contribute a fix, otherwise we'll have this ready for v1.

metonym:
Fixed in [v0.82.10](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.82.10)

PR #1906, opened 2024-01-27
fix(slider): dispatch `on:input` event

mmamedel:
Fixes [#1643](https://github.com/carbon-design-system/carbon-components-svelte/issues/1643)


metonym:
Thanks for the fix. Released in [v0.82.10](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.82.10)

PR #1907, opened 2024-01-27
fix(types): improve `e.detail` type for dispatched events

metonym:
Some dispatched events do not have an explicit type annotation, and thus default to `CustomEvent<any>`.

This PR enhances the DX by explicitly annotating the type.

```svelte
<Slider
  on:change="{(e) => {
    console.log(e.detail); // number
  }}"
/>
```

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -2151,9 +2151,9 @@ None.
 
 ### Events
 
-| Event name | Type       | Detail |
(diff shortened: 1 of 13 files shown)


theetrain (approved):
Looks good, and works well!

metonym:
Fixed in [v0.82.9](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.82.9)

PR #1911, opened 2024-02-07
fix(context-menu): prevent `on:contextmenu` default behavior only when opening menu

### LINKED ISSUE #1909: ContextMenu prevents context menu invocations outside of the target
The use of preventDefault on this line https://github.com/carbon-design-system/carbon-components-svelte/blob/65904575743ba06344fb75e14685e42494c13cde/src/ContextMenu/ContextMenu.svelte#L141

applies to contextmenu events anywhere on the page, so even if you right click somewhere on the page that's not inside the target, it'll prevent the browser's context menu from opening

metonym:
Fixes #1909

Right clicking a non-`target` should allow the default `contextmenu` event.

### DIFF
--- a/src/ContextMenu/ContextMenu.svelte
+++ b/src/ContextMenu/ContextMenu.svelte
@@ -59,6 +59,7 @@
 
   /** @type {(e: MouseEvent) => void} */
   function openMenu(e) {
+    e.preventDefault();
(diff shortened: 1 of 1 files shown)


metonym:
Fixed in [v0.82.9](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.82.9)

PR #1913, opened 2024-02-16
fix(overflow-menu): use `offsetWidth`, `offsetHeight` to compute menu dimensions

Sahasrara:


metonym:
Could you provide an example or some context as to what this solves?

Sahasrara:
> Could you provide an example or some context as to what this solves?

Oh sorry, so specifically I was using these dropdowns nested inside an element with scale transformation applied to it. The client rect calculations actually takes the scale into account so if you use those values to position the dropdown menu, the menu will appear in the wrong position in that context.

Here's a much more specific example:
https://svelteflow.dev/examples

Imagine a dropdown on one of the nodes in the graph on this page. The position of the dropdown will be incorrect if you zoom in or out since they use transform scale.

metonym (approved):
Seems like a reasonable approach.

metonym:

PR #1915, opened 2024-02-19
fix(toast-notification): clear `autoclose` timeout correctly 

### LINKED ISSUE #1914: `ToastNotification` close event can fire after close due to timeout
`ToastNotification` has a built-in timeout mechanism, which automatically closes the toast and fires the close event.

However, if the toast is dismissed manually in that timeframe, the timeout is not stopped and there will be two close events. First the one from the click, then the one from the timeout.

Example:
```js
const t = new ToastNotification({
	target: document.body,
	props: { timeout: 2000, title: 'Close this before timeout' },
});
t.$on('close', e => { console.log(e.detail); });
```
[REPL](https://svelte.dev/repl/bad1207d3b634955be52b911d6b5dca9?version=4.2.11)

The timeout should probably be cleared as soon as the toast is closed.
Also, the it analogously should be started on open, not just `onMount`.

metonym:
Does several things:

**Docs**
- Adds an example using the autoclose `timeout` prop.

**Fixes #1914**

- Fixes `ToastNotification` to start the timer in a reactive statement instead of `onMount`. This means if the `timeout` prop is set or changed after component initialization, the timeout should work as expected.
- Fixes `ToastNotifcation` to reset the timer if the close button is clicked.

---

https://github.com/carbon-design-system/carbon-components-svelte/assets/10718366/eacea953-d93d-4018-bd9f-796b547839c0

### DIFF
--- a/src/Notification/ToastNotification.svelte
+++ b/src/Notification/ToastNotification.svelte
@@ -55,6 +55,9 @@
   let timeoutId = undefined;
 
   function close(closeFromTimeout) {
+    // Clear the timer if the close button was clicked.
+    clearTimeout(timeoutId);
+
     const shouldContinue = dispatch(
       "close",
       { timeout: closeFromTimeout === true },
@@ -66,14 +69,24 @@
   }
 
   onMount(() => {
-    if (timeout) {
-      timeoutId = setTimeout(() => close(true), timeout);
-    }
-
     return () => {
       clearTimeout(timeoutId);
     };
   });
+
+  $: if (typeof window !== "undefined") {
+    /**
+     * Clear the timer if {@link timeout} changes.
+     * If set to `0`, no new timeout is started.
+     * Else, a new timeout is started if {@link open} is not set to `false`.
+     */
+    clearTimeout(timeoutId);
+
+    /** Only start the timer of {@link open} has not been set to `false`. */
+    if (open && timeout) {
+      timeoutId = setTimeout(() => close(true), timeout);
+    }
+  }
 </script>
 
 <!-- svelte-ignore a11y-mouse-events-have-key-events -->
--- a/docs/src/pages/components/ToastNotification.svx
+++ b/docs/src/pages/components/ToastNotification.svx
@@ -14,6 +14,14 @@ See also: [InlineNotification](InlineNotification)
 
 <ToastNotification title="Error" subtitle="An internal server error occurred." caption="{new Date().toLocaleString()}" />
 
+## Autoclose
+
+By default, `ToastNotification` does not automatically close.
+
+Specify the `timeout` prop to automatically close the notification after a specified duration (in milliseconds).
+
+<FileSource src="/framed/ToastNotification/ToastNotificationTimeout" />
+
 ## Prevent default close behavior
 
 `ToastNotification` is a controlled component. Prevent the default close behavior using the `e.preventDefault()` method in the dispatched `on:close` event.
--- a/docs/src/pages/framed/ToastNotification/ToastNotificationTimeout.svelte
+++ b/docs/src/pages/framed/ToastNotification/ToastNotificationTimeout.svelte
@@ -1,0 +1,32 @@
+<script>
+  import { Button, ToastNotification } from "carbon-components-svelte";
+  import { fade } from "svelte/transition";
+
+  let timeout = undefined;
+  $: showNotification = timeout !== undefined;
+</script>
+
+<Button
+  disabled="{showNotification}"
+  on:click="{() => {
+    timeout = 3_000; // 3 seconds
+  }}"
+>
+  Show notification
+</Button>
+
+{#if showNotification}
+  <div transition:fade>
+    <ToastNotification
+      timeout="{timeout}"
+      kind="success"
+      title="Success"
+      subtitle="This notification will autoclose in {timeout.toLocaleString()} ms."
+      caption="{new Date().toLocaleString()}"
+      on:close="{(e) => {
+        timeout = undefined;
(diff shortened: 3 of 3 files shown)


src/Notification/ToastNotification.svelte
```diff
@@ -55,6 +55,11 @@
   let timeoutId = undefined;
 
   function close(closeFromTimeout) {
+    if (closeFromTimeout !== true && timeoutId) {
+      // If the close button is clicked, clear the timeout.
+      clearTimeout(timeoutId);
+    }
```
brunnerh, line 61: The `if` wrapper seems a bit redundant. `clearTimeout` does nothing on expired timeout handler or `undefined`.

src/Notification/ToastNotification.svelte
```diff
@@ -66,14 +71,15 @@
...
-
     return () => {
       clearTimeout(timeoutId);
     };
   });
+
+  $: if (timeout) {
+    clearTimeout(timeoutId);
+    timeoutId = setTimeout(() => close(true), timeout);
+  }
```
brunnerh, line 82: - This would also restart the timeout while the toast is closed and then potentially re-fire the event.
- Does not clear an existing timeout when set to `0` because the `if` would be `false`. Maybe just remove the condition.
- Probably requires something like a `mounted` flag guard (set in `onMount`) to prevent running in SSR.

brunnerh (changes requested):

src/Notification/ToastNotification.svelte
metonym: - Could you give an example?
- The default `timeout` prop value is `0`, so the `if` statement here is intentional (`0` being falsy)
- Good flag on SSR, however, an SSR guard isn't required here as setTimeout/clearTimeout are supported in Node.js

brunnerh: - Just any regular set on the `timeout` after the toast has been closed, would trigger the reactive statement which schedules a timeout which in turn causes the event to fire. This is currently unlikely to happen intentionally because there is no way to reopen the component, as `open` is not accessible.
  In general the `open` state should probably be checked before scheduling the timeout.
- I know why the if is there, but it should only guard the `setTimeout`, not the `clearTimeout`.
  Right now, once the timeout has started, setting it to `0` will *not* cause `clearTimeout` to run, so it's impossible to properly stop the automatic close logic once it has started.
- Regarding SSR, one should not let timeouts like this start on the server, they serve no purpose can only cause problems (e.g. wasting resources, causing errors and taking the server with them, keeping things alive that should not be), right now the timeout will not be cleared properly either.

By the way, I would not be opposed to just dropping this automatic close logic from the component entirely.

If `open` were to be exposed, it would be possible to do this in user code and having to deal with this reactively here (because properties *might* change) just complicates things. It is already possible to do this user code to some degree by just destroying the component after a timeout (when using the client-side component API).

metonym: I've amended the code to address the following:

- Only start the timer if `open` is not already `false`
- Omit the `clearTimeout`
- Add in an SSR guard as we don't need this to run server-side

brunnerh (commented):
Looks better, but as it stands, the `timeout` has to be considered unchangeable after setting it once.

While the toast is open, the timeout cannot be disabled by setting the property to 0 and it cannot be changed to something else without starting a parallel timeout (with the old one not being cleaned up).

Changing the `timeout` later probably is not a common use case but this could be fixed or documented as a limitation. Could also look into a fix if you want. But I would have to look at this in more detail and do some testing. Reactive statements with multiple dependencies can be a bit tricky...

src/Notification/ToastNotification.svelte
```diff
@@ -66,14 +69,17 @@
...
-    }
-
     return () => {
       clearTimeout(timeoutId);
     };
   });
+
+  $: if (typeof window !== "undefined") {
+    /** Only start the timer of {@link open} has not been set to `false`. */
+    if (open && timeout) {
```
metonym, line 79:
```suggestion
    if (open && timeout && timeoutId === undefined) {
```

> While the toast is open, the timeout cannot be disabled by setting the property to 0 and it cannot be changed to something else without starting a parallel timeout (with the old one not being cleaned up).

@brunnerh Thanks for the review – regarding your latest feedback, WDYT about adding an additional check that a timer has not already been set?

brunnerh: I would add this:
```diff
  $: if (typeof window !== "undefined") {
+   clearTimeout(timeoutId);

    /** Only start the timer of {@link open} has not been set to `false`. */
    if (open && timeout) {
```

If the timeout is changed to 0, it cancels but no new timeout is started.
If the timeout is changed to something else, it also cancels and starts a new one (unless the toast is already closed).

<details>
<summary>Test code I used for switching timeouts</summary>

```svelte
<script>
    import { ToastNotification } from 'carbon-components-svelte';

	let timeout = 3000;
	let events = [];
</script>

<button on:click={() => timeout = 1000}>1s</button>
<button on:click={() => timeout = 3000}>3s</button>
<button on:click={() => timeout = 0}>disable</button>

<ToastNotification
	title="Auto-Close"
	{timeout}
	on:close={e => (events = [...events, e.detail])}>
	<svelte:fragment slot="subtitle">
		Timeout is {timeout}ms
	</svelte:fragment>
</ToastNotification>

<div>
	{#each events as event, i}
		<pre>{JSON.stringify(event, null, 2)}</pre>
	{/each}
</div>
```
</details>

metonym: Makes sense – thanks. Amended the commit with you as a co-author.

metonym:

Issue #1536, opened 2022-10-22
Disable "show more" button for CodeSnippet


j-tai:
There should be a way to disable the "show more" button in a multiline `CodeSnippet`, and just let it be scrollable.

Setting the prop `showMoreLess={false}` has no effect.

PR #1917, opened 2024-02-24
fix(code-snippet): `showMoreLess={false}` should hide button 

metonym:
Fixes https://github.com/carbon-design-system/carbon-components-svelte/issues/1536

Currently, the following does not work as intended:

```svelte
<CodeSnippet type="multi" {code} showMoreLess={false} />
```

The root cause is that `showMoreLess` is dynamically computed based on the code snippet height (the button is shown if a multi-line code snippet exceeds 255px). Additionally, the `showMoreLess` prop is initialized as `false` when it should be the opposite (allowing the consumer to set it to `false`).

This PR fixes the logic:

- flip the default `boolean` value
- only compute the show logic if `showMoreLess` is not `false`

Additionally, it adds the following examples to the docs:

- "Hidden show more button"
- "Hidden copy, show more buttons"


metonym:

Issue #1677, opened 2023-03-02
[ImageLoader] cache issues


tmarsik42:
I've come across a bug that affects `ImageLoader`.

Changing images in `ImageLoader` using reactive url parameter does not change an image.

Reproduction
https://svelte.dev/repl/acd821474a1648f688597ce4ffd4004b?version=3.50.1

1. Open console (so you can check logs)
2. Make sure Disable cache is DISABLED in network tab (unchecked)
<img width="537" alt="Snímek obrazovky 2023-03-02 v 9 13 47" src="https://user-images.githubusercontent.com/95856146/222369996-42a77a6a-7093-4d2b-9e01-d65b486b639a.png">
3. change images by clicking on list of images, html img changes just fine, `ImageLoader` doesn't change

<img width="673" alt="Snímek obrazovky 2023-03-02 v 9 14 50" src="https://user-images.githubusercontent.com/95856146/222370231-4a299285-6e1a-4bfe-b9bb-424f0cf3e170.png">

4. ENABLE "disable cache" option in console
<img width="541" alt="Snímek obrazovky 2023-03-02 v 9 17 10" src="https://user-images.githubusercontent.com/95856146/222370733-7fd9a5b1-a9c1-40a8-aac7-8ab56a6ad631.png">

5. It works now even for `ImageLoader`

<img width="670" alt="Snímek obrazovky 2023-03-02 v 9 17 33" src="https://user-images.githubusercontent.com/95856146/222370814-fa75e732-25bb-42c1-b6dd-11f669eebfdf.png">

I checked the implementation of `ImageLoader` but I'm not sure what the issue is.

Tihomir971:
Same problem

PR #1919, opened 2024-02-24
fix(image-loader): updated `src` should update the image 

metonym:
Fixes https://github.com/carbon-design-system/carbon-components-svelte/issues/1677

This one was interesting. The `src` prop was indeed correct, but the DOM was not updated to reflect this. The workaround here is to use a `key` block to recreate the element if `src` changes.


metonym:

PR #1926, opened 2024-03-07
Make library interoperable with Svelte 5

### LINKED ISSUE #1924: Link: do not render `<p>`
Original report: https://github.com/carbon-design-system/carbon-components-svelte/discussions/1908#discussioncomment-8698373

## Issue

Svelte 5 yields a compiler error for `<Link>` because `<div>` is an invalid descendent of `<p>`, which is correct as per W3 spec.

I recommend we align behaviour with [Carbon React](https://react.carbondesignsystem.com/?path=/story/components-link--playground&args=disabled:!true):

- [ ] When disabled, render an `<a aria-disabled="true">` (⚠️ breaking change since `<p>` has different layout behaviour)

Reference:

https://github.com/carbon-design-system/carbon-components-svelte/blob/96aef1705cc2812d47a8015a1499a6c6021ced0f/src/Link/Link.svelte#L37-L60

## Reproduction

Demo: https://stackblitz.com/edit/sveltejs-kit-template-default-mxnwjf?file=src%2Froutes%2F%2Bpage.svelte

Screenshot:

image

metonym:
Review by commit – rebase merge.

https://github.com/carbon-design-system/carbon-components-svelte/discussions/1908

There are several low-lift revisions that make this library interoperable with Svelte 5.

- [fix: avoid using reserved $ for Svelte 5 compat](https://github.com/carbon-design-system/carbon-components-svelte/commit/e5ea940da9643aa309cb5b3ea70b089c7dd1936a)
- [fix(link)!: do not render p for disabled link](https://github.com/carbon-design-system/carbon-components-svelte/commit/0b7aca13341d71062699c77a969626cd8e9ec62d) (Closes #1924)
- [fix(search): hoist ignore a11y autofocus comment](https://github.com/carbon-design-system/carbon-components-svelte/commit/6e9cd6e6af0d040554ac275cefa1689df4eacb25)

### DIFF
--- a/src/Search/Search.svelte
+++ b/src/Search/Search.svelte
@@ -79,6 +79,7 @@
   $: dispatch(expanded ? "expand" : "collapse");
 </script>
 
+<!-- svelte-ignore a11y-autofocus -->
 {#if skeleton}
   <SearchSkeleton
     size="{size}"
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -1963,15 +1963,15 @@ None.
 
 ### Props
 
-| Prop name | Required | Kind             | Reactive | Type                                                                   | Default value          | Description                                              |
-| :-------- | :------- | :--------------- | :------- | ---------------------------------------------------------------------- | ---------------------- | -------------------------------------------------------- |
-| ref       | No       | <code>let</code> | Yes      | <code>null &#124; HTMLAnchorElement &#124; HTMLParagraphElement</code> | <code>null</code>      | Obtain a reference to the top-level HTML element         |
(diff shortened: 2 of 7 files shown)


src/Search/Search.svelte
```diff
@@ -79,6 +79,7 @@
   $: dispatch(expanded ? "expand" : "collapse");
 </script>
 
+<!-- svelte-ignore a11y-autofocus -->
```
theetrain, line 82: Long-term, I recommend we remove `autofocus` functionality and guide users to instead utilize `ref` to manually set focus based on user or server events.

theetrain (approved):
Tested with `svelte@5.0.0-next.72`

src/Search/Search.svelte
metonym: Agreed – would love to sort out the ignorable a11y warnings from the rest.

PR #1927, opened 2024-03-07
fix(exports): resolve imports with explicit *.js extension

### LINKED ISSUE #1925: Unable to resolve *.js imports
Using the preprocessor, I'm getting this resolve error when attempting to import `breakpoints` from the source *.js` file.

```
Internal server error: Failed to resolve import "carbon-components-svelte/src/Breakpoint/breakpoints.js" from "src/App.svelte".
```

We need to explicitly map *.js files in the exports map in `package.json`:

```json
"./src/*.js": {
  "types": "./types/*.d.ts",
  "import": "./src/*.js"
},
```

metonym:
Fixes #1925

Importing using the explicit `.*js` extension will throw an error as these are not included in the package exports map. Normally, this would be acceptable if it's not part of the API.

However, the `optimizeImports` Carbon preprocessor rewrites certain imports (e.g., `breakpoints`) to the source `*.js` file.

```sh
Internal server error: Failed to resolve import "carbon-components-svelte/src/Breakpoint/breakpoints.js" from "src/App.svelte".
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,10 @@
       "types": "./types/*.svelte.d.ts",
       "import": "./src/*.svelte"
     },
+    "./src/*.js": {
+      "types": "./types/*.d.ts",
+      "import": "./src/*.js"
+    },
     "./src/*": {
       "types": "./types/*.d.ts",
       "import": "./src/*.js"


metonym:
Fixed in [v0.84.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.84.0)

PR #1932, opened 2024-03-09
feat(layer): first pass at `Layer` component

SimpleProgrammingAU:
I've had a crack at implementing the Layer component.

I wasn't sure how to approach porting the `useLayer` function from React.

Also, how should we set the `Context` value?


src/Layer/Layer.svelte
```diff
@@ -0,0 +1,45 @@
...
+   * Specify the HTML element to render. If none is specified, a `div` is rendered.
+   * @type {typeof import("svelte").SvelteComponent}
+   */
+  export let as = "div";
+
+  /**
+   * Specify the Layer HTML element props
+   * @type {import('svelte/elements').HTMLElementAttributes}
+   */
+  export let layerProps = {};
```
brunnerh, line 20: Since there is only one element I would stick to spreading `$$restProps`.

Is there a consensus how to handle this?

src/Layer/Layer.svelte
```diff
@@ -0,0 +1,45 @@
...
+  export let id = "ccs-" + Math.random().toString(36);
+
+  let ref = null;
+
+  // If there is no level override, determine the Level based on the hierarchy
+  const parentLevel = getContext("LAYER_CONTEXT");
+  if (level === undefined) {
+    level = typeof parentLevel === "number" ? parentLevel + 1 : 0;
+  }
+  setContext("LAYER_CONTEXT", level);
```
brunnerh, line 32: - This does not account for dynamically changing levels, which would require the use of a store in the context and a reactive statement to update the local level/store on change. Not sure if this is worth supporting, though.
- Personally, I would prefix context keys consistently to preempt collisions, e.g. use something like `'carbon.layer'`/`'carbon-components.layer'` in this case. If something like this is done, it should apply everywhere, though. So maybe a separate PR would make more sense.

src/Layer/Layer.svelte
```diff
@@ -0,0 +1,45 @@
...
+  export let as = "div";
+
+  /**
+   * Specify the Layer HTML element props
+   * @type {import('svelte/elements').HTMLElementAttributes}
+   */
+  export let layerProps = {};
+
+  /** Set an id for the Layer component */
+  export let id = "ccs-" + Math.random().toString(36);
```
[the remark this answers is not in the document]
SimpleProgrammingAU: TBH, it had been a while an I couldn't remember the rule for applying IDs by default to a Component. I see now it was just for form inputs.

src/Layer/Layer.svelte
SimpleProgrammingAU: @metonym @theetrain do you have a view on this?

src/Layer/Layer.svelte
SimpleProgrammingAU: Can we make the context reactive? Could something like this work?

```svelte
  $: parentLevel = getContext("LAYER_CONTEXT");
  $: {if (level === undefined) {
      level = typeof parentLevel === "number" ? parentLevel + 1 : 0;
    }
    setContext("LAYER_CONTEXT", level);
  }
```

Will take advice on the approach to the context name.

brunnerh: This will not work, contexts are fixed, they are set during child component construction.
That is why you would need a store.

SimpleProgrammingAU: Alright. I'll wait to see what others say about whether or not to use stores.

src/Layer/Layer.svelte
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * Specify the layer level and override any existing levels based on hierarchy.
+   * @type {0 | 1 | 2 }
+   */
+  export let level = undefined;
+
+  /**
+   * Specify the HTML element to render. If none is specified, a `div` is rendered.
+   * @type {typeof import("svelte").SvelteComponent}
```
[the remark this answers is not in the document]
brunnerh: @SimpleProgrammingAU (docs/types still need another re-build at some point)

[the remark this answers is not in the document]
SimpleProgrammingAU: Glad you raise this.

If I submitted all the changes that git tracks, it would be all the type files (right now, 167 files according to VSC).

Should we do a commit to v11/styles with a rebuild of the docs/types? Will that fix the issue in branches from v11?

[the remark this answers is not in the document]
brunnerh: I was just referring to the fact that `types/Layer/Layer.svelte.d.ts` still has the outdated property information (still contains `id` and different type for `as`).

Not sure what those other changes are/when to include them.

COMPONENT_INDEX.md
```diff
@@ -351,10 +352,18 @@ export type BreakpointValue = 320 | 672 | 1056 | 1312 | 1584;
...
+| :-------- | :------- | :--------------- | :------- | -------------------------------------------- | ------------- | ----------- |
+| sizes     | No       | <code>let</code> | Yes      | <code>Record<BreakpointSize, boolean></code> | <code>{       |
+
+     sm: false,
+     md: false,
+     lg: false,
+     xlg: false,
+     max: false,
+
+}</code> | Carbon grid sizes as an object |
```
brunnerh, line 365: Was confused by all the changes here, looks like code sections are not generated correctly. Pretty sure stuff like this breaks the table.

brunnerh: Could be a platform issue, `\r\n` vs just `\n` (i.e. Windows vs other systems).
In the JSON file there are now a bunch of `\r`, so the `\n` of an `\r\n` was probably stripped.

If this is correct, the issue might be in [sveld](https://github.com/carbon-design-system/sveld).

src/Layer/Layer.svelte
metonym: For v1, we want to avoid `$$restProps`, per #1621

COMPONENT_INDEX.md
SimpleProgrammingAU: Lemme know if there is anything I can do on my end to help resolve. I'm on a Win11 system using VSC.

As mentioned in a previous thread, I get heaps of changes in git when I do a rebuild of the docs. Could it be something to do with my configuration?

brunnerh: If you have the Git setting [`core.autocrlf`](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf) on `true`, changing it to `false` could help (a new checkout may also be necessary to update the files). This flag causes files to be checked out with `\r\n` rather than just `\n`, in most cases it should not be necessary, though.

Of course ideally the docs would be generated correctly either way.

metonym (approved):
Excellent, thank you!

PR #1933, opened 2024-03-12
fix(text-area): type `value` prop as nullable

CordlessWool:
Allow null as value attribute on TextArea as on TextInput

### DIFF
--- a/types/TextArea/TextArea.svelte.d.ts
+++ b/types/TextArea/TextArea.svelte.d.ts
@@ -5,10 +5,10 @@ type RestProps = SvelteHTMLElements["textarea"];
 
 export interface TextAreaProps extends RestProps {
   /**
-   * Specify the textarea value
+   * Specify the textarea value.
    * @default ""
    */
-  value?: string;
+  value?: null | string;
 
   /**
    * Specify the placeholder text
--- a/types/TextArea/TextArea.svelte.d.ts
+++ b/types/TextArea/TextArea.svelte.d.ts
@@ -5,10 +5,12 @@ type RestProps = SvelteHTMLElements["textarea"];
 
 export interface TextAreaProps extends RestProps {
   /**
-   * Specify the textarea value
+   * Specify the textarea value.
+   *
+   * and the value is empty.
    * @default ""
    */
-  value?: null | string;
+  value?: string;
 
   /**
    * Specify the placeholder text
--- a/types/TextArea/TextArea.svelte.d.ts
+++ b/types/TextArea/TextArea.svelte.d.ts
@@ -6,8 +6,6 @@ type RestProps = SvelteHTMLElements["textarea"];
 export interface TextAreaProps extends RestProps {
   /**
    * Specify the textarea value.
-   *
-   * and the value is empty.
    * @default ""
    */
   value?: string;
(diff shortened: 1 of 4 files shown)


types/TextArea/TextArea.svelte.d.ts
```diff
@@ -8,7 +8,7 @@ export interface TextAreaProps extends RestProps {
    * Specify the textarea value
    * @default ""
    */
-  value?: string;
+  value?: null | string;
```
brunnerh, line 11: These files are generated and should not to be edited manually.

The property would need a `@type {...}` annotation in the source code.

For reference, the property in `TextInput`:
https://github.com/carbon-design-system/carbon-components-svelte/blob/21b1d78d8ebda608d78c1ab22833c36138f9ef3c/src/TextInput/TextInput.svelte#L13-L20

But, I do not think changing this at all makes sense.

The input only has `null` to accommodate `type="number"` (see documentation above).

CordlessWool: Ok I first wanted to add it there, but then I was confused how it get the type if there is nothing defined before.

At least for the component it does not make a different if it is null or undefined, doesn't it? To put values in the user mabybe have to change the type or do a cast, without the need to, because the textarea component can handle null values as it do for undefined. That is currently our problem.

metonym (commented):
I think that it makes sense to loosen the `value` type to accept `null` so that it works in TypeScript strict mode.

Could you amend the JSDoc comment and run `yarn build:docs`? Doing so will auto-generate the corresponding TypeScript definitions.

brunnerh:
> I think that it makes sense to loosen the `value` type to accept `null` so that it works in TypeScript strict mode.

If the Svelte language tools were actually strict, this would just create problems elsewhere.
With `null` in the type, it should not be possible to `bind:value` to a `string` variable any more because `null` is not assignable to `string`.

If this were to cause errors on all existing bindings, I would be quite opposed to this, but it looks like this is not the case...

CordlessWool:
Since when is null related to number, null is not 0. To be 100% sure I also made a short implementation and I didn't had any problems with combining null, undefinded and string in combination with a bind. TS is in strict mode.

brunnerh:
> Since when is null related to number, null is not 0.

That was exactly my point, `null` is not compatible with other types, which is why such assignments are not allowed by TS itself.
The Svelte language tools (conveniently?) ignore this.

[Pure TS example](https://www.typescriptlang.org/play?jsx=0#code/DYUwLgBAxg9gtgBxgOxMsAFATjBItgCeAXBAM5hYCWyA5hAD4TICuwwEAvBAN4C+EAIZkhyQgG4AUKEgA3QdUEAjUABUAFoLABJMgCEYLZABNSFanS68Bw0RMmT5ilSA1bdBo8auxEKNJg4eAT2QA):

```ts
let componentProperty: string | null = {} as any;
let variableThatIsBound: string = {} as any;

variableThatIsBound = componentProperty;
```

Error on `variableThatIsBound`:
```
Type 'string | null' is not assignable to type 'string'.
  Type 'null' is not assignable to type 'string'.(2322)
```

CordlessWool:
Yes and exacatly this is why we need null as option because we have a variable that could be null | undefined | string and this is not assign able to undefined | string. To handle undefiend or null will make no difference for the component.

Undefiend will also not assign able to string and anyway is it allowed, so the comonent already handle it and you will be able to assign string to string | null.

metonym (approved):

metonym:

PR #1935, opened 2024-03-13
fix: DataTable, change result of sort to number

### LINKED ISSUE #1934: DataTable sort function return type is restricted to -1, 0 or 1
The expected return type for sorting rows in a DataTable is currently:
https://github.com/carbon-design-system/carbon-components-svelte/blob/21b1d78d8ebda608d78c1ab22833c36138f9ef3c/types/DataTable/DataTable.svelte.d.ts#L12

This restriction isn't used in DataTable.svelte and it wouldn't be a problem to use `number` as return type instead.

Doing so would give consumers of this library an easier way to write comparison functions, for example when sorting by Date: You could then write something like: `(a:Date, b:Date) => a.getTime() - b.getTime()` instead of having to write `Math.max(Math.min(a.getTime()-b.getTime(), 1), -1)`. This would also improve performance slightly.

berndfuhrmann:
Fixes #1934

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -1,6 +1,6 @@
 # Component Index
 
-> 165 components exported from carbon-components-svelte@0.83.1.
+> 165 components exported from carbon-components-svelte@0.84.0.
 
 ## Components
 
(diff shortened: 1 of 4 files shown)


metonym (approved):

metonym:
Released in [v0.84.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.84.1)

PR #1936, opened 2024-03-14
refactor: use strict equals

metonym:


src/ComposedModal/ComposedModal.svelte
```diff
@@ -67,7 +67,7 @@
     if (selectorPrimaryFocus == null) return;
     const node =
       (element || innerModal)?.querySelector(selectorPrimaryFocus) || buttonRef;
-    if (node != null) node.focus();
+    if (node !== null) node.focus();
```
brunnerh, line 70: Line above already has some optional chaining, could be used here as well.
```suggestion
    node?.focus();
```

brunnerh (commented):
Changes look fine, but their validity was non-trivial to check without IDE support on the various variable and function types.

Overall value of this change seems a bit questionable.

PR #1937, opened 2024-03-14
fix(checkbox): forward `focus`, `blur` events to `Checkbox` and `InlineCheckbox`

wickning1:


brunnerh (commented):
- Docs need to be rebuilt so the events are added to the `Forwarded events` lists.
- Order of `focus` and `blur` should ideally be consistent, though I don't think it matters.

wickning1:
Done.

brunnerh (approved):

metonym:
Released in [v0.84.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.84.1)

PR #1945, opened 2024-03-28
FIX: Problem with cursor jumps and input behaviour for NumberInput

divinedesign-se:
Fixes the jumping of input cursor in NumberInput when using keyboard backspace and delete to remove decimal values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,8 +176,8 @@ Import components from `carbon-components-svelte` in the `script` tag of your Sv
 
 [carbon-preprocess-svelte](https://github.com/carbon-design-system/carbon-preprocess-svelte) is a collection of Svelte preprocessors for Carbon.
 
-> [!NOTE]
-> Using `carbon-preprocess-svelte` is optional and not a prerequisite for this library.
+> [!INFO]
+> Using `carbon-preprocess-svelte` is optional and not a prerequisite to using this library. It's designed for better developer experience in addition to CSS build performance.
 
 ```sh
 # Yarn
--- a/README.md
+++ b/README.md
@@ -193,6 +193,7 @@ pnpm i -D carbon-preprocess-svelte
 ### `optimizeImports`
 
 `optimizeImports` is a Svelte preprocessor that rewrites barrel imports from Carbon components/icons/pictograms packages to their source Svelte code paths. This can significantly speed up development and production build compile times while preserving typeahead and autocompletion offered by integrated development environments (IDE) like VS Code.
+`optimizeImports` is a Svelte preprocessor that rewrites barrel imports from Carbon components/icons/pictograms packages to their source Svelte code paths. This can significantly speed up development and production build compile times while preserving typeahead and autocompletion offered by integrated development environments (IDE) like VS Code.
 
 The preprocessor optimizes imports from the following packages:
 
--- a/src/NumberInput/NumberInput.svelte
+++ b/src/NumberInput/NumberInput.svelte
@@ -18,6 +18,8 @@
    */
   export let value = null;
 
+  let previousValue = null;
+
   /** Specify the step increment */
   export let step = 1;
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -192,6 +192,7 @@ pnpm i -D carbon-preprocess-svelte
 
 ### `optimizeImports`
 
+`optimizeImports` is a Svelte preprocessor that rewrites barrel imports from Carbon components/icons/pictograms packages to their source Svelte code paths. This can significantly speed up development and production build compile times while preserving typeahead and autocompletion offered by integrated development environments (IDE) like VS Code.
 `optimizeImports` is a Svelte preprocessor that rewrites barrel imports from Carbon components/icons/pictograms packages to their source Svelte code paths. This can significantly speed up development and production build compile times while preserving typeahead and autocompletion offered by integrated development environments (IDE) like VS Code.
```
brunnerh, line 196: Why is this duplicated?

src/NumberInput/NumberInput.svelte
```diff
@@ -137,16 +141,67 @@
     $$props["aria-label"] ||
     "Numeric input field with increment and decrement buttons";
 
+  //$: value && !!changeCaretPosition;
```
brunnerh, line 144: If this is not needed, it should be removed.

src/NumberInput/NumberInput.svelte
```diff
@@ -99,12 +101,12 @@
...
-  import Subtract from "../icons/Subtract.svelte";
-  import WarningFilled from "../icons/WarningFilled.svelte";
-  import WarningAltFilled from "../icons/WarningAltFilled.svelte";
-  import EditOff from "../icons/EditOff.svelte";
+  import { createEventDispatcher, tick } from "svelte";
+  import Add from "carbon-components-svelte/src/icons/Add.svelte";
+  import Subtract from "carbon-components-svelte/src/icons/Subtract.svelte";
+  import WarningFilled from "carbon-components-svelte/src/icons/WarningFilled.svelte";
+  import WarningAltFilled from "carbon-components-svelte/src/icons/WarningAltFilled.svelte";
+  import EditOff from "carbon-components-svelte/src/icons/EditOff.svelte";
```
brunnerh, line 109: Don't see a compelling reason to change the imports here, it's not like this is going up five levels.

src/NumberInput/NumberInput.svelte
```diff
@@ -201,6 +256,7 @@
         max="{max}"
         min="{min}"
         step="{step}"
+        on:input="{updatePreviousValue}"
```
brunnerh, line 259: - Events should be at the end, after rest props.
- There already is an `input` event handler, the logic can be combined into/called from the existing `onInput` function; there should be no need to attach two different handlers.

src/NumberInput/NumberInput.svelte
```diff
@@ -137,16 +141,67 @@
...
+      !isIntegerTransition &&
+      previousValue &&
+      value === previousValue &&
+      inputType === "deleteContentForward"
+    ) {
+      value = parse(0);
+      await setSelectionAfterTick(0);
+    } else {
+      value = possibleNewValue;
+    }
```
brunnerh, line 188: This entire logic is hard to understand/follow, though I also don't know if/how it could be improved...

README.md
divinedesign-se: just a bad merge, sorry

src/NumberInput/NumberInput.svelte
divinedesign-se: true. Removing

src/NumberInput/NumberInput.svelte
divinedesign-se: Thats's fine. I forgot to change it back. This is a copy and paste from the code in our custom component mentioned in the issue, so the locations are different. Fixing.

PR #1968, opened 2024-05-04
Tweaks to DataTable generics

brunnerh:
Adjustments for #1954.
- Make key in `cell` slot prop less strict to prevent type errors in markup.
  Type errors are hard to deal with in Svelte 3/4, this could be reverted again in Svelte 5.
- No longer exclude `id` from possible keys.
- Resolve property path names up to one level deep for header keys.
  Limited to one level for now as combinatorial explosion can make type resolution quite heavy.
  For deeper levels a custom helper function could be used or, as with synthetic keys, `as any` can be used to ignore the type error.

I added some line breaks for readability in the source, these seem to be added to the `COMPONENT_API.json` but the docs are not visibly affected.

The types in `DataTableTypes.d.ts` are currently not directly accessible via the docs. (Also, the name and location of that file is debatable. The types are not really specific to this component and could potentially become useful elsewhere.)

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -926,7 +926,8 @@ None.
 ### Types
 
 ```ts
-export type DataTableKey<Row = DataTableRow> = Exclude<keyof Row, "id">;
+export type DataTableKey<Row = DataTableRow> =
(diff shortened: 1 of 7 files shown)


metonym:
The nested keys are great.

Screenshot 2024-05-04 at 2 37 16 PM

PR #1972, opened 2024-05-27
fix(multi-select): fix `filterItem` return type

eytanProxi:


metonym (approved):
Nice fix, thanks!

metonym:

PR #1983, opened 2024-07-16
chore(ibm-common): add mandatory IBM instrumentation tags

jeffchew:
Adding mandatory IBM intrumentation tags for Carbon Svelte.

### DIFF
--- a/docs/src/global.css
+++ b/docs/src/global.css
@@ -35,6 +35,11 @@ html[theme="g90"] .code-override {
   overflow-x: auto;
 }
 
+/* Override to accommodate mandatory footer */
+.bx--header~.bx--side-nav {
+  height: calc(100% - 120px);
+}
+
 /* Addig this to the layout grid fixes overstretching. */
 .fix-overflow {
   min-width: 0;
--- a/docs/src/global.css
+++ b/docs/src/global.css
@@ -37,7 +37,7 @@ html[theme="g90"] .code-override {
 
 /* Override to accommodate mandatory footer */
 .bx--header~.bx--side-nav {
-  height: calc(100% - 120px);
+  height: calc(100% - 100px);
 }
 
 /* Addig this to the layout grid fixes overstretching. */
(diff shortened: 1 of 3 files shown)


jeffchew:
@metonym fyi, this tag and footer needs to be added due to a recent IBM OPRT audit. This needs to be live by mid-August.

docs/src/global.css
```diff
@@ -35,6 +35,11 @@ html[theme="g90"] .code-override {
   overflow-x: auto;
 }
 
+/* Override to accommodate mandatory footer */
+.bx--header~.bx--side-nav {
+  height: calc(100% - 120px);
```
metonym, line 40: The magic number may need some adjusting, as the side nav appears to be cut off.

Screenshot 2024-07-17 at 8 45 21 AM

metonym: Screenshot 2024-07-17 at 8 46 26 AM

metonym (changes requested):
Appreciate the heads up – reviewed.

metonym (commented):
@jeffchew I'm viewing this locally on Chrome – weirdly, the Sidenav height doesn't seem to fill in the viewport height until the footer is in view.

https://github.com/user-attachments/assets/13ef7d43-2281-4d2f-8ef5-922834d13191

jeffchew:
The sidenav is fixed position and I think it probably won't be able to fill the viewport height without a layout rewrite which is what I was trying to avoid.

metonym:
Gotcha – I'll take a closer look at this.

jeffchew:
@metonym wondering if we can merge this for now? We need to complete the compliance remediation by the 15th.

metonym:
Moved to #1988 

I built upon this PR so that the footer does not obstruct the side nav.

Screenshot 2024-08-07 at 8 07 42 AM

jeffchew:
> Moved to #1988
> 
> I built upon this PR so that the footer does not obstruct the side nav.
> 

Thank you so much @metonym !

PR #1987, opened 2024-07-29
fix(search): collapse expandable search if value is falsy

### LINKED ISSUE #1981: Display glitch in Search component
There is an edge case in the Search component:

* Enter some whitespace in the search with `expandable=true`
* Blur the Search (focus something else or using JS)
* The Search collapses but the _clear_ button stays visible.

Screenshot:

image

This is due to the fact that the visibility of the _clear_ button is decided by `value === ''`:

https://github.com/carbon-design-system/carbon-components-svelte/blob/ba27830b2f6ad1fb2c8cc7dc1a8c833cf9e236d0/src/Search/Search.svelte#L162-L162

However the search is collapsed when the _trimmed_ value is empty: 

https://github.com/carbon-design-system/carbon-components-svelte/blob/ba27830b2f6ad1fb2c8cc7dc1a8c833cf9e236d0/src/Search/Search.svelte#L142-L146

Possible fixes:

* Do not collapse when the value is whitespace only.
* On blur, clear the value if it's whitespace only. However this mutates the user input, which _could_ be whitespace intentionally.
* Hide the clear button when `expanded` is false. Again this might hide the search when the contained value is whitespace.

The first version sounds the most consistent with API to me. Users can choose to clear the value if its empty on blur manually.

I can do a PR if this sounds good.

jmeurice:
Fixes #1981

Changed the value checks to also consider `null`/`undefined` since it's typed as `any`.

### DIFF
--- a/src/Search/Search.svelte
+++ b/src/Search/Search.svelte
@@ -140,7 +140,7 @@
       }}"
       on:blur
       on:blur="{() => {
-        if (expanded && value.trim().length === 0) {
+        if (expanded && (value === '' || value == null)) {
           expanded = false;
         }
       }}"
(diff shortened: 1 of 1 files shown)


metonym (approved):

metonym:

PR #1992, opened 2024-08-13
fix(header-action): allow vertical scroll when expanded

metonym:
This PR hot fixes a usability issue – if the `HeaderAction` has vertical overflow, the `overflow: hidden` rule is not overridden. Note that this should ideally be fixed upstream: https://github.com/carbon-design-system/carbon/pull/17160

In the meantime, the fix for Carbon Svelte ensures that the content is vertically scrollable (if necessary).

Screenshot 2024-08-13 at 9 27 13 AM


metonym:
Fixed in [v0.85.2](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.85.2)

PR #1994, opened 2024-08-20
build!: remove bundled ESM/UMD support

metonym:


metonym:
Closed in favor of #1995

PR #2011, opened 2024-10-20
fix: address Svelte 5 warnings

### LINKED ISSUE #2016: [Svelte 5] Development warnings
Using this library with Svelte 5 produces the following development warnings:

```sh
11:18:07 [vite] ✨ new dependencies optimized: svelte/internal/client
11:18:07 [vite] ✨ optimized dependencies changed. reloading
11:18:07 AM [vite-plugin-svelte] node_modules/carbon-components-svelte/src/Breakpoint/Breakpoint.svelte:43:48 Properties of objects and arrays are not reactive unless in runes mode. Changes to this property will not cause the reactive statement to update
11:18:08 AM [vite-plugin-svelte] node_modules/carbon-components-svelte/src/OverflowMenu/OverflowMenu.svelte:195:2 The value of 'aria-haspopup' must be exactly one of "false", "true", "menu", "listbox", "tree", "grid" or "dialog"
11:18:08 AM [vite-plugin-svelte] node_modules/carbon-components-svelte/src/TextInput/TextInput.svelte:97:24 Properties of objects and arrays are not reactive unless in runes mode. Changes to this property will not cause the reactive statement to update
11:18:08 AM [vite-plugin-svelte] node_modules/carbon-components-svelte/src/TextInput/PasswordInput.svelte:91:24 Properties of objects and arrays are not reactive unless in runes mode. Changes to this property will not cause the reactive statement to update
```

Although not blocking, these are annoying and should be addressed, especially if the fixes are backwards-compatible.

metonym:
Fixes #2016.

When using this library with Svelte 5, I'm seeing development warnings in the console.

```sh
11:18:07 [vite] ✨ new dependencies optimized: svelte/internal/client
11:18:07 [vite] ✨ optimized dependencies changed. reloading
11:18:07 AM [vite-plugin-svelte] node_modules/carbon-components-svelte/src/Breakpoint/Breakpoint.svelte:43:48 Properties of objects and arrays are not reactive unless in runes mode. Changes to this property will not cause the reactive statement to update
11:18:08 AM [vite-plugin-svelte] node_modules/carbon-components-svelte/src/OverflowMenu/OverflowMenu.svelte:195:2 The value of 'aria-haspopup' must be exactly one of "false", "true", "menu", "listbox", "tree", "grid" or "dialog"
11:18:08 AM [vite-plugin-svelte] node_modules/carbon-components-svelte/src/TextInput/TextInput.svelte:97:24 Properties of objects and arrays are not reactive unless in runes mode. Changes to this property will not cause the reactive statement to update
11:18:08 AM [vite-plugin-svelte] node_modules/carbon-components-svelte/src/TextInput/PasswordInput.svelte:91:24 Properties of objects and arrays are not reactive unless in runes mode. Changes to this property will not cause the reactive statement to update
```

This isn't a blocker, but it'd be nice to address these, especially if it's a backwards-compatible change.

My approach in #2010:

1. Upgrade tooling to Svelte 5 (including upgrading TypeScript).
2. Run `svelte-check` on source code and/or run `npm run build:lib` to identify warnings emitted by Svelte 5.
3. Address the warnings until they disappear.
4. Cherry pick https://github.com/carbon-design-system/carbon-components-svelte/pull/2010/commits/c2644a59baa72f7f01e3ddbe0b8183f58a651b8c and open this PR.

### DIFF
--- a/src/TextInput/PasswordInput.svelte
+++ b/src/TextInput/PasswordInput.svelte
@@ -88,6 +88,7 @@
 
   const ctx = getContext("Form");
 
+  // svelte-ignore reactive_declaration_non_reactive_property
   $: isFluid = !!ctx && ctx.isFluid;
   $: helperId = `helper-${id}`;
   $: errorId = `error-${id}`;
--- a/src/TextInput/PasswordInput.svelte
+++ b/src/TextInput/PasswordInput.svelte
@@ -88,8 +88,7 @@
 
   const ctx = getContext("Form");
 
-  // svelte-ignore reactive_declaration_non_reactive_property
-  $: isFluid = !!ctx && ctx.isFluid;
+  const isFluid = !!ctx && ctx.isFluid;
   $: helperId = `helper-${id}`;
   $: errorId = `error-${id}`;
   $: warnId = `warn-${id}`;
--- a/src/TextInput/TextInput.svelte
+++ b/src/TextInput/TextInput.svelte
@@ -94,6 +94,7 @@
     dispatch("change", parse(e.target.value));
   };
 
+  // svelte-ignore reactive_declaration_non_reactive_property
   $: isFluid = !!ctx && ctx.isFluid;
   $: error = invalid && !readonly;
   $: helperId = `helper-${id}`;
--- a/src/TextInput/TextInput.svelte
+++ b/src/TextInput/TextInput.svelte
@@ -94,8 +94,7 @@
     dispatch("change", parse(e.target.value));
   };
 
-  // svelte-ignore reactive_declaration_non_reactive_property
-  $: isFluid = !!ctx && ctx.isFluid;
+  const isFluid = !!ctx && ctx.isFluid;
   $: error = invalid && !readonly;
   $: helperId = `helper-${id}`;
   $: errorId = `error-${id}`;
--- a/src/Breakpoint/Breakpoint.svelte
+++ b/src/Breakpoint/Breakpoint.svelte
@@ -40,6 +40,7 @@
     max: size == "max",
   };
   $: if (size != undefined)
+    // svelte-ignore reactive_declaration_non_reactive_property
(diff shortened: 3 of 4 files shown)


metonym:
CC @brunnerh – would value any insight/thoughts you might have here. Short term goal is to avoid development warnings if using Svelte 5.

src/TextInput/PasswordInput.svelte
```diff
@@ -88,6 +88,7 @@
 
   const ctx = getContext("Form");
 
+  // svelte-ignore reactive_declaration_non_reactive_property
   $: isFluid = !!ctx && ctx.isFluid;
```
brunnerh, line 92: The warning here looks valid.
Contexts are fixed on component creation and properties themselves do not trigger reactivity, so this statement will only run once => could just be a regular variable.
```suggestion
  const isFluid = !!ctx && ctx.isFluid;
```

src/TextInput/TextInput.svelte
```diff
@@ -94,6 +94,7 @@
     dispatch("change", parse(e.target.value));
   };
 
+  // svelte-ignore reactive_declaration_non_reactive_property
   $: isFluid = !!ctx && ctx.isFluid;
```
brunnerh, line 98: Same as the password component.
```suggestion
  const isFluid = !!ctx && ctx.isFluid;
```

brunnerh (commented):
The approach seems valid, though all components should also be tested for compatibility. Since the Svelte 5 reactivity system is completely different, there is definitely potential for breakage.

metonym:
Great, thanks for the review. Keeping track of Svelte 5 compatibility issues in #2012.

metonym:
Fixed in [v0.85.3](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.85.3)

Issue #2013, opened 2024-10-20
[Svelte 5] `children` prop is reserved


metonym:
One of the Svelte 5 breaking changes is that the [`children` prop is reserved](https://svelte-5-preview.vercel.app/docs/breaking-changes#the-children-prop-is-reserved).

As such, components like `TreeView` and `RecursiveList` (which have `children` props) are affected.

From what I've observed:

- `TreeView` with Svelte 5 can render the tree, but toggling nodes does nothing (silently breaking).
- `RecursiveList` with Svelte 5 seems to render fine. Regardless, the prop should be renamed if Svelte 5 considers it reserved.

The fix would be to rename the `children` prop. This is a breaking change that is relatively easy to adapt to, and unblocks Svelte 5 usage.

Sheepux:
Combobox also has the issue 
`{#snippet children({ item })}`
It will not allow item as parameter

brunnerh:
That [should not be an issue](https://svelte.dev/playground/untitled?version=5.1.9#H4sIAAAAAAAACnWQwU7DMBBEf2W1RWoqWY2AWxoicePAH9Q9hGSrWnJsy15KUeR_R7ZcIILe7PXMrN_MaPqJsMEX0trCh_V6hIpGxTRuUOBRaQrY7GfkT5d0aYDi6np2bhvOpDnN3vpA_80Ha5gMB2ywDYNXjjtpJKvJWc_wqgLD0dsJ1ts6XYpxvZMmyQZrAoNimgI8wV3gnqna3wt4EPB42OykaeufVNPmvDnrY94zr4JRzhHDcFJ69GSqOedB3MQkkNyO6txlT2zrdM6-uvhiWpFiO2lQINOFsWH_TlHcqOUXxrKX5cONYuiSi9FUsP8gfv9wRf1wKt30IR-uREHbUkOEuvAkdYZJAUuSQ_wCIj8DTQoCAAA=).
Did you get type errors?

Sheepux:
Yeah, as it was complaining about Snippet and item I thought it was the same issue.

> Type '{ id: string; placeholder: string; selectedId: string; size: "sm"; items: ComboBoxItem[]; shouldFilterItem: (item: any, value: string) => any; children: ({ item }: { item: any; }) => any; }' is not assignable to type 'Properties<ComboBoxProps, { default: { item: ComboBoxItem; index: number; }; titleText: {}; }>'.
[...]

PR #2014, opened 2024-10-20
feat(tree-view)!: rename `children` prop to `nodes` for Svelte 5 compatibility

metonym:
Fixes [#2013](https://github.com/carbon-design-system/carbon-components-svelte/issues/2013)

`TreeView` silently fails when using it with Svelte 5 because the [`children` prop is reserved](https://svelte.dev/docs/svelte/v5-migration-guide#The-children-prop-is-reserved).

To fix this, rename the `children` prop to `nodes`. Additionally, rename the `children` property on `nodes` to `nodes`. This is to avoid confusion, but it's also necessary since `TreeViewNodeList` is a recursive component (i.e., passes props to `svelte:self`).

This is a breaking change but should be relatively straightforward to adapt to.


metonym:

PR #2015, opened 2024-10-20
feat(recursive-list)!: rename `children` prop to `nodes` for Svelte 5 compatibility

metonym:
Fixes [#2013](https://github.com/carbon-design-system/carbon-components-svelte/issues/2013)


metonym:

PR #2027, opened 2024-10-25
Use type alias instead of interface

metonym:
This resolves the TypeScript errors for the generated types, but now causes issues for consumption.

This is because certain attributes that exist on RestProps do not overlap, resulting in  `Error: Type <X> is not assignable to type 'never'. (ts)`.

The overlapping custom props would need to be omitted from RestProps, e.g., `Omit<RestProps, "size">`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
         "rollup-plugin-terser": "^7.0.2",
         "sass": "^1.49.11",
         "standard-version": "^9.5.0",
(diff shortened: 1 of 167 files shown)


metonym:
Needs something like:

```ts
type RestProps = SvelteHTMLElements["input"];

type ComponentProps = { /** ... */ };

export type SearchProps = Omit<RestProps, keyof ComponentProps> & ComponentProps;
```

PR #2026, opened 2024-10-25
ci: add script to test generated types

metonym:
The generated TypeScript definitions need to be periodically validated.

A flurry of recent PRs have aimed to correct type issues:

- #2019
- #2020
- #2022
- #2023
- #2027

Additionally, TypeScript was upgraded to the latest minor version in #2024.

This PR adds a script to type check generated definitions for the library to ensure basic workability and guard against regressions.

The current check should expectedly fail as there are still TypeScript errors that I'm working through.

PR #2030, opened 2024-11-03
fix(types)!: use type alias instead of interface for $$restProps

metonym:
Closes #2027

#2027 updates the auto-generated types the components props extending rest props. See [`sveld@0.21.0`](https://github.com/carbon-design-system/sveld/releases/tag/v0.21.0) for more details.

This PR cherry picks the last two commits from #2027 so that #2026 can be merged with CI green.

---

Even though this fixes a type error, this will be considered a breaking change since the `<Component>Props` type is exported, and thus may be relied upon by consumers.

The breaking change is that the component props type changes from an interface to a type alias.

```diff
- export interface AccordionProps extends AccordionSkeletonProps {
+ export type AccordionProps = AccordionSkeletonProps & {
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
         "rollup-plugin-terser": "^7.0.2",
         "sass": "^1.49.11",
         "standard-version": "^9.5.0",
-        "sveld": "^0.20.3",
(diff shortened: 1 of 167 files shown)


metonym:

PR #2034, opened 2024-11-09
fix(multi-select): avoid cyclic dependency for Svelte 5 compatibility

### LINKED ISSUE #1986: [Svelte 5] `MultiSelect` triggers `effect_update_depth_exceeded` error
An `effect_update_depth_exceeded` error is printed to the console when a MultiSelect component is present on the page. I have tested it in `svelte@5.0.0-next.197`, but also checked some earlier versions.

## Console output
```
Uncaught (in promise) Svelte error: effect_update_depth_exceeded
Maximum update depth exceeded. This can happen when a reactive block or effect repeatedly sets a new value. Svelte limits the number of nested updates to prevent infinite loops
    effect_update_depth_exceeded errors.js:171
    infinite_loop_guard runtime.js:477
    flush_sync runtime.js:681
    flush_sync runtime.js:698
    flush_sync runtime.js:698
    flush_sync runtime.js:698
    flush_sync runtime.js:698
    flush_sync runtime.js:698
    flush_sync runtime.js:698
    flush_sync runtime.js:698
errors.js:171:16
```

## Reproduction
https://github.com/malinowskip/carbon-multiselect-svelte-5-error

malinowskip:
Fixes #1986

Here’s an attempt to fix the issues in the `MultiSelect` component that caused incompatibility with Svelte 5 by creating infinite loops (https://github.com/carbon-design-system/carbon-components-svelte/issues/1986). It seems like the core issue was the circular dependency between the `items` and `sortedItems` variables:

`items` assigned to`sortedItems` in an `afterUpdate` block:
https://github.com/carbon-design-system/carbon-components-svelte/blob/ac38e9d3531af85b413ba9919fdc70bca62c2002/src/MultiSelect/MultiSelect.svelte#L261

`sortedItems` depends on `items`:
```svelte
  $: sortedItems = items.map((item) => ({
    ...item,
    checked: selectedIds.includes(item.id),
  }));
```

The commit removes the `sortedItems` variable, which served an intermediate role, and changes are made to `items` directly.

### DIFF
--- a/src/MultiSelect/MultiSelect.svelte
+++ b/src/MultiSelect/MultiSelect.svelte
@@ -168,7 +168,7 @@
    */
   export let highlightedId = null;
 
-  import { afterUpdate, createEventDispatcher, setContext } from "svelte";
+  import { afterUpdate, createEventDispatcher, setContext, tick } from "svelte";
   import WarningFilled from "../icons/WarningFilled.svelte";
   import WarningAltFilled from "../icons/WarningAltFilled.svelte";
   import Checkbox from "../Checkbox/Checkbox.svelte";
@@ -237,7 +237,7 @@
   afterUpdate(() => {
     if (checked.length !== prevChecked.length) {
       if (selectionFeedback === "top") {
-        sortedItems = sort();
+        items = sort();
       }
       prevChecked = checked;
       selectedIds = checked.map(({ id }) => id);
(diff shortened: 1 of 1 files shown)


src/MultiSelect/MultiSelect.svelte
```diff
@@ -247,33 +247,29 @@
         unselected: unchecked,
       });
     }
+  });
 
-    if (!open) {
-      if (!initialSorted || selectionFeedback !== "fixed") {
-        sortedItems = sort();
+  $: if (!open) {
```
malinowskip, line 252: Inside `afterUpdate`, this block would cause an infinite loop.

src/MultiSelect/MultiSelect.svelte
```diff
@@ -247,33 +247,29 @@
...
+    if (!initialSorted || selectionFeedback !== "fixed") {
+      tick().then(() => {
```
malinowskip, line 254: `tick` ensures that the variables are instantiated when the sorting happens.

metonym (commented):
Thanks for taking the time to get this going. You're spot on in identifying the root issue of the infinite update loop in Svelte 5. There's a cyclical `items <> sortedItems` dependency in `afterUpdate`.

However, one thing I noticed is that this current change does not support [initially set `selectedIds`.](https://svelte.carbondesignsystem.com/components/MultiSelect#initial-selected-items)

```svelte
<MultiSelect
  selectedIds={["0", "1"]}
  titleText="Contact"
  label="Select contact methods..."
  items={[
    { id: "0", text: "Slack" },
    { id: "1", text: "Email" },
    { id: "2", text: "Fax" },
  ]}
/>
```

metonym:
I've pushed a [6fbedab](https://github.com/carbon-design-system/carbon-components-svelte/pull/2034/commits/6fbedabfd8038e3a3c965e62dad1e24e2d294efa) that does the following:

- Remove the cyclical dependency in `afterUpdate`
- Consolidate around `sortedItems` internally to avoid the cyclical dependency

I've so far tested with both Svelte 4 and Svelte 5.

metonym (approved):

malinowskip:
Thank for reviewing and catching the `selectedIds` issue. I think that your solution introduces one breaking change, though: the sorting logic doesn’t update the `items` variable, so the new sort order is not propagated to the parent (if `items` is bound reactively).

brunnerh:
Arguably that was a bug; the property was not marked as being reactive and who really wants this?

malinowskip:
It is marked as reactive in the docs (https://svelte.carbondesignsystem.com/components/MultiSelect#props), though, as you say, maybe it’s not useful for this prop to be reactive.

metonym:
> Arguably that was a bug; the property was not marked as being reactive and who really wants this?

Nice catch – `MultiSelect` items should not be reactive, and that was identified via this cyclic dependency fix. Generated docs need to be updated to reflect this – created #2042

metonym:

PR #2037, opened 2024-11-10
fix(data-table): move `ToolbarSearch` filtering to `DataTable` for Svelte 5 compatibility

### LINKED ISSUE #2033: [Svelte 5] Filterable `DataTable` does not reset correctly
eg.

Search Terms -> Shows Filtered Rows 

-> Remove Search Term -> Filtered Rows is still the same

metonym:
Fixes #2033 

Svelte 5 exposes a bug when using `DataTable` and `ToolbarSearch`. Initial filtering might work, but original rows are not reset.

The root cause is a reset loop caused by `ToolbarSearch`. It currently reads and writes `$tableRows`, which is the source of truth used by `DataTable`.

One of the constraints is for `filteredIds` to be bindable to `ToolbarSearch` (but not two-way reactivity). Beyond that, `ToolbarSearch` should not need to read the source `rows` or manipulate them directly.

The solution is to move the filtering logic to the parent `DataTable`. A copy of the original rows is set (unless `rows` changes). Filtering will then filter on a subset of the copy without modifying the source.

### DIFF
--- a/src/DataTable/DataTable.svelte
+++ b/src/DataTable/DataTable.svelte
@@ -147,6 +147,10 @@
   const dispatch = createEventDispatcher();
   const batchSelectedIds = writable(false);
   const tableRows = writable(rows);
+
+  // Store a copy of the original rows for filter restoration.
+  $: originalRows = [...rows];
+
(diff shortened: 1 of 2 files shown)


metonym:

PR #2044, opened 2024-11-10
fix(checkbox): bind indeterminate for Svelte 5 compatibility

### LINKED ISSUE #2039: [Svelte 5] `Checkbox` indeterminate state not updating
In Svelte 5, the following example from the docs doesn’t work: https://svelte.carbondesignsystem.com/components/Checkbox#indeterminate.

```svelte
<Checkbox labelText="Label text" indeterminate />
```

The initial indeterminate icon is frozen regardless of the current `checked` value.

While looking into this, I also discovered that `indeterminate` can’t be set as a HTML attribute, but should be set dynamically in JavaScript (source: [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes)).

The problem could be fixed by removing the `indeterminate` HTML attribute and setting the value dynamically:

```js
onMount(() => {
  if (ref && indeterminate) {
    ref.indeterminate = true;
  }
});
```

metonym:
Fixes #2039

In Svelte 5, `<input indeterminate={indeterminate} />` does not update the value. Binding the variable seems to work.

Tested the following with Svelte 4 and Svelte 5. The difference now is that `indeterminate` is reactive (two-way).

```svelte
<script>
  import { Checkbox } from "carbon-components-svelte";

  let indeterminate = true;
</script>

<Checkbox labelText="Label text" bind:indeterminate />

<p>Indeterminate: {indeterminate}</p>

<button on:click={() => (indeterminate = !indeterminate)}>Toggle indeterminate</button>
```

---

https://github.com/user-attachments/assets/7814d0e7-2c92-4e7e-ad1d-981043a878b1

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -458,10 +458,10 @@ None.
 | :------------ | :------- | :--------------- | :------- | ----------------------------------------- | ------------------------------------------------ | ------------------------------------------------- |
 | ref           | No       | <code>let</code> | Yes      | <code>null &#124; HTMLInputElement</code> | <code>null</code>                                | Obtain a reference to the input HTML element      |
 | title         | No       | <code>let</code> | Yes      | <code>string</code>                       | <code>undefined</code>                           | Specify the title attribute for the label element |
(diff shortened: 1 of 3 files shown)


metonym:

PR #2082, opened 2025-01-08
fix(data-table): prefix internal ID for radio button, checkbox

### LINKED ISSUE #2081: Selecting a row on one DataTable unchecks radio button on another DataTable
Problem: 
- Selecting a row in one DataTable will result in unselecting the radio button of a row with the same id of another DataTable on the same page. 

Situation: 
- Two DataTables on the same page. 
- DataTables using radio buttons for selection. 
- Both DataTables use the same id property for the rows.

Reproduce issue:
- Example on: https://svelte.dev/playground/b1b4a7ad80bb4885a74622c9029f40fe?version=5.16.5
- Select first row on first DataTable (id abc123), than select first row on second DataTable (id abc123).
- Notice radio button being unchecked in the first DataTable.

Analysis
See analysis of @theetrain on [Discord](https://discord.com/channels/689212587170201628/1048235099583418448/1325963576816435230). 

> It appears that when `radio={true}` on the DataTable, the name attribute is set as the row's ID. I think the radio prop ought to accept a string as the desired name, and then its value attribute would be the Row's ID. 
> In other words, a row with an id `abc123` should have a radio input like this:
> `<input type="radio" name="table-1-row" value="abc123" />`
> Link to code:  https://github.com/carbon-design-system/carbon-components-svelte/blob/f3a8d9972c2c4019b71a45416056ed6deaa988d3/src/DataTable/DataTable.svelte#L502

Workaround: 
- use different id's for DataTables.

metonym:
Fixes #2081

Create an internal ID prefix for the radio buttons and checkboxes used within `DataTable`.

This addresses the case where multiple `DataTable` instances may be instantiated, and the data contains overlapping IDs in `rows`.

### DIFF
--- a/src/DataTable/DataTable.svelte
+++ b/src/DataTable/DataTable.svelte
@@ -170,6 +170,10 @@
   const batchSelectedIds = writable(false);
   const tableRows = writable(rows);
 
+  // Internal ID prefix for radio buttons, checkboxes, etc.
+  // since there may be multiple `DataTable` instances that have overlapping row ids.
+  const id = "ccs-" + Math.random().toString(36);
+
   // Store a copy of the original rows for filter restoration.
   $: originalRows = [...rows];
 
@@ -499,7 +503,7 @@
               {#if !nonSelectableRowIds.includes(row.id)}
                 {#if radio}
                   <RadioButton
-                    name="select-row-{row.id}"
+                    name="{id}-{row.id}"
                     checked={selectedRowIds.includes(row.id)}
                     on:change={() => {
                       selectedRowIds = [row.id];
@@ -508,7 +512,7 @@
                   />
                 {:else}
                   <InlineCheckbox
-                    name="select-row-{row.id}"
+                    name="{id}-{row.id}"
                     checked={selectedRowIds.includes(row.id)}
                     on:change={() => {
                       if (selectedRowIds.includes(row.id)) {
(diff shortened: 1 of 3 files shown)


src/DataTable/DataTable.svelte
```diff
@@ -499,7 +503,7 @@
               {#if !nonSelectableRowIds.includes(row.id)}
                 {#if radio}
                   <RadioButton
-                    name="select-row-{row.id}"
+                    name="{id}-{row.id}"
                     checked={selectedRowIds.includes(row.id)}
                     on:change={() => {
                       selectedRowIds = [row.id];
```
theetrain, line 509: If we can, I suggest taking this opportunity to provide a solution that affords progressive enhancement:

- Have the `name` attribute be defined by the component consumers; perhaps exposed as a prop, that way every Radio button in the table has a matching `name` to be gathered server-side via `formData.get('school')`
- Have the `value` attribute match the given Row's ID, available in `row.id`

In this way, a DataTable can be wrapped in `<form>` and be submittable without JS.

metonym: Good call. You're right – the `name` should in fact be the same for all rows in the table (barring the select all).

metonym:

bhavers:
Thank you, i updated and it works perfectly.

PR #2084, opened 2025-01-19
fix: avoid manual `blur` in `Dropdown`, `MultiSelect` to allow keyboard navigation in Firefox

### LINKED ISSUE #2083: Accessibility issue on Firefox: Can not tab out of Dropdown
Problem: 

- **Dropdown is not keyboard-accessible**. Can tab into it and select the different items. **But can not tab out of it.**
- Works on Chrome, but not in Firefox

Reproduce issue:
- Issue exists in the embedded Dropdowns in the documentation as well: https://carbon-components-svelte.onrender.com/components/Dropdown
 To Reproduce issue: Use a firefox browser (whether developer edition or regular). Tab through the page and you'll get stuck and not able to tab out of the Dropdowns.
  -  It works on Google Chrome. Also, compared this to React's Carbon Component Dropdown. And it does not have the same issue. 

Is there any discussion around getting a fix for this issue soon? As my current app uses this Dropdown quite a lot and we're trying to get the app to be accessible. Probably will switch over to the Select component for now bc of limited time. But am interested in any potential updates and discussions around this. Thank you.

metonym:
Fixes #2083

There's a bug in Firefox where the focus is trapped inside the Dropdown button and MultiSelect field. I was able to repro with the latest FF version 134.0.1 (64-bit).

From my research, the symptoms seem similar to this [bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=53579) (although the report itself is over a decade old, and has reportedly been fixed).

This removes the unnecessary manual `blur` in the "Tab" events. I've also tested Chrome, Safari, and the behavior is consistent.

### DIFF
--- a/src/Dropdown/Dropdown.svelte
+++ b/src/Dropdown/Dropdown.svelte
@@ -250,7 +250,6 @@
           }
         } else if (key === "Tab") {
           open = false;
-          ref.blur();
         } else if (key === "ArrowDown") {
           if (!open) open = true;
           change(1);
--- a/src/MultiSelect/MultiSelect.svelte
+++ b/src/MultiSelect/MultiSelect.svelte
@@ -374,7 +374,6 @@
             selectionRef.focus();
           } else {
             open = false;
-            fieldRef.blur();
           }
         } else if (key === "ArrowDown") {
           change(1);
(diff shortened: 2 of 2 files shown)


metonym:

PR #2086, opened 2025-01-20
fix(data-table): improve `expandable` accessibility

metonym:
Improves the accessibility of an expandable `DataTable`.

Adds in missing:

- `aria-controls` < > `id` attributes
- Sets `aria-hidden="true"` on expandable icons since the button already has a label


metonym:

PR #2087, opened 2025-01-20
feat(data-table): allow custom `inputName` for radio/checkbox

### LINKED ISSUE #2085: `DataTable` supports `name` prop for selectable variant
> If we can, I suggest taking this opportunity to provide a solution that affords progressive enhancement:
> - Have the `name` attribute be defined by the component consumers; perhaps exposed as a prop, that way every Radio button in the table has a matching `name` to be gathered server-side via `formData.get('school')`
> - Have the `value` attribute match the given Row's ID, available in `row.id`

> In this way, a DataTable can be wrapped in `<form>` and be submittable without JS.

_Originally posted by @theetrain in https://github.com/carbon-design-system/carbon-components-svelte/pull/2082#discussion_r1906238924_

For a selectable or radio selectable table, the `checkbox/input` should all have the same `name` value and encode the toggled value.

metonym:
Closes #2085 

Adds an `inputName` prop to `DataTable` to allow customizing the `name` value set on the `input` for selectable rows (checkbox or radio button).

`name` should be the same on all rows in a `DataTable` but unique across multiple `DataTable` instances. The row input IDs remain unique to avoid clashing (see #2081).


metonym:
CC @theetrain does this address your use case?

metonym:
Released in [v0.88.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.88.0)

PR #2089, opened 2025-01-21
fix(text-area): counter supports null `value`

bartosjiri:
The [TextArea](https://svelte.carbondesignsystem.com/components/TextArea) component allows `null` [value](https://github.com/carbon-design-system/carbon-components-svelte/blob/master/src/TextArea/TextArea.svelte#L2-L6), however the component breaks when a null value is provided/bound to the value property since "_Cannot read properties of null (reading 'length')_" at [line 88](https://github.com/carbon-design-system/carbon-components-svelte/blob/master/src/TextArea/TextArea.svelte#L88).

### DIFF
--- a/src/TextArea/TextArea.svelte
+++ b/src/TextArea/TextArea.svelte
@@ -85,7 +85,7 @@
       </label>
       {#if maxCount}
         <div class:bx--label={true} class:bx--label--disabled={disabled}>
-          {value.length}/{maxCount}
+          {value?.length || 0}/{maxCount}
         </div>
       {/if}
     </div>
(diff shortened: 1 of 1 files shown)


src/TextArea/TextArea.svelte
```diff
@@ -85,7 +85,7 @@
       </label>
       {#if maxCount}
         <div class:bx--label={true} class:bx--label--disabled={disabled}>
-          {value.length}/{maxCount}
+          {value?.length || 0}/{maxCount}
```
metonym, line 88: Let's coalesce the `value` to an empty string.

```suggestion
          {((value ?? '').length)}/{maxCount}
```

metonym (approved):

metonym:
Fixed in [v0.87.2](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.87.2) – thanks!

PR #2095, opened 2025-02-03
fix(types): loosen `icon` prop type to `any`

metonym:
https://github.com/carbon-design-system/carbon-icons-svelte/issues/207

`carbon-icons-svelte@13` and `carbon-pictograms-svelte@13` now only offer TypeScript support Svelte 4/5.

However, the new `Component` type is incompatible with `icon` prop types in the `carbon-components-svelte` library. If using Svelte, a type error is shown for the `icon` prop, since `typeof SvelteComponent` is not compatible with the new `Component` type from Svelte 5.

Unfortunately, `Component` is not available in Svelte 3 or 4. For this library to be compatible with Svelte 3, 4, and 5, this PR loosens the `icon` prop type to `any`.


metonym:
Released in [v0.87.4](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.87.4)

PR #2097, opened 2025-02-04
fix(tag): allow `on:close` to work with Svelte 5

### LINKED ISSUE #2096: Svelte 5 onclose and on:close don't work on Tag
Problem: 
The onclose (and on:close) event on the Tag component doesn't seem to work from a Svelte 5 component. Regular dom events, like click (onclick and on:click), do fire. 
Here's an example of the close event that doesn't work (event doesn't fire)
```
<Tag size="default" id="test" filter
  onclose={(e) => {
    console.log('closing tag');
  }}>TEST</Tag>
```
Analysis:
The close event is using svelte 4 dispatch.
line 86: https://github.com/carbon-design-system/carbon-components-svelte/blob/master/src/Tag/Tag.svelte
```
on:click|stopPropagation={() => {
  dispatch("close");
}}
```

Could it be that, because dispatch is deprecated in Svelte 5, it doesn't work if the event is being dispatched to a component running in svelte 5 runes mode?

Maybe [this issue](https://github.com/sveltejs/svelte/issues/12147) is related?

metonym:
Fixes #2096

Interestingly, `<Tag filter on:close />` does not fire because the forwarded `on:click` uses the `stopPropagation` modifier.

Removing it allows the dispatched "close" event to fire. `on:close` is effectively an alias for `on:click`, which is only forwarded to the button for a filterable tag.

## Expected

Both "click" and "close" should be logged.

```svelte
<Tag
  filter
  on:click={() => {
    console.log("click");
  }}
  on:close={() => {
    console.log("close");
  }}
>
  Tag
</Tag>
```

## Expected

If the consumer manually stops propagation, "click" is logged but "close" will not. This works in Svelte 5, but not Svelte 4.

```svelte
<Tag
  filter
  on:click={(e) => {
    e.stopPropagation();
    console.log("click");
  }}
  on:close={() => {
    console.log("close");
  }}
>
  Tag
</Tag>
```

### DIFF
--- a/src/Tag/Tag.svelte
+++ b/src/Tag/Tag.svelte
@@ -82,7 +82,7 @@
       class:bx--tag__close-icon={true}
       {disabled}
       {title}
-      on:click|stopPropagation
+      on:click
       on:click|stopPropagation={() => {
         dispatch("close");
       }}


metonym:
Released in [v0.87.5](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.87.5)

PR #2100, opened 2025-02-17
fix(overflow-menu): add `aria-controls` to trigger button

metonym:
Adds a missing `aria-controls` to the trigger element. The associated ID is the overflow menu.


metonym:
Fixed in [v0.87.6](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.87.6)

PR #2108, opened 2025-03-07
fix(select): avoid infinite update loop in Svelte 5

### LINKED ISSUE #2107: [Svelte 5] `Select` causes `effect_update_depth_exceeded` error when binding to same object
If multiple `Select` instances bind to properties of the same object in legacy mode, this will cause an `$effect` loop because each select will invalidate the source object in turn.

E.g.
```svelte
<script>
	import { Select, SelectItem } from 'carbon-components-svelte@0.87';

	let product = {
		productType: 0,
		priceCurrency: 0,
	};
</script>
<svelte:head>
	<link rel="stylesheet" href="https://unpkg.com/carbon-components-svelte@0.87/css/g90.css"/>
</svelte:head>

<Select bind:selected={product.productType}
	labelText="Type"
	required>
	<SelectItem hidden/>
	<SelectItem value={0} text="Single" />
	<SelectItem value={1} text="Set" />
</Select>

<Select bind:selected={product.priceCurrency}
	labelText="Currency"
	required>
	<SelectItem value={0} text="USD" />
	<SelectItem value={1} text="EUR" />
</Select>
```
[Playground](https://svelte.dev/playground/5827754bf06e41d28ee9b1163d2e3ddd?version=5.22.5) (see console)

There would not be an issue if the bound values were separate variables or if `$state` were to be used (since that is more granular).

I looked at the source of `Select` and a potential fix might be to not set `selected` unless it needs updating.

```diff
  afterUpdate(() => {
+   if (selected !== $selectedValue) {
      selected = $selectedValue;
+   }

    if (prevSelected !== undefined && selected !== prevSelected) {
      dispatch("update", $selectedValue);
    }

    prevSelected = selected;
  });
```

metonym:
Fixes #2107 – applies the suggestion from https://github.com/carbon-design-system/carbon-components-svelte/issues/2107#issue-2900333913

Tested with Svelte 4 and Svelte 5.

### DIFF
--- a/src/Select/Select.svelte
+++ b/src/Select/Select.svelte
@@ -111,10 +111,11 @@
   let prevSelected = undefined;
 
   afterUpdate(() => {
-    selected = $selectedValue;
-
-    if (prevSelected !== undefined && selected !== prevSelected) {
-      dispatch("update", $selectedValue);
+    if (selected !== $selectedValue) {
+      selected = $selectedValue;
+      if (prevSelected !== undefined) {
+        dispatch("update", $selectedValue);
+      }
     }
 
     prevSelected = selected;


metonym:
Fixed in [v0.87.7](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.87.7)

PR #2111, opened 2025-03-08
feat(ui-shell): `HeaderAction` supports tooltip

### LINKED ISSUE #2110: Use `Button` in `HeaderAction`
Analogous to:
- #1893

Would allow setting an `iconDescription` and related properties, generating a nice tooltip.

metonym:
Closes #2110 

#1894 added tooltip support for `HeaderGlobalAction`. Similarly, this PR adds tooltip support for `HeaderAction`. The difference between these two components is that `HeaderAction` opens a panel (right-aligned to the browser) when toggled. This feature is also supported by the upstream [React implementation](https://react.carbondesignsystem.com/iframe.html?globals=&args=&id=components-ui-shell-header--header-w-actions-and-switcher&viewMode=story).

**Sample Usage**

```svelte
<HeaderAction
  iconDescription="View switcher"
  tooltipAlignment="end"
/>

```

Screenshot 2025-03-08 at 11 47 43 AM

**Implementation Notes**

Initially, I tried refactoring to re-use `Button`. However, the internal `hasIconOnly` is not triggered since the `HeaderAction` *does* have a default slot. This is because it supports text being displayed alongside the icon (this should be better documented). Additionally, this component has an open and closed state.

As a result, I opted for a more minimal approach: applying the minimal required `Button` classes to support the icon tooltip. Notably, if either the `text` prop or "text" slot is used, the tooltip will not be shown.


metonym:
Released in [v0.88.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.88.0).

Docs demo [Header with multiple switchers](https://svelte.carbondesignsystem.com/components/UIShell#header-with-multiple-switchers)

PR #2118, opened 2025-03-11
fix(select): falsy item `text` should not override value

### LINKED ISSUE #2117: `SelectItem` text cannot be set to empty string if there is a value.
This will show `-1` which is unintended:

```svelte
<Select labelText="Type">
	<SelectItem hidden value={-1} text="" />
	<SelectItem value={0} text="Single" />
	<SelectItem value={1} text="Set" />
</Select>
```
[Playground](https://svelte.dev/playground/untitled?version=5.22.6#H4sIAAAAAAAACoWPwU7DMBBEf8XaSy9pHG4lJFU5cm45YQ6ps22sOmvL3lZUUf4dpS4FBILTSjNPMzsDUNMjlPBMbNhiCxnsjMUI5csAfPaTNwmQfZCP3ufxhJYnbdtE_E3XjhiJI5RQRR2M56Uixab3LrAYxBotas6u94mxF6PYBdeLmW7C1tFcu947mjLmKXVV5ItFXsweFFXyllkls-ywaS8VlTV0EAFtrSDy2WLsEFmB6ALuagUds4-llEfyh32uXS__KZQ6Rrm_L3IdowK5vNR_L6Uq7RC22aLd4BvXCjZnjwrST19mdqZtkcSpsUesh_ndKDjxCoT8QV-x4katDe0t_sV-Jq4vu9PHiVsqggwmG0oORxxfx3fE0F87AgIAAA==)

The source has code like this:
```svelte
<option {value}>
  {text || value}
</option>
```

(I would generally advise to never use `||` for anything that is not a boolean.)

Personally, I would prefer if `SelectItem` would follow the semantics of the plain `<option>` elements, which people might already be used to. Here the `value` follows the text content if no explicit `value` is set, not the other way around. ([Examples](https://svelte.dev/playground/untitled?version=5.22.6#H4sIAAAAAAAACp2QMWvDMBCF_4o4CpbBddrVtQ2hdOnQDGmnOIOjnIpBkVzpHBKM_3uRFTdpSZdMh969797petD1DiGDD00NKdxCArJR6CBb9UDH1ve8AMnknLdt6vaoyGub2uE1XRhNqMlBBrkTtmmprHRFdyglCuI8ZkXJei9VJI1lXBjtiDlUKIgZybZGdDvUlH51aI_LUTd2rhSPgimK42lARYHe16rDF4WeY8V5hLBY09TgUWsxip9-o32AE6bqDSo2sOK0SxoKbhctNUa71cN6Ii_jUsIDPYdPs4K9LhdvqSPb6M9GHvnf6ef0U0gtCS2_HDhZBl9GIJ-dD6nzAI5Hzc24WvmOB8pnp8fo__Fc84edigoquBm9f6ygvBn8PxUS8AeFjGyHw3r4BoABkO2nAgAA))

---

**Workaround:** Set `text` that does not evaluate to falsy, e.g. a space or zero-width-space.
```svelte
<SelectItem hidden value={-1} text=" " />
```

metonym:
Fixes #2117

The logical OR operator in `SelectItem` currently defaults to `value` if the provided `text` is falsy (e.g., empty string).

Intead, it should coalesce to `value` if `text` is `null` or `undefined`.

### DIFF
--- a/src/Select/SelectItem.svelte
+++ b/src/Select/SelectItem.svelte
@@ -55,5 +55,5 @@
   class={className}
   {style}
 >
-  {text || value}
+  {text ?? value}
 </option>
--- a/tests/Select/Select.falsy.test.svelte
+++ b/tests/Select/Select.falsy.test.svelte
@@ -1,0 +1,9 @@
+<script lang="ts">
+  import { Select, SelectItem } from "carbon-components-svelte";
+</script>
+
+<Select labelText="Falsy text">
+  <SelectItem value={-1} text="" />
+  <SelectItem value={0} text="Zero" />
+  <SelectItem value={1} text="One" />
+</Select>
--- a/tests/Select/Select.test.ts
+++ b/tests/Select/Select.test.ts
@@ -3,6 +3,7 @@ import { user } from "../setup-tests";
 import Select from "./Select.test.svelte";
 import SelectGroup from "./Select.group.test.svelte";
 import SelectSkeleton from "./Select.skeleton.test.svelte";
+import SelectFalsy from "./Select.falsy.test.svelte";
 
 describe("Select", () => {
   beforeEach(() => {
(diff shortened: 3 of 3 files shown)


metonym:
Fixed in [v0.88.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.88.1)

PR #2119, opened 2025-03-12
fix(ui-shell): `HeaderAction` uses dark color scheme

metonym:
Currently, the UI Shell only supports the dark theme. The `HeaderAction` panel should instruct browsers to apply a dark color scheme to the scrollbar.

## Before

Screenshot 2025-03-11 at 5 12 45 PM

## After

Screenshot 2025-03-11 at 5 12 35 PM


metonym:
Fixed in [v0.88.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.88.1)

PR #2121, opened 2025-03-14
fix: ComboBox tab handling

b-r-i-a-n-w-e-s-t:
Trying to fix two issues with this:

1. when tabbing into a combobox, the initial character was not being input. Fixed via afterUpdate adjustment.
2. when tabbing into a combobox that already had a selection, the current selection was highlighted, but typing would not overwrite it. Instead it would place text (after skipping the first character) after the selection. Fixed via additional on:keydown

### DIFF
--- a/src/ComboBox/ComboBox.svelte
+++ b/src/ComboBox/ComboBox.svelte
@@ -175,7 +175,6 @@
       filteredItems = [];
       if (!selectedItem) {
         selectedId = undefined;
-        value = "";
         highlightedIndex = -1;
         highlightedId = undefined;
       } else {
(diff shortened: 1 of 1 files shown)


metonym:
Hey, there – thank you for flagging these bugs.

I revised the implementation – the `value` update in `afterUpdate` is necessary to handle the case where the `selectedId` is programmatically unset by the consumer (this [example](https://svelte.carbondesignsystem.com/components/ComboBox#reactive-example)).

It seems the fix is to only update `value` if the input is not focused.


https://github.com/user-attachments/assets/c0ff8db0-690e-46dc-a454-abd1b1d46f73

b-r-i-a-n-w-e-s-t:
Thanks for taking a look at these issues. 

Looks like your changes have solved all the problems I was seeing.

metonym:

PR #2125, opened 2025-03-19
fix(list-box): use `aria-disabled` instead of invalid `disabled` attribute

metonym:
`ListBoxMenuItem` is used by `ComboBox`, `Dropdown`, and `MultiSelect`.

When the item has `disabled: true`, the disabled state for the item is applied. However, the `disabled` attribute on `<div role="option"` is invalid markup. This fixes it to use `aria-disabled` instead.


metonym:
Initially released in [v0.88.2](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.88.2) but reverted in [v0.88.3](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.88.3) due to a styling regression.

PR #2130, opened 2025-03-19
Revert "fix(list-box): use `aria-disabled` instead of invalid `disabled` attribute"

metonym:
Reverts carbon-design-system/carbon-components-svelte#2125

It turns out that changing `disabled` --> `aria-disabled` did cause a style regression here. Carbon v10 styles targets the `[disabled]` attribute.

Reverting this change.

Screenshot 2025-03-19 at 1 01 43 PM


metonym:
Fixed in [v0.88.3](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.88.3)

PR #2134, opened 2025-03-22
fix(list-box-selection): fix `aria-label` for clear button

metonym:
Identified in #2131

`ListBoxSelection` is used by `MultiSelect` and `ComboBox`.

Currently, it uses the incorrect aria-label for the clear selection button. It's using the `translateId` (e.g., `"clearAll"`) instead of the user-friendly copy.

A follow-up would be to adjust the copy based on `selectionCount`. For example, for `MultiSelect`, there may be multiple values selected. For `ComboBox`, it wouldn't make much sense to have "Clear all selected items" if there is only one selected item.


metonym:

PR #2135, opened 2025-03-22
fix(radio-button): forward `focus`, `blur` events

metonym:
Identified in https://github.com/carbon-design-system/carbon-components-svelte/pull/2131

`focus` and `blur` events should be forwarded to the underlying `input` element used in `RadioButton`.

```svelte
<RadioButton on:focus on:blur />
```


metonym:

PR #2136, opened 2025-03-22
fix(radio-tile): allow standalone `RadioTile` usage

metonym:
Identified in https://github.com/carbon-design-system/carbon-components-svelte/pull/2131

Although `RadioTile` is intended to be used inside a `TileGroup`, it feels unpolished for standalone usage to crash. This is because no parent context is found.

This fixes `RadioTile` to fail open by supplying an `add: () => {}` no-op in the case where `TileGroup` context does not exist.


metonym:

PR #2138, opened 2025-03-23
fix(list-box): set `aria-disabled` if `disabled` (#2125)

metonym:
This re-applies https://github.com/carbon-design-system/carbon-components-svelte/pull/2125, which was reverted in #2130

Carbon styles depend on the `[disabled]` attribute. The `disabled` attribute is not valid mark-up when applied to `div role="option"`, so this PR preserved the existing `disabled` attribute while also setting `aria-disabled`.


metonym:

PR #2139, opened 2025-03-23
fix(list-box): correct button/description translations based on selection count

metonym:
The `ListBoxSelection` component now properly handles translations for the clear button based on the selected items:

- Fix `buttonLabel` and `description` to use the same translation logic
- Add tests for custom translations in both `ComboBox` and `MultiSelect`

Stylistically, use an explicit `selectionCount === undefined` to use the default `clearSelection` id.


metonym:

PR #2152, opened 2025-04-18
SelectItem: fix to use the value when text is undefined after #2118

wickning1:
#2118 introduced a bug where `<SelectItem value="blah" />` rendered as `<option value="blah"></option>` instead of `<option value="blah">blah</option>`. 

This PR fixes that issue while preserving the intended fix in #2118 of making `<SelectItem value="blah" text="">` render as `<option value="blah"></option>`. I have also updated the automated tests to cover both cases.

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -3220,14 +3220,14 @@ None.
 
 ### Props
 
-| Prop name | Required | Kind             | Reactive | Type                              | Default value          | Description                               |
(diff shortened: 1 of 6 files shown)


metonym (approved):
Nice find, thank you.

metonym:
Released in [v0.89.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.89.0)

PR #2153, opened 2025-04-18
feat(combo-box): add `hideLabel` prop

wickning1:
Many form components have a `hideLabel` prop that makes the label text only show up for screenreaders. It's useful when the placeholder is descriptive enough, but it was missing from ComboBox even though ComboBox does have a placeholder. This PR adds the functionality.

Worth noting: this component uses `titleText` instead of `labelText`, which is inconsistent with (most) other components and could make `hideLabel` confusing. I went with `hideLabel` because it's more consistent with the other components and we're still literally referring to a label role in the DOM. I would hope we could change `titleText` to `labelText` in a future breaking-changes-are-ok update.


metonym (approved):
Fantastic, thank you! I added a unit test and updated the docs.

metonym:
Released in [v0.89.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.89.0)

PR #2154, opened 2025-04-19
fix(toolbar-search): re-filter rows if `DataTable` rows change

### LINKED ISSUE #2143: `DataTable` rows used by `ToolbarSearch` filtering is not reactive
I'm using DataTable with Search Toolbar and my rows array is a Svelte store. (`rows={$rows}`)

When search is applied and rows value change, DataTable is refreshed without filter. It would be nice if it could detect rows change and run `filterRows` function by default, or if an option is enabled.

In the meantime, I resorted to creating a child component in DataTable for this:

```js
// hacky component, child of DataTable and sibling of Toolbar
<script>
import { getContext } from "svelte";

// searchValue requires bind:value={searchValue} in ToolbarSearch component and passing it to this component as prop
let { rows = $bindable(), searchValue = $bindable() } = $props()

const ctx = getContext("DataTable") ?? {};

$effect(() => {
    if (rows.length > 0) {
      ctx?.filterRows(searchValue, true)
    }
})
```

metonym:
Fixes #2143

Currently, `ToolbarSearch` filtering is not reactive. The component reads `DataTable` rows passed via Svelte context. The expected behavior is that the filtered rows are updated if `rows` passed to `DataTable` (the source of truth) updates.

The solution is to subscribe to the rows passed down from `DataTable`. The key is to only run `filterRows` in the `afterUpdate` callback to avoid an infinite update loop.

### DIFF
--- a/src/DataTable/ToolbarSearch.svelte
+++ b/src/DataTable/ToolbarSearch.svelte
@@ -46,14 +46,38 @@
    */
   export let ref = null;
 
-  import { tick, getContext } from "svelte";
+  import { tick, getContext, afterUpdate, onMount } from "svelte";
   import Search from "../Search/Search.svelte";
 
   const ctx = getContext("DataTable") ?? {};
 
+  let rows = null;
+  let unsubscribe = null;
+
   $: if (shouldFilterRows) {
-    filteredRowIds = ctx?.filterRows(value, shouldFilterRows);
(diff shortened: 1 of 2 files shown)


metonym:
Fixed in [v0.89.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.89.1)

PR #2158, opened 2025-04-25
fix(to-hierarchy): revert to previous implementation

### LINKED ISSUE #2157: toHierarchy utility broke at version 89.0
I use the toHierarchy utility to feed the nodes property of TreeView. It doesn't return the hierarchy properly anymore since version 89.0

Instruction to reproduce issue:

- Visit: [https://svelte.dev/playground/331d2cff80cd4e368f0721926e4b8a25?version=5.28.2](https://svelte.dev/playground/331d2cff80cd4e368f0721926e4b8a25?version=5.28.2)
- Switch between version 88.4 and 89.0 by commenting out the import statement at the top.
- With version 88.4 you will see three folders at top level with items in it (this is correct)
- With version 89.0 and onwards you will five folders (with items in it) + one item at top level (this is incorrect).

Note: for experimental reasons i copied in the example from the docs also in this test, that can be ignored.

Kind regards,
Bram

metonym:
Fixes #2157

Reverts #2148

### DIFF
--- a/src/utils/toHierarchy.js
+++ b/src/utils/toHierarchy.js
@@ -13,22 +13,35 @@
 export function toHierarchy(flatArray, getParentId) {
   /** @type {NodeLike[]} */
   const tree = [];
-  const nodeMap = new Map();
+  const childrenOf = new Map();
+  const itemsMap = new Map(flatArray.map((item) => [item.id, item]));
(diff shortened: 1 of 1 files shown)


metonym:

PR #2159, opened 2025-04-25
fix(composed-modal): ignore a11y warning in Svelte 5

metonym:
Silence another a11y warning if using Svelte 5.


metonym:

PR #2160, opened 2025-04-26
fix(pagination): window `totalItems` for performance

### LINKED ISSUE #2156: The whole application freezes due to the pagination component
If you pass a large number (for example - 500000) to the totalItems prop, the entire application freezes

metonym:
Fixes #2156 

The number of native select options rendered in `Pagination` is derived from the `totalItems` provided.

If `totalItems` is a large number (e.g., 500,000), the component attempts to render 50k options (assuming the default page size is `10`). Modern browsers can typically render ~15k.

Arguably, the best UX is to avoid this scenario entirely. For example, use `Pagination` with `totalItems` but don't render the page selector.

```svelte
<Pagination totalItems={102} pageInputDisabled />
```

However, it's best to tolerate larger numbers by default. It should be considered a bug that the component freezes the browser if `totalItems` is a large number. This PR uses a windowed approach. By default, cap the number of native options rendered to 1,000. This value can be overridden using the `pageWindow` prop.


metonym:

PR #2161, opened 2025-04-26
fix(pagination): use `toLocaleString` for default text formatting

metonym:
Use `toLocaleString() for default `Pagination` text formatting.

E.g., `10000` --> `10,000`


metonym:

PR #2170, opened 2025-06-10
forward on:input from Multiselect text input when filterable prop is true

wickning1:
I ran into a case where bind:value wasn't very convenient because I'm generating multiple Multiselect components inside an `{#each}` and trying to search for options individually.

```
{#each category as cat}
  <Multiselect on:input={(e) => { searchForOptions(cat.id, e.target.value) }} />
{/each}
```
is a lot easier.


metonym (approved):
LGTM

metonym:
Released in [v0.89.4](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.89.4)

PR #2178, opened 2025-08-05
fix(checkbox): prevent potential effect loop.

### LINKED ISSUE #2177: [Svelte 5] `Checkbox` causes `effect_update_depth_exceeded` error when binding to same object
This can happen in Svelte 5 if multiple checkboxes bind to the same object. E.g.

```js
let obj = {
	a: false,
	b: false,
};
```
```svelte
<Checkbox bind:checked={obj.a} labelText="A" />
<Checkbox bind:checked={obj.b} labelText="B" />
```
[Playground](https://svelte.dev/playground/726793f01fa14611819e7e0c7a7992f1?version=5.37.3)

> playground:output:377 Uncaught Svelte error: effect_update_depth_exceeded
Maximum update depth exceeded. This typically indicates that an effect reads and writes the same piece of state
https://svelte.dev/e/effect_update_depth_exceeded

brunnerh:
Removes dependency of `checked` on itself which causes issues if an entire object is invalidated that way.

Fixes #2177.

### DIFF
--- a/src/Checkbox/Checkbox.svelte
+++ b/src/Checkbox/Checkbox.svelte
@@ -61,7 +61,7 @@
   const dispatch = createEventDispatcher();
 
   $: useGroup = Array.isArray(group);
-  $: checked = useGroup ? group.includes(value) : checked;
+  $: if (useGroup) checked = group.includes(value);
   $: dispatch("check", checked);
 
   let refLabel = null;


metonym:
Fixed in [v0.89.5](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.89.5). Thank you!

PR #2180, opened 2025-08-06
chore: replace npm with Bun

metonym:
This PR replaces npm with Bun as the package manager and JavaScript runtime for the carbon-components-svelte codebase, modernizing the development toolchain with significant performance and developer experience improvements.

**Benefits**

- **Performance**: Faster installs, faster test runs, faster CI pipelines
- **Consistency**: All Carbon Svelte repositories now use the same toolchain
- **Modern DX**: Native TypeScript execution and Bun APIs


brunnerh:
Personally, I would prefer Deno.
Though is there any real need for switching in the first place?

PR #2184, opened 2025-08-16
fix(toggle): dispatch change event on change, not as effect

spburtsev:
Changes in this PR make the Svelte implementation consistent with React's behavior where the toggle event acts as onChange wrapper.

In current implementation toggle event is dispatched as an effect whenever `toggled` state is changed. Because of that, toggle event handler runs when component is mounted, which does not make sense.
Also, in React implementation it works differently and `onToggle` handler behaves like a convenient onChange wrapper:
https://stackblitz.com/edit/vitejs-vite-ze6q4qgm?file=src%2FApp.tsx

In svelte, `on:toggle` handler is called when component is mounted, which is very counter-intuitive and is harder to work-around:
https://stackblitz.com/edit/vitejs-vite-eyr7trez?file=src%2FApp.svelte

### DIFF
--- a/src/Toggle/Toggle.svelte
+++ b/src/Toggle/Toggle.svelte
@@ -39,8 +39,6 @@
   import { createEventDispatcher } from "svelte";
 
   const dispatch = createEventDispatcher();
-
-  $: dispatch("toggle", { toggled });
 </script>
 
 <!-- svelte-ignore a11y-mouse-events-have-key-events -->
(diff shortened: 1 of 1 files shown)


metonym (approved):
Thanks so much!

metonym:
Fixed in [v0.89.6](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.89.6)

Issue #2172, opened 2025-07-16
Accessibility Violations in MultiSelect Component (carbon-components-svelte@^0.89.4) using IBM Equal Access Checker


PratikAnand09:
We’ve encountered several accessibility violations using the MultiSelect component from carbon-components-svelte, flagged by the IBM Equal Access Accessibility Checker.

These issues affect keyboard and screen reader users, and impact WCAG 2.1 compliance.

Component Affected:
MultiSelect (filterable variant, multiple selection)

Accessibility Violations:

-     The tabbable element does not have a valid widget role
-     Accessible name does not match or contain the visible label text
-     The element with role="button" contains descendants with roles="button"
-     The element with role="button" contains descendants with role="combobox"

   

Impact
These issues:

    Break keyboard accessibility (users can tab to elements with no valid role).

    Confuse screen reader users due to invalid role nesting.

    Violate WCAG 2.1 Success Criteria (e.g., 1.3.1, 4.1.2).

Environment:

    carbon-components-svelte: ^0.89.4

    Browser: Firefox + Chrome (latest)

    Accessibility checker: IBM Equal Access Accessibility Checker

<img width="1736" height="962" alt="Image" src="https://github.com/user-attachments/assets/3e928cf0-af01-4601-9cb3-4a5846bf62f8" />

<img width="1736" height="962" alt="Image" src="https://github.com/user-attachments/assets/c8b28572-ef85-4653-a9c9-dd575c6e6478" />

metonym:
Thanks for raising this.

metonym:
The ComboBox a11y issues are addressed in [v0.89.7](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.89.7).

MultiSelect needs reworking, so I will keep this issue open.

metonym:
Fixed in [v0.89.8](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.89.8)

PR #2186, opened 2025-08-17
fix(combo-box): address a11y issues

metonym:
Supports #2172

Fixes accessibility issues for `ComboBox`, using the IBM Accessibility Checker extension.

The crux of the issue is that the usage of `ListBoxField` adds an element with redundant attributes (e.g., button role, aria-expanded, id, disabled, translateWithId). These attributes are already present on the input.

The second issue is that the `input` should use `role="combobox" type="text"`, which mirrors the upstream React implementation. This requires updating matchers in the test suite.

Screenshot 2025-08-17 at 12 30 58 PM

Screenshot 2025-08-17 at 12 28 23 PM


metonym:

PR #2187, opened 2025-08-17
fix(list-box): address a11y issues

metonym:
Supports #2172

This actually may not be the correct fix; I'm opening this PR as a reference.

The issue is more so in the usage of `ListBoxField`, where there are nested buttons. Referencing the upstream React implementation, these buttons should be siblings.

Screenshot 2025-08-17 at 12 47 30 PM


metonym:

PR #2188, opened 2025-08-17
fix(multi-select): address a11y issues

metonym:
Supports #2172

Related #2186

This attempts to fix a11y issues reported by the IBM Accessibility Checker extension.

The default (non-filterable) `MultiSelect` has different semantics than the filterable variant.

This PR currently contains a fix for the nested buttons by making them siblings. Review with [whitespace-hidden.](https://github.com/carbon-design-system/carbon-components-svelte/pull/2188/files?w=1)

This PR is not complete:

- Non-filterable uses a button with combobox role, e.g.,:

```svelte
<button type="button" aria-describedby="multiselect-helper-text-id-«r2e»" aria-activedescendant="" aria-controls="downshift-«r2g»-menu" aria-expanded="false" aria-haspopup="listbox" aria-labelledby="downshift-«r2g»-label" id="downshift-«r2g»-toggle-button" role="combobox" tabindex="0" />
```

- Filterable variant uses an input, and the list box field should not have any a11y attributes, similar to #2186

```svelte
<input aria-activedescendant="" aria-autocomplete="list" aria-expanded="false" autocomplete="off" id="carbon-multiselect-example-3-input" role="combobox" aria-describedby="filterablemultiselect-helper-text-id-«r4»" aria-haspopup="listbox" value="">
```

### DIFF
--- a/src/MultiSelect/MultiSelect.svelte
+++ b/src/MultiSelect/MultiSelect.svelte
@@ -347,63 +347,7 @@
         class="bx--list-box__invalid-icon bx--list-box__invalid-icon--warning"
       />
     {/if}
-    <ListBoxField
-      role="button"
-      tabindex="0"
-      aria-expanded={open}
-      on:click={() => {
-        if (disabled) return;
-        if (filterable) {
-          open = true;
-          inputRef.focus();
-        } else {
-          open = !open;
-        }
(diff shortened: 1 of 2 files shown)


metonym:

PR #2194, opened 2025-08-27
fix(pagination): `on:change` dispatches with correct value

metonym:
Currently, `on:change` in Pagination dispatches with the previous value. Pagination should use `on:update` instead.


metonym:

PR #2195, opened 2025-08-29
fix(data-table): handle dynamic `headers`

### LINKED ISSUE #2193: DataTable crashes on Svelte 5.38.2 and above
Situation:
I have Carbon Tabs. In each tab i render a DataTable.
That table can contain different number of columns (DataTableHeaders).
See: https://mybram.com/ and switch between Work tab (five columns) and Personal tab (three columns). 
Since Svelte 5.38.2 switching between Work and Personal tabs causes DataTable to crash.

Problem determination:
When the amount of headers (DataTableHeader) differs (so instead of five headers, i give it three headers), the DataTable component crashes.
This happens only happens when i render the DataTable at the moment that i select another tab. When i render all tabs at once it does not crash.
As this is hard to explain, see reproduction below.

Reproduction:
See playground: https://svelte.dev/playground/b281263c84d5440f9e9203a16edd4f4a?version=5.36.0 and follow instructions.

Additionally:
On my local machine this started since Svelte 5.38.2.
On the playground this started since Svelte 5.36.0

The playground is a simplified version of my application. In my application i get into an indefinite loop with many errors. That might be because that DataTable is handling more data and is more complex. But that loop looks like a race condition problem, though i'm not sure. That said, in think the behaviour of the simplified version on playground is the root of the problem.

metonym:
Fixes #2193

In #2193, a `DataTable` with dynamic, mismatched headers and rows can crash with:

```sh
TypeError: Cannot read properties of undefined (reading 'empty')
```

If keys are mismatched between headers, `#if headers[j].empty` may cause a runtime error (i.e., the error message above).

The root cause was in the reactive statement creating `tableCellsByRowId` which used `headerKeys.map()` but then accessed headers[index].

The solution is to consolidate `headers` directly, and stub out potentially missing properties to eliminate the potential for the out-of-bounds array access.

### DIFF
--- a/src/DataTable/DataTable.svelte
+++ b/src/DataTable/DataTable.svelte
@@ -263,12 +263,13 @@
     expanded = expandedRowIds.length === expandableRowIds.length;
   }
   $: if (radio || batchSelection) selectable = true;
-  $: headerKeys = headers.map(({ key }) => key);
   $: tableCellsByRowId = rows.reduce((rows, row) => {
-    rows[row.id] = headerKeys.map((key, index) => ({
-      key,
-      value: resolvePath(row, key),
-      display: headers[index].display,
+    rows[row.id] = headers.map((header, index) => ({
+      key: header.key || `key-${index}`,
+      value: header.key ? resolvePath(row, header.key) : undefined,
+      display: header.display,
+      empty: header.empty,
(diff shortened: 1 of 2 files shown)


metonym:

PR #2207, opened 2025-10-14
fix(text-input): wrap `value` prop description to one line

metonym:
Fixes prop wrapping for `TextInput` value. It currently appears malformed in the docs.

---

## Before

Screenshot 2025-10-14 at 2 59 30 PM

## After

Screenshot 2025-10-14 at 2 48 31 PM


metonym:

PR #2211, opened 2025-10-15
fix(modal): `ModalFooter` prop type `secondaryButtons` allows empty array

metonym:
Identified via #2210

`ModalFooter` secondaryButtons is typed as a two-tuple. However, the default value is an empty array. The prop type should allow an empty array to match this default value.


metonym:

PR #2213, opened 2025-10-16
fix(data-table): default filtering supports nested keys

### LINKED ISSUE #1897: Support nested properties in DataTable filtering search
Here I have a simple example of what I mean:
```html
<DataTable
  headers={[
    { key: "id", value: "id" },
    { key: "contact.company", value: "Company name" },
  ]}
  {rows}
  {pageSize}
  {page}
>
  <Toolbar>
    <ToolbarContent>
      <ToolbarSearch
        persistent
        value=""
        shouldFilterRows
        bind:filteredRowIds
      />
    </ToolbarContent>
  </Toolbar>
</DataTable>

<Pagination
  bind:pageSize
  bind:page
  totalItems={filteredRowIds.length}
  pageSizeInputDisabled
/>
```

This example works by showing the correct field in the table. However, when using the search filter it will not find anything in the contact.company field, but it will find a match in the id field.

My guess is the culprit is somewhere here:
https://github.com/carbon-design-system/carbon-components-svelte/blob/f1cafd49596e8957ac436a3c7cb7b2645ff86417/src/DataTable/ToolbarSearch.svelte#L57-L67

The table itself I seen in a past issue had a similar problem.
https://github.com/carbon-design-system/carbon-components-svelte/pull/602/files

My guess would be the fix now lives around here in the more modern version of the code:
https://github.com/carbon-design-system/carbon-components-svelte/blob/f1cafd49596e8957ac436a3c7cb7b2645ff86417/src/DataTable/DataTable.svelte#L199-L206

I'm not familiar enough with all pieces involved to make a proper PR, but it seems like a reasonably straight forward thing to fix from the little I do understand about the code. It would be greatly appriciated.

metonym:
Fixes #1897

The default datatable filter behavior should account for nested keys, expressed via dot notation (i.e., `a.b.c`). This adds a regression test.

```svelte
<DataTable
  headers={[
    { key: "id", value: "ID" },
    { key: "contact.company", value: "Company name" },
  ]}
  {rows}
>
  <Toolbar>
    <ToolbarContent>
      <ToolbarSearch persistent {value} shouldFilterRows bind:filteredRowIds />
    </ToolbarContent>
  </Toolbar>
</DataTable>
```

### DIFF
--- a/src/DataTable/DataTable.svelte
+++ b/src/DataTable/DataTable.svelte
@@ -217,15 +217,22 @@
         // Apply custom filter if provided.
         filteredRows = originalRows.filter((row) => customFilter(row, value));
       } else {
-        // Default filter checks all non-id fields for a basic, case-insensitive match (non-fuzzy).
+        // Get searchable keys from headers (non-empty headers with keys).
+        const searchableKeys = headers
+          .filter((header) => !header.empty && header.key)
+          .map((header) => header.key);
+
+        // Default filter checks fields defined in headers
+        // for a basic, case-insensitive match (non-fuzzy).
+        // This supports nested keys like "contact.company".
         filteredRows = originalRows.filter((row) => {
-          return Object.entries(row)
-            .filter(([key]) => key !== "id")
-            .some(([key, _value]) => {
-              if (typeof _value === "string" || typeof _value === "number") {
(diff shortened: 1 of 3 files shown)


metonym:

PR #2215, opened 2025-10-16
fix(slider): dispatch `on:input` for keyboard events

### LINKED ISSUE #1980: Slider on:input not called for keyboard input
When controlling the slider with cursor keys, the input event doesn't fire.

metonym:
Fixes #1980


metonym:

PR #2216, opened 2025-10-16
fix(slider): round `Shift+Arrow` values to valid steps

### LINKED ISSUE #1219: Slider component does not round value when on key [Shift]+[Arrow]
Hi. I believe some rounding logic is missed here:  
https://github.com/carbon-design-system/carbon-components-svelte/blob/ea7c0d446e989a9435788a9edacdc0542b683da0/src/Slider/Slider.svelte#L166

whereas another part has the rounding logic properly done:  
https://github.com/carbon-design-system/carbon-components-svelte/blob/ea7c0d446e989a9435788a9edacdc0542b683da0/src/Slider/Slider.svelte#L87

For instance with the [Shift] key decimal values may appear even when we have `step = 1`.

metonym:
Fixes #1219

### DIFF
--- a/src/Slider/Slider.svelte
+++ b/src/Slider/Slider.svelte
@@ -183,8 +183,9 @@
           ArrowUp: 1,
         };
         if (keys[key]) {
-          value +=
+          const delta =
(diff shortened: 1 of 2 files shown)


metonym:

PR #2217, opened 2025-10-16
fix(toggle): dispatch change event for keyboard interactions

### LINKED ISSUE #1611: [Toggle] keyboard interactions don't dispatch `change` event
## Issue

`e.preventDefault()` is called during a `keyup` event:

https://github.com/carbon-design-system/carbon-components-svelte/blob/c1c8309443386652137beedc129b06d3c9d776de/src/Toggle/Toggle.svelte#L67-L72

This prevents the `change` event from dispatching that otherwise should. Is this intentional? 

## Reproduction

Demo: https://stackblitz.com/edit/sveltejs-kit-template-default-ewqjmk?file=src%2Froutes%2F%2Bpage.svelte (inspired by https://github.com/carbon-design-system/carbon-components-svelte/discussions/1609)

- clicking on the toggle will call the `on:change` handler
- using the ENTER or SPACE keys will not call the `on:change` handler

## Proposal

I think it would be fair to forward `on:keydown` without handling it. The HTML `<input type="checkbox">` will fire `change` on click and SPACE, not ENTER.

metonym:
Fixes #1611

The Toggle component's keyboard handlers (Space and Enter keys) were preventing the native `change` event from firing due to `e.preventDefault()` being called in the `keyup` handler. This created inconsistent behavior where `on:change` listeners worked for mouse clicks but not keyboard interactions.

**Changes**
- Removed Space key from custom handler to allow native checkbox behavior
- Modified Enter key handler to call `click()` instead of manually toggling state
- Moved Enter key handling from `keyup` to `keydown` for proper preventDefault timing
- Added input element ref binding to enable programmatic click
- Updated tests to verify `change` event fires for both click and keyboard events

The `change` event now fires consistently across all interaction methods while maintaining backward compatibility with the existing `toggle` event.

### DIFF
--- a/src/Toggle/Toggle.svelte
+++ b/src/Toggle/Toggle.svelte
@@ -39,6 +39,8 @@
   import { createEventDispatcher } from "svelte";
 
   const dispatch = createEventDispatcher();
+
+  let ref = null;
 </script>
 
 <!-- svelte-ignore a11y-mouse-events-have-key-events -->
@@ -54,6 +56,7 @@
 >
   <!-- svelte-ignore a11y-role-has-required-aria-props -->
   <input
+    bind:this={ref}
     role="switch"
     type="checkbox"
     class:bx--toggle-input={true}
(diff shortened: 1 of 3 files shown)


metonym:

PR #2219, opened 2025-10-16
fix(context-menu): prevent parent menu closure when clicking submenu on mobile

### LINKED ISSUE #1848: ContextMenu parent menu closes when submenu opens
Related to #1847

[Test link](https://carbon-components-svelte.onrender.com/framed/ContextMenu/ContextMenu?theme=white)

image
image

On desktop, sub-menus can be opened by hovering on the parent menu, but on mobile, hovering is not possible, users are required to click on the parent menu to access the submenu, but currently whenever I click on a parent menu item, the whole menu closes (and the submenu opens at some weird location, described in #1847).

I expect the menu to persist until I click outside or click on an option that doesn't have a submenu.

metonym:
Fixes #1848

Fixes a mobile interaction bug where clicking a `ContextMenu` item with a submenu would close the entire menu instead of opening the submenu. This made nested menus inaccessible on touch devices.

The fix modifies `ContextMenuOption` to prevent click events from bubbling up to the window-level handler when clicking submenu triggers. Now, clicking a submenu trigger stops propagation and explicitly opens the submenu, enabling mobile users to navigate nested menus while preserving the existing desktop hover behavior.

### DIFF
--- a/src/ContextMenu/ContextMenuOption.svelte
+++ b/src/ContextMenu/ContextMenuOption.svelte
@@ -236,7 +236,14 @@
       submenuOpen = false;
     }
   }}
-  on:click={handleClick}
+  on:click={(e) => {
+    if (subOptions) {
+      e.stopPropagation();
+      submenuOpen = true;
+      return;
+    }
+    handleClick();
+  }}
 >
   {#if subOptions}
     <div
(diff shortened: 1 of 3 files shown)


metonym:

PR #2221, opened 2025-10-16
fix(context-menu): nested menu implements hover intent triangle

### LINKED ISSUE #2220: `ContextMenu` with submenu should have a safe triangle
The ContextMenu supports nested menus when hovering/clicking the item. There should a be a safe-hover triangle when moving the mouse to the submenu, so that it does not prematurely close.

---

Image

metonym:
Fixes #2220

Implements safe triangle logic to preserve the nested context menu when transferring the cursor to a submenu item. This is a fundamental interaction pattern for nested menus.

---

## Before

https://github.com/user-attachments/assets/0354cadc-3597-43ad-a35e-8a2a1a63fe8d

## After

https://github.com/user-attachments/assets/8dd4effd-5b83-41d1-8367-54059193626f

### DIFF
--- a/src/ContextMenu/ContextMenuOption.svelte
+++ b/src/ContextMenu/ContextMenuOption.svelte
@@ -60,10 +60,12 @@
 
   // "moderate-01" duration (ms) from Carbon motion recommended for small expansion, short distance movements
   const moderate01 = 150;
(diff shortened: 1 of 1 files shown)


metonym:

PR #2223, opened 2025-10-16
fix(file-uploader): `multiple` should append instead of replace files

### LINKED ISSUE #1774: FileUploader: Files are replaced rather than added
When `multiple` is set, the existing file list is cleared when new files are selected.
This should not happen, newly selected files should be added to the existing list of files.

See also the [demo on the spec page](https://carbondesignsystem.com/components/file-uploader/usage/#live-demo).
(There may be a bug in that version as well: When adding the same file multiple times, it creates multiple entries, but removing one of them removes all instances. Not sure why, maybe the browser tracks file identity and if the items are removed via `filter` all references get removed. Ideally existing files should not be added multiple times in the first place.)

metonym:
Fixes #1774

### DIFF
--- a/src/FileUploader/FileUploaderButton.svelte
+++ b/src/FileUploader/FileUploaderButton.svelte
@@ -110,7 +110,7 @@
   class:bx--visually-hidden={true}
   {...$$restProps}
   on:change|stopPropagation={({ target }) => {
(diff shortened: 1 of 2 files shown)


metonym:

Issue #1785, opened 2023-07-23
FileUploader: `files` not fully synchronized with underlying input


brunnerh:
In the `"edit"` state, files can be removed, but the invisible file `input` will still contain them, so with a regular form post, they are still submitted. Programmatically changing the `files` property will also have no effect on the input. `FileUploaderDropContainer` should have the same issue.

[REPL demo](https://svelte.dev/repl/8c19d3e6d55f4a6eb91467794a4e8360?version=4.1.1)

A note on a potential fix approach:
The [`input.files`](https://html.spec.whatwg.org/multipage/input.html#dom-input-files) property can be set, but `FileList` objects cannot easily be constructed (at least last time I checked). A common way around this is to use a `DataTransfer`.

```js
const data = new DataTransfer(); 
data.items.add(file);
input.files = data.files;
```

Related/overlap with:
- #1774

metonym:
Fixed in [v0.89.8](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.89.8)

PR #2224, opened 2025-10-16
fix(file-uploader): synchronize `input.files` when files removed

metonym:
Fixes [#1785](https://github.com/carbon-design-system/carbon-components-svelte/issues/1785)

Previously, when users removed files by clicking the close button, the files were only removed from the visual display but remained in the input element's `FileList`. This caused the removed files to still be submitted with forms, resulting in a data integrity issue.

The fix adds a reactive statement in `FileUploaderButton` that uses the DataTransfer API to rebuild the input's FileList whenever the files array changes. The `FileUploader` component now binds the files prop to `FileUploaderButton` to ensure updates propagate correctly.

### DIFF
--- a/src/FileUploader/FileUploader.svelte
+++ b/src/FileUploader/FileUploader.svelte
@@ -146,6 +146,7 @@
     {multiple}
     {kind}
     {size}
+    bind:files
     on:change
     on:change={(e) => {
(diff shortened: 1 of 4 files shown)


metonym:

PR #2227, opened 2025-10-16
fix(context-menu): avoid clipping nested menu on narrow screens

### LINKED ISSUE #1847: ContextMenu submenu position goes out of screen on small screen
This issue might relate to #577

I faced this issue while testing the website on mobile.
Tested again on browser responsive mode with the example on the docs: [link](https://carbon-components-svelte.onrender.com/framed/ContextMenu/ContextMenu?theme=white)

image
image

The submenu position should be constrained inside the viewport.
This is the positioning behavior on Carbon React for reference: 

image

metonym:
Fixes #1847

This fixes submenu positioning on narrow viewports by improving the horizontal constraint logic in `ContextMenuOption.svelte`. 

Currently, the submenu overflows the viewport on small screens because it only checks if there's space on the right side using an incorrect reference point (`menuOffsetX`). The fix now uses the actual submenu width to check both right and left viewport boundaries, falling back to the viewport edge when the submenu would overflow on both sides.

This fix is consistent with the React implementation.

---

## Before

Screenshot 2025-10-16 at 2 53 01 PM

## After

Screenshot 2025-10-16 at 2 53 34 PM

### DIFF
--- a/src/ContextMenu/ContextMenuOption.svelte
+++ b/src/ContextMenu/ContextMenuOption.svelte
@@ -74,6 +74,7 @@
   let submenuPosition = [0, 0];
   let menuOffsetX = 0;
   let mousePosition = { x: 0, y: 0 };
+  /** @type {HTMLUListElement | null} */
   let submenuRef = null;
 
   const unsubPosition = ctx.position.subscribe((position) => {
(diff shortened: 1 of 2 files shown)


metonym:

PR #2229, opened 2025-10-17
fix(date-picker): dispatch `change` for "single" variant

### LINKED ISSUE #314: DatePicker does not dispatch change event after changing the input when type is single
If the DatePicker type is `single` and the calendar is shown, the `change` event is not triggered after I manually input the date and input is blurred.

The event is triggered if I click on a date from the calendar and also if I input the date when the DatePicker type is `simple`.

### LINKED ISSUE #950: DatePicker does not dispatch change event if value is manually cleared or force cleared
This applies to `datePickerType="single"`, probably to `"range"` as well. In this case the `change` event is fired indirectly via the `onChange` event of flatpickr.

Steps to reproduce:

1. Add `DatePicker` with...
    1. A `maxDate` multiple years in the past (e.g. 20 years), this can cause a forced clear if a partial date is typed.
    2. `dateFormat="Y-m-d"`
2. Add `on:change` listener that logs `dateStr`/processes date.
3. Select a date with the calendar popup.
4. Clear the value either...
    1. manually via the keyboard (delete everything)
    2. via a forced clear. e.g. reduce `2000-01-31` to just `200` and click/tab away. (If it only gets reduced to something like `2000-`, the value gets completed, in this case to `2000-01-01`. This fires a change event.)

[REPL](https://svelte.dev/repl/5eafc60744e7406bbccb8cc45d6705d1?version=3.44.3) (Use `With max in past` for the forced clear case)

This REPL shows one of the use cases for the event: Creating a `value` input/`change` event output loop to create a component that has a `Date`-typed prop. `value` alone cannot be bound and reactively set at the same time as that would constitute a circular dependency.

---

Flatpickr has an event `onValueUpdate` that probably could be considered, as `onChange` does not appear to fire on clear.

metonym:
Fixes #314, fixes #950

Fixes `DatePicker` not dispatching `change` events when manually typing or clearing dates in `single`/`range` modes (#314, #950).

The bug was caused by a condition that blocked events when a Flatpickr calendar existed, assuming Flatpickr would handle all cases, but it only fires for calendar selections, not manual input. This fix removes that blocking condition, ensuring `change` events are dispatched consistently for all input methods while maintaining backward compatibility. Includes unit tests for manual typing and clearing.

### DIFF
--- a/src/DatePicker/DatePicker.svelte
+++ b/src/DatePicker/DatePicker.svelte
@@ -116,8 +116,23 @@
         inputValue.set(value);
       }
 
-      if (!calendar && type === "change") {
-        dispatch("change", value);
+      if (type === "change") {
+        if (calendar) {
+          const detail = { selectedDates: calendar.selectedDates || [] };
+
+          if ($range) {
+            detail.dateStr = {
+              from: inputRef?.value || "",
+              to: inputRefTo?.value || "",
+            };
+          } else {
+            detail.dateStr = inputRef?.value || "";
+          }
+
+          dispatch("change", detail);
+        } else {
+          dispatch("change", value);
+        }
       }
     },
     blurInput: (relatedTarget) => {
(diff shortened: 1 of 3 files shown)


metonym:

PR #2230, opened 2025-10-17
fix(text-area)!: disallow `resize` if `cols` is specified

### LINKED ISSUE #1661: Disallow `TextArea` resize if `cols` is specified
A recent fix to the upstream Carbon library caught my eye: https://github.com/carbon-design-system/carbon/pull/13069

It disables the "resize" attribute for `TextArea` if the `cols` prop is specified.

The differences in behavior compared to the Svelte implementation:

- `cols` does not have a default value (the Svelte implementation defaults to `50`)
- if `cols` is not specified, the textarea has `width=100%`
- if `cols` is specified, the textarea unsets `width` and applies the style rule `resize=none`

If we want to align with the upstream library, this would be a breaking change:

```diff
- export let cols = 50;
+ export let cols = undefined;

+ <textarea style:resize={typeof cols === "number" ? "none" : undefined} />
```

metonym:
Fixes #1661

This aligns `TextArea` with the upstream React implementation by removing the default `cols` value and automatically disabling textarea resize when `cols` is explicitly specified. Previously, the component had a default `cols={50}` and did not disable resizing. Now, when `cols` is not specified, no `cols` attribute is rendered and the textarea remains resizable. When `cols` is explicitly set to a number, `resize: none` is automatically applied to match upstream behavior.

This can be overridden by specifying a custom `resize` styel attribute.

**Breaking Change:** The `cols` prop no longer defaults to `50`. `Textarea` will now use the browser's default width unless `cols` is explicitly specified.


metonym:

PR #2234, opened 2025-10-18
fix(text-input): preserve HTML5 validation for `type="number"`

### LINKED ISSUE #1836: TextInput `number` type is not compatible with HTML `step` attribute
The `onInput` and `onChange` event handlers inside `TextInput` component call a `parse` function which converts the string value into a `Number` before being dispatched.
```
  function parse(raw) {
    if ($$restProps.type !== "number") return raw;
    return raw != "" ? Number(raw) : null;
  }

  /** @type {(e: Event) => void} */
  const onInput = (e) => {
    value = parse(e.target.value);
    dispatch("input", value);
  };

  /** @type {(e: Event) => void} */
  const onChange = (e) => {
  };
```

This behaviour breaks the [step](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step) and [pattern](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern) attributes from being used along with `number` inputs.

Suggested solution: Remove the `parse` function.

metonym:
Fixes #1836

This preserves HTML5 validation attributes (`step`, `min`, `max`) for `type="number"` text inputs. The component now maintains an internal string value bound to the DOM input element, allowing native browser validation to work correctly, while the exported `value` prop continues to provide typed numbers (or `null` when empty) for backward compatibility.

This means users can now use `<TextInput type="number" step="0.01" />` and have the browser's native step validation work as expected, while `bind:value` and event handlers continue to behave exactly as before. All existing tests pass with no breaking changes.

### DIFF
--- a/src/TextInput/TextInput.svelte
+++ b/src/TextInput/TextInput.svelte
@@ -77,13 +77,24 @@
   const ctx = getContext("Form");
   const dispatch = createEventDispatcher();
 
+  // Internal string value for the input element (preserves HTML5 validation)
+  let inputValue = "";
+
+  // Sync inputValue with value prop
+  $: if ($$restProps.type === "number") {
+    inputValue = value == null ? "" : String(value);
+  } else {
+    inputValue = value == null ? "" : value;
+  }
+
   function parse(raw) {
-    if ($$restProps.type !== "number") return raw;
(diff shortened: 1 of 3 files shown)


metonym:

PR #2236, opened 2025-10-18
fix(date-picker): allow clearing range dates by setting values to empty strings

### LINKED ISSUE #1862: Can't unset from and to dates on DatePicker range
Hi all,

Using the DatePicker in range mode, I want to be able to unset the values when the user clicks a button (which clears an entire form of page filters).

Unfortunately doing the obvious thing of setting valueFrom and valueTo to "" doesn't work - what happens is that they both end up being set to whatever the valueTo was set to.

After a bit of digging around in the code, it looks like this is happening in the afterUpdate function in DatePicker.svelte. Here:

```
        // workaround to remove the default range plugin separator "to"
        inputRef.value = $inputValueFrom;
```

This seems like a valid workaround to prevent the From box showing "<date> to <date>" when there are some values set. However it seems to break things if there is no value set.

My workaround for the workaround is simply to only do the workaround if the value isn't "", like this:

```
        // workaround to remove the default range plugin separator "to"
        if ($inputValueFrom != "") {
            inputRef.value = $inputValueFrom;
        }
```

I don't know if this is a sensible/correct fix, but it seems to work for me!

Is there any chance that this fix could be incorporated into the official code?

Thanks!

Joe

metonym:
Fixes #1862

Addresses an issue where setting valueFrom and valueTo to "" would not clear the input fields in range mode due to unconditional value assignment in the `afterUpdate` method. Incorporates the valid fix suggested by @JOldak in #1862.

It now only applies the range separator workaround when a value exists.

### DIFF
--- a/src/DatePicker/DatePicker.svelte
+++ b/src/DatePicker/DatePicker.svelte
@@ -217,7 +217,9 @@
         calendar.setDate([$inputValueFrom, $inputValueTo]);
 
         // workaround to remove the default range plugin separator "to"
-        inputRef.value = $inputValueFrom;
+        if ($inputValueFrom !== "") {
+          inputRef.value = $inputValueFrom;
+        }
       } else {
         calendar.setDate($inputValue);
       }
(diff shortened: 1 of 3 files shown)


metonym:

Issue #1544, opened 2022-10-28
Readonly checkbox will still allow to be edited


clouedoc:
Hello,

First off, thanks for the good code.
I've tried a few different component libraries before Carbon, and it's the best one so far.

My problem is that I cannot find a way to have a read-only checkbox.

## Why do you want a read-only checkbox?

I want to create a checkbox that can only be enabled via JavaScript.
The reason is that the user should do an action outside the page.
The checkbox serves as a reminder (you need to do X before clicking "submit" - the submit button being disabled)

## What did you try?

I have tried the following code, unsuccessfully:

```svelte
<Checkbox
  required
  checked={receivedSessionData}
  readonly
  on:change={() => {}}
  on:check={() => {}}
  on:click={() => {}}
  labelText="I have ran the session extraction script (will update automatically)"
/>
```

Note that both of these methods are ineffective:

1. Adding the `readonly` property
2. Overriding `on:change`, `on:check` or `on:click` with a NOOP

## Alternative solution

Setting `disabled` on the checkbox works, but grays out the text and makes it harder to read for the user.
I have tried overriding CSS styles with `color: black !important;` unsuccessfully.

iJustErikk:
I don't think native checkboxes support readonly (at least in Chrome). However, I think Svelte Carbon should support them.

I'd be down to put in a PR with readonly checkboxes with this behavior:

If readonly is set, then the cursor should be a regular cursor and clicking does nothing.

I'll put the pr in tonight or tomorrow

brunnerh:
Natively that is not a thing. It would have to be supported by aria attributes, otherwise screen reader users would just be confused if nothing changes on interaction.

Also, things have to be styled with styles from the flagship Carbon package, it looks like there is a read-only state [in the spec](https://carbondesignsystem.com/components/checkbox/usage/#states), but the flagship is on v11, this is still on v10; not sure if the state would exist there.

brunnerh:
> Seems like they are inconsistent between readonly and disabled.

They are supposed to, they are not the same after all, if you want `disabled`, just use that.  
There are example images in the spec I linked to.

iJustErikk:
I was referring to how they select within that file

```
.#{$prefix}--checkbox:disabled + .#{$prefix}--checkbox-label,
  .#{$prefix}--checkbox-label[data-contained-checkbox-disabled='true'] {
    color: $text-disabled;
    cursor: not-allowed;
  }
```

vs 

```
[...]

PR #2237, opened 2025-10-18
fix(checkbox): prevent state changes when `readonly` is true

metonym:
Fixes https://github.com/carbon-design-system/carbon-components-svelte/issues/1544

This PR fixes the `readonly` prop for Checkbox, which was declared but non-functional. Previously, clicking a read-only checkbox would still toggle its state. This PR prevents user interaction by adding `preventDefault()` to the click and change handlers while allowing programmatic control and adding the `aria-readonly` attribute for accessibility.

This aligns the Svelte implementation with the upstream implementation.

For styling, we need to polyfill the upstream [readonly checkbox SCSS.](https://github.com/carbon-design-system/carbon/blob/8d765233aa44c655de9e2917d7ec07f93816bfd5/packages/styles/scss/components/checkbox/_checkbox.scss)

### DIFF
--- a/docs/src/pages/components/Checkbox.svx
+++ b/docs/src/pages/components/Checkbox.svx
@@ -37,6 +37,13 @@ Set `disabled` to `true` to prevent user interaction.
 
 <Checkbox labelText="Label text" disabled />
 
+## Readonly state
+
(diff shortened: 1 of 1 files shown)


metonym:

PR #2240, opened 2025-10-18
fix(tooltip): prevent closing on layout shift

### LINKED ISSUE #925: Tooltip can close immediately after opening when focus causes a layout change
If a Tooltip with an auto-focused element is positioned e.g. at the bottom edge of the document, clicking it can cause the button that opens the tooltip to move away from the cursor. This then causes the Tooltip to close again immediately.

Could be a browser bug. Tested on Windows 10 in Chrome and Firefox. Chrome had the issue, Firefox did not.

[Repro REPL](https://svelte.dev/repl/1f9dd6c9c8a44047b80b0debc90bc3d5?version=3.44.2)

---

In general, button clicks should trigger when releasing the mouse.
Is there a good reason why it uses `mousedown` rather than a `click` here? I suspect it has to do with the focus...

Keeping the mouse-down behavior, switching to pointer events and using pointer capture would help here; could possibly have unintended side effects, though:

```diff
- function onMousedown() {
+ function onPointerdown(e) {
+   e.currentTarget.setPointerCapture(e.pointerId);

    // determine the desired state before the focus event triggers.
    const shouldClose = open;
    // ensure changes are scheduled at the end, i.e. after the possible focus event.
    setTimeout(() => {
      open = shouldClose ? false : true;
    });
  }

+ function onPointerup(e) {
+   e.currentTarget.releasePointerCapture(e.pointerId);
+ }

...

  {#if !hideIcon}
    <div bind:this="{ref}" id="{triggerId}" class:bx--tooltip__label="{true}">
      <slot name="triggerText">{triggerText}</slot>
      <div
        bind:this="{refIcon}"
        {...buttonProps}
        aria-describedby="{tooltipId}"
-       on:mousedown="{onMousedown}"
+       on:pointerdown="{onPointerdown}"
+       on:pointerup="{onPointerup}"
        on:focus="{onFocus}"
        on:keydown="{onKeydown}"
      >
        <slot name="icon">
          <svelte:component this="{icon}" name="{iconName}" />
        </slot>
      </div>
    </div>
  {:else}
    <div
      bind:this="{ref}"
      {...buttonProps}
      aria-describedby="{tooltipId}"
-     on:mousedown="{onMousedown}"
+     on:pointerdown="{onPointerdown}"
+     on:pointerup="{onPointerup}"
      on:focus="{onFocus}"
      on:blur="{onBlur}"
      on:keydown="{onKeydown}"
    >
      <slot name="triggerText">{triggerText}</slot>
    </div>
  {/if}
```

metonym:
Fixes #925

Fixes an issue where tooltips would close immediately after opening when positioned near the document edge. The problem occurred because focus events could cause layout changes that moved the trigger button away from the cursor, causing the click handler to interpret this as clicking outside the tooltip. This change replaces mousedown/mouseup events with pointerdown/pointerup and uses setPointerCapture() (suggested by @brunnerh) to ensure all pointer events remain captured to the trigger element even if it moves during interaction, preventing premature closure.

Crediting Git commit co-authorship to @brunnerh.

### DIFF
--- a/src/Tooltip/Tooltip.svelte
+++ b/src/Tooltip/Tooltip.svelte
@@ -99,13 +99,22 @@
     open = true;
   }
 
-  function onMousedown() {
+  function onPointerdown(e) {
+    if (e.currentTarget.setPointerCapture) {
+      e.currentTarget.setPointerCapture(e.pointerId);
+    }
     // determine the desired state before the focus event triggers.
     const shouldClose = open;
     // ensure changes are scheduled at the end, i.e. after the possible focus event.
     setTimeout(() => {
       open = shouldClose ? false : true;
     });
+  }
+
+  function onPointerup(e) {
+    if (e.currentTarget.releasePointerCapture) {
+      e.currentTarget.releasePointerCapture(e.pointerId);
+    }
   }
 
   afterUpdate(() => {
@@ -212,7 +221,8 @@
         bind:this={refIcon}
         {...buttonProps}
         aria-describedby={tooltipId}
-        on:mousedown={onMousedown}
+        on:pointerdown={onPointerdown}
+        on:pointerup={onPointerup}
         on:focus={onFocus}
         on:keydown={onKeydown}
       >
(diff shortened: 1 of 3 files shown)


metonym:

PR #2242, opened 2025-10-18
fix(ui-shell): prevent focus on hidden links when `SideNav` is collapsed

### LINKED ISSUE #604: SideNav links are focusable while hidden
You can tab to the `SideNav` links even when the nav is not visible.
Would recommend setting `visibility: hidden;` in that state to prevent this.

metonym:
Fixes #604

Fixes an accessibility issue where `SideNav` links remained keyboard-focusable even when the navigation was visually hidden. This created a confusing experience for keyboard users who could tab to invisible elements.

Assigning Git commit co-authorship to @brunnerh 

- All existing tests pass
- Added 3 new unit tests covering:
  - Collapsed state applies `visibility: hidden`
  - Open state does not apply `visibility: hidden`
  - Rail mode when closed does not apply `visibility: hidden`

### DIFF
--- a/src/UIShell/SideNav.svelte
+++ b/src/UIShell/SideNav.svelte
@@ -78,6 +78,7 @@
     : isOpen}
   class:bx--side-nav--collapsed={!isOpen && !rail}
   class:bx--side-nav--rail={rail}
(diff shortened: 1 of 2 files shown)


metonym:

Issue #617, opened 2021-04-29
Combobox not support async


pakorn-tp:
Please implement combobox support autocomplete with async.

theetrain:
Hello, to check my understanding, are you looking to have Combobox query results from a remote server and display them after a debounce period?

yashsavla9:
@theetrain is this feature planned in any upcoming release ? - I'm looking for the same functionality you mentioned above
"looking to have Combobox query results from a remote server and display them after a debounce period"

metonym:
Released in [v0.90.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.90.0)

Documentation: https://svelte.carbondesignsystem.com/components/ComboBox#filtering-async

PR #2254, opened 2025-10-19
fix(combo-box): forward `on:input` to enable async filtering

metonym:
Fixes [#617](https://github.com/carbon-design-system/carbon-components-svelte/issues/617)

`ComboBox` needs to forward the `on:input` event so the consumer can asynchronously update `items` based on its value (in the case where debouncing the input is a must).

This also adds a minimal example illustrating async usage.


metonym:

Issue #1594, opened 2022-12-20
[DataTable] add control to `title`'s rendered heading level


theetrain:
By default `<DataTable title="My title">` renders a `<h4>My title</h4>`. For its title.

Let's give the user control over which heading level renders in the DOM, while retaining font size.

theetrain:
I guess this is already controllable with the `title` slot provided by DataTable. 🤔 

If that's adequate, this can be closed.

metonym:
This is a valid point, since the `title` slot is still descendent of the `h4` tag itself.

This reminds me: I think as a separate effort we should phase out the `as` props and leverage Svelte's built-in `svelte:element` API now that it's been supported for some time now.

theetrain:
How does this sound:

- By default, `title` will insert a provided string to the title area, which will render as `<div id={generatedId><p>{title}</p></div>`, and the corresponding table will reference it as `<table aria-describedby={generatedId}>`. [Reference: WAI-ARIA table captions](https://www.w3.org/WAI/tutorials/tables/caption-summary/#using-aria-describedby-to-provide-a-table-summary).
- An optional `titleTag` prop will wrap the `{title}` string with a heading tag. We could provide a `<Heading>` component as well to facilitate this
- The named slot `title` will replace the contents of the title area completely. We'll provide documented recipes to make use of `<Heading>` or the appropriate v10/v11 Carbon tokens for H4 text size.

Demo: https://svelte.dev/repl/5afabba5287048f1b53dbba481d1a4ed?version=3.55.0

metonym:
Fixed in [v0.89.8](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.89.8)

PR #2258, opened 2025-10-19
fix(data-table): title/description slots should override elements

metonym:
Fixes [#1594](https://github.com/carbon-design-system/carbon-components-svelte/issues/1594)

The slottable title and description should override the elements.


metonym:

PR #2264, opened 2025-10-19
feat(data-table)!: include `target` and `currentTarget` in row/cell click events

### LINKED ISSUE #1904: DataTable on:click:row is missing target and currentTarget information
I'm not sure if this is a bug or by design, but it's currently not possible to discern which element the user clicked on when you use on:click:row with DataTables.

My use case: I'm hoping to execute a function only when the user clicked on empty space in a row. If they click on buttons or other elements I have embedded in the table, I'd like to go with the default behavior. But since currentTarget/target don't get set, it's not possible.

metonym:
Closes #1904

The click:row, click:cell, and click:header events now include target and currentTarget properties in addition to their existing data. This allows the consumer to determine which specific element was clicked and implement conditional logic based on the click target.

**Breaking change**: Event detail structure has changed from a single value to an object:
- click:row: Row -> `{ row, target, currentTarget }`
- click:cell: DataTableCell -> `{ cell, target, currentTarget }`
- click:header: maintains `header` and `sortDirection`, adds target and currentTarget

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -1014,18 +1014,18 @@ export interface DataTableCell<Row = DataTableRow> {
 
 ### Events
 
-| Event name           | Type       | Detail                                                                                                       |
-| :------------------- | :--------- | :----------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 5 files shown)


metonym:

PR #2269, opened 2025-10-19
feat(tree-view): add `filterTreeNodes` utilities

### LINKED ISSUE #2281: `TreeView` has filtering utilities
Basic treeview functionality should support filtering nodes. For example, searching a node by its name should filter out non-matching nodes and expand the path to the match child node(s). Same thing for ID.

These utilities should be exported and used separately from `TreeView`, to avoid penalizing the baseline component.

Draft: #2269

metonym:
Closes #2281

Add tree node filtering utilities for `TreeView`. Three new functions enable filtering hierarchical tree structures by predicate, ID, or text search.

- `filterTreeNodes`: generic predicate-based filtering
- `filterTreeById`: filter by node ID(s)
- `filterTreeByText`: case-insensitive text search

The `filterTreeNodes` function accepts a predicate function to filter nodes based on custom logic. It returns a new tree containing only matching nodes and their ancestors by default.

```javascript
import { filterTreeNodes } from "carbon-components-svelte";

const tree = [
  { id: 1, text: "Documents", nodes: [
    { id: 2, text: "Work", nodes: [
      { id: 3, text: "Report.docx" }
    ]}
  ]}
];

const filtered = filterTreeNodes(tree, (node) => node.id === 3);
// Returns: [{ id: 1, text: "Documents", nodes: [{ id: 2, text: "Work", nodes: [{ id: 3, text: "Report.docx" }] }] }]
```

The `filterTreeById` function filters nodes by ID, accepting either a single ID or an array of IDs.

```javascript
import { filterTreeById } from "carbon-components-svelte";

const filtered = filterTreeById(tree, 3);
// Or filter by multiple IDs:
const filtered = filterTreeById(tree, [3, 4]);
```

The `filterTreeByText` function performs case-insensitive substring matching on node text properties.

```javascript
import { filterTreeByText } from "carbon-components-svelte";

const filtered = filterTreeByText(tree, "work");
// Matches any node where text contains "work" (case-insensitive)
```

All filter functions accept an optional options object with two properties. The `includeAncestors` option (default: true) includes parent nodes of matching nodes, ensuring the filtered tree maintains valid parent-child relationships. The `includeChildren` option (default: false) includes all descendants of matching nodes.

```javascript
// Include all children of matching nodes
const filtered = filterTreeByText(tree, "Documents", { includeChildren: true });

// Exclude ancestors, only return matching nodes
const filtered = filterTreeNodes(tree, (node) => node.id === 3, { includeAncestors: false });
```

### DIFF
--- a/docs/src/pages/components/TreeView.svx
+++ b/docs/src/pages/components/TreeView.svx
@@ -113,3 +113,21 @@ Use `TreeView.showNode` to expand, select, and focus a specific node.
 Convert flat data to a hierarchical structure using the `toHierarchy` utility.
 
 <FileSource src="/framed/TreeView/TreeViewFlatArray" />
+
+## Filter by text with search input
+
+Combine a `Search` input with `filterTreeByText` to create an interactive searchable tree. Type to filter nodes by name (case-insensitive substring matching). By default, matching nodes and their ancestors are included to maintain valid tree structure.
+
+<FileSource src="/framed/TreeView/TreeViewFilter" />
+
+## Filter by ID
+
+Use `filterTreeById` to filter nodes by their ID. Accepts either a single ID or an array of IDs. Click the buttons to filter by a single ID or multiple IDs. Matching nodes and their ancestors are included by default.
+
+<FileSource src="/framed/TreeView/TreeViewFilterById" />
+
+## Filter with custom predicate
+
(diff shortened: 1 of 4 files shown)


metonym:

PR #2290, opened 2025-10-26
Revert "fix(text-input): preserve HTML5 validation for `type="number"`"

### LINKED ISSUE #2289: TextInput: 0.89.8 and 0.90.0 Break TextInput Component
I have a login form with two inputs: email and password. On all versions up to 0.89.7, the form worked fine. On 0.89.8 and 0.90.0, the email field clears itself out when I submit the form. When I attempt to type into the field, it will not accept any input. 

I've tested this with a new, blank field that has no bindings or preset values, and the same occurs. This happens in all browsers, on all operating systems. Reverting back to 0.89.7 fixes the issue.

To add to this, even the TextInput examples on the Carbon Components Svelte website will not accept any keyboard input:  https://carbon-components-svelte.onrender.com/components/TextInput

metonym:
This reverts commit 936677da26df6c39a91b5101b3e52319d2cdd898.

Fixes #2289 

#2234 caused a regression to `TextInput`, affecting 0.89 and 0.90.

Hotfix is to revert the change until root cause is determined and a forward fix is implemented. Will backport to 0.89.

### DIFF
--- a/src/TextInput/TextInput.svelte
+++ b/src/TextInput/TextInput.svelte
@@ -77,24 +77,13 @@
   const ctx = getContext("Form");
   const dispatch = createEventDispatcher();
 
-  // Internal string value for the input element (preserves HTML5 validation)
-  let inputValue = "";
-
-  // Sync inputValue with value prop
-  $: if ($$restProps.type === "number") {
-    inputValue = value == null ? "" : String(value);
-  } else {
-    inputValue = value == null ? "" : value;
-  }
-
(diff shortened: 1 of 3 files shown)


metonym:
Fixed in:

- [v0.89.9](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.89.9)
- [v0.90.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.90.1)

PR #2317, opened 2025-11-08
fix(multi-select): restore complete keyboard navigation for filterable variant

### LINKED ISSUE #2313: MultiSelect won't open on focus since v0.89.0
Problem:
When you use the tab to focus a (filterable) MultiSelect field it won't open the menu since v0.89.0.
It does open when you mouse click on the field. 
Additionally: 
- if you tab into the field it won't accept input
- if there is already an item selected (you see "1X" indication in the field), tabbing into the field will focus the X and not the field

To reproduce:
- Visit the components page: [https://carbon-components-svelte.onrender.com/components/MultiSelect](https://carbon-components-svelte.onrender.com/components/MultiSelect) 
- Go to Filterable example
- Click above on the label (twice, first time to focus (the multiselect opens), second time to close the multiselect.
- Now Tab, you cursor moves into the field but the MultiSelect does not open (problem 1)
- Try typing something, it won't accept input (problem 2) 
- Click in the menu to open it, select Email
- Try typing "Fa", it will show the Fax option (because it got focus through the click)
- Click the close arrow to close the menu
- Click above the field to make MultiSelect loose focus
- Tab, now you will see that the X gets focus and not the field (problem 3).

I don't have a REPL but i went back in version and v0.88.4 doesn't have the problem.

Expected behaviour
- open the multiselect field when it get focus by tab.
- field accepts input after focus through tab
- tab focuses the field, not the X

metonym:
Follow-up to #2314, based on https://github.com/carbon-design-system/carbon-components-svelte/issues/2313#issuecomment-3506765505

Fixes #2313 

Fixes three keyboard navigation issues in filterable MultiSelect:

1. Menu now opens immediately when field receives focus via Tab, preventing the need to click or type first
2. First character typed after Tab focus now displays correctly
3. Clear button removed from tab order (`tabindex="-1"`) so Tab goes directly to input field instead of stopping at the X button

### DIFF
--- a/src/ListBox/ListBoxSelection.svelte
+++ b/src/ListBox/ListBoxSelection.svelte
@@ -67,7 +67,7 @@
     <div
       bind:this={ref}
       role="button"
-      tabindex={disabled ? -1 : 0}
+      tabindex="-1"
       class:bx--tag__close-icon={true}
       on:click|preventDefault|stopPropagation={(e) => {
         if (!disabled) {
@@ -92,7 +92,7 @@
     role="button"
     aria-label={description}
     title={description}
-    tabindex={disabled ? "-1" : "0"}
+    tabindex="-1"
     class:bx--list-box__selection={true}
     class:bx--tag--filter={selectionCount}
     class:bx--list-box__selection--multi={selectionCount}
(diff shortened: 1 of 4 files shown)


metonym:

PR #2318, opened 2025-11-08
perf(data-table): hoist regex for nested keys

metonym:
Adds stricter Biome performance rules. Optimizes scripts where regular expressions can be hoisted.

Optimizes `DataTable` by hoisting the regex used for nested keys.


metonym:

PR #2320, opened 2025-11-08
fix(slider): store event object for drag calculations

metonym:
Blocks #2319

Fixes a TypeError when dragging by storing the event object in the `move()` handler instead of relying on an undefined global event variable.


metonym:

PR #2322, opened 2025-11-08
feat(types): annotate typedefs and dispatched events

metonym:
sveld@0.23 supports `@property` notation in typedefs, allowing JSDoc descriptions to be added to typedefs and events.

This improves developer experience for IDEs with hover definition support.

---

Screenshot 2025-11-08 at 3 20 38 PM


metonym:

PR #2325, opened 2025-11-09
fix(types): correctly generate `@default` values

metonym:
Upgrade to [sveld@0.23.3](https://github.com/carbon-design-system/sveld/releases/tag/v0.23.3) so that literal values are correctly used in the `@default` JSDoc annotations in generated TypeScript definitions.


metonym:

PR #2330, opened 2025-11-09
fix(breakpoint): use `declare const` in types

metonym:
Breakpoint types were manually authored; the const should include the `declare` keyword.


metonym:

Issue #2282, opened 2025-10-20
`TreeView` supports showing a node without selecting it


metonym:
`TreeView` currently supports "showing" a node.

However, it "selects" it by default. I would like to show the node *without* selecting it.

- Should selection be the default?

bhavers:
Great function that i could use as well. Maybe this would be a good default:
- when clicking the twisty it expands the node but does not select it
- when clicking the label it exoands the node and selects is.

Additionally, maybe an option to leave other branches expanded when selecting a node. Currently showNode() expands up to the selected node and closes what was already open.

Here's what i use for that:
```
/** Expand the selected node, and leaves previously opened nodes open */
	function showMultipleNodes() {
		if (items.selectedIndex && items.currentItem !== undefined) {
			addParents(items.currentItem);
			expandedIds = [...new Set(expandedIds)]; // remove duplicates

			function addParents(item: Item) {
				const parent = items.list[items.getIndex(item.parentId)];
				if (parent) {
					// console.log(`Parent of ${item.name} is ${parent.name}`);
					expandedIds = [...expandedIds, parent.uuid];
					addParents(parent);
				}
			}
		}
		refToTree?.expandNodes((node) => {
			return expandedIds.includes(node.id);
		});
	}
```

metonym:
Thanks for the feedback!

metonym:
Released in [v0.91.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.91.0)

Docs: https://svelte.carbondesignsystem.com/components/TreeView#show-node-with-options

PR #2332, opened 2025-11-09
feat(tree-view): `showNode` is configurable

metonym:
Closes [#2282](https://github.com/carbon-design-system/carbon-components-svelte/issues/2282), closes #2206

Currently, `TreeView` supports an imperative `showNode` method. However, it currently will select, focus, and expand the node.

This PR adds configurable options to the `showNode` method, allowing independent control over expand, select, and focus behaviors. Previously, calling `showNode(id)` would always expand, select, and focus the target node. This change enables use cases like expanding without selecting (issue #2282) or selecting without expanding (PR #2206).

This is backwards compatible, as the method will still select/focus/expand the node by default.

**API Changes**

The `showNode` method now accepts an optional second parameter:

```typescript
showNode(id: TreeNodeId, options?: ShowNodeOptions): void

interface ShowNodeOptions {
  expand?: boolean;  // default: true
  select?: boolean;  // default: true
  focus?: boolean;   // default: true
}
```

**Usage Examples**

```javascript
// Default behavior (backwards compatible)
treeView.showNode("node-1");

// Expand ancestors without selecting the node
treeView.showNode("node-2", { select: false, focus: false });

// Select node without expanding ancestors
treeView.showNode("node-3", { expand: false, focus: false });

// Only update keyboard focus
treeView.showNode("node-4", { expand: false, select: false });
```

https://github.com/user-attachments/assets/966d28a5-a0ed-4cf8-9a7d-e28a08c50343

### DIFF
--- a/docs/src/pages/components/TreeView.svx
+++ b/docs/src/pages/components/TreeView.svx
@@ -108,6 +108,12 @@ Use `TreeView.showNode` to expand, select, and focus a specific node.
 
 <FileSource src="/framed/TreeView/TreeViewShowNode" />
 
+## Show node with options
+
+Configure `TreeView.showNode` behavior using optional parameters. By default, all three actions occur (`expand`, `select`, `focus`), but you can disable any combination. This is useful for expanding nodes without selecting them, or selecting without expanding.
+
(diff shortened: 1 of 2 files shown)


metonym:

PR #2333, opened 2025-11-09
fix(tree-view-node): correctly type `computeTreeLeafDepth` export

metonym:
Upgrades to [svelte@0.24.2](https://github.com/carbon-design-system/sveld/releases/tag/v0.24.2) so that `@params` are correctly parsed in JSDocs, and included in the generated types.


metonym:

Issue #534, opened 2021-02-24
Inconsistent label property names?


brunnerh:
- Checkbox: `labelText`
- ComboBox: `titleText`
- ComboBox: `titleText`
- DatePickerInput: `labelText`
- Dropdown: `titleText`
- FileUploaderButton: `labelText`
- MultiSelect: `titleText`
- NumberInput: `label`
- PasswordInput: `labelText`
- ProgressStep: `label`
- RadioButton: `labelText`
- Select: `labelText`
- SelectItem: `text`
- Slider: `labelText`
- Tab: `label`
- TextArea: `labelText`
- TextInput: `labelText`
- TimePicker: `labelText`
- Toggle: `labelText`

(List might not be complete)

cupcakearmy:
Noticed that too, would be nice to make them consistent

metonym:
Released in [v0.92.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.92.0)

PR #2336, opened 2025-11-10
Standardize form component labels to use `labelText` property

spburtsev:
Based on [https://github.com/carbon-design-system/carbon-components-svelte/issues/534](https://github.com/carbon-design-system/carbon-components-svelte/issues/534)

**Breaking changes:**
- Combobox: `titleText` -> `labelText`
- Dropdown: `titleText` -> `labelText`
- MultiSelect: `titleText` -> `labelText`
- NumberInput: `label` -> `labelText`

I changed label property names to `labelText` (as it was the most common name already) in some components so it is consistent for those components that are often used in form as inputs.

I did not change `ProgressStep`, `SelectItem` and `Tab` as those components are less likely to be used in forms and in those components `label` or `text` properties actually refer to something different.

### DIFF
--- a/tests/ComboBox/ComboBoxCustom.test.svelte
+++ b/tests/ComboBox/ComboBoxCustom.test.svelte
@@ -24,7 +24,7 @@
   {direction}
   {shouldFilterItem}
   {itemToString}
-  titleText="Contact with icons"
+  labelText="Contact with icons"
   placeholder="Select contact method"
   on:select={(e) => {
     console.log("select", e.detail);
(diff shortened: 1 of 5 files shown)


metonym:
@spburtsev Thanks for this!

I've re-created this PR in #2337 since I'm unable to merge via the GitHub UI (even though there aren't conflicts).

Preserved your Git commit authorship, but tweaked the commit messages to signify breaking changes (`!). I also updated the docs accordingly.

---

Screenshot 2025-11-10 at 1 51 58 PM

PR #2337, opened 2025-11-10
fix!: rename form label prop names for consistency

metonym:
Closes #2336, closes https://github.com/carbon-design-system/carbon-components-svelte/issues/534

---

PR description copied from #2336

**Breaking changes:**
- Combobox: `titleText` -> `labelText`
- Dropdown: `titleText` -> `labelText`
- MultiSelect: `titleText` -> `labelText`
- NumberInput: `label` -> `labelText`

I changed label property names to `labelText` (as it was the most common name already) in some components so it is consistent for those components that are often used in form as inputs.

I did not change `ProgressStep`, `SelectItem` and `Tab` as those components are less likely to be used in forms and in those components `label` or `text` properties actually refer to something different.


metonym:
Released in [v0.92.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.92.0)

Issue #2338, opened 2025-11-11
modal design logic


Nyx2022:
hello, i'm using modal to show something with the bind:open={open},when i set open to false,it do not actually show the modal in ui,howerver,the display od modal's div is not none.whick make the elements in th modal will be captured by mouse,which affect the elements not in modal,so i advise you change the logic

metonym:
Released in [v0.92.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.92.0)

PR #2339, opened 2025-11-11
fix(modal): prevent pointer events on closed modals

metonym:
Modals in this library are rendered immediately to the DOM; the `open` prop only controls the display state.

This fixes a bug where closed modals were intercepting mouse events despite being visually hidden. This was reported in #2338 where hidden modals with `visibility: hidden` were still capturing clicks.

**Changes**

- Added `style:pointer-events={open ? undefined : "none"}` to both `Modal` and `ComposedModal` components.
  - When a modal is closed, it now has `pointer-events: none`, allowing clicks to pass through to underlying elements.
  - When open, the style is undefined (normal pointer event behavior). This approach maintains the existing visibility transitions and accessibility features while fixing the interaction issue.

### DIFF
--- a/src/Modal/Modal.svelte
+++ b/src/Modal/Modal.svelte
@@ -162,6 +162,7 @@
   class:bx--modal-tall={!passiveModal}
   class:is-visible={open}
   class:bx--modal--danger={danger}
+  inert={open ? undefined : true}
   {...$$restProps}
   on:keydown
   on:keydown={(e) => {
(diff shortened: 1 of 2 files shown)


brunnerh:
The only way I can see clicks possibly being captured would be during the fade out animation.
Elements with `visibility: hidden` are not interactive.

Also, if anything I would recommend setting `inert`, this also captures keyboard interactions.

metonym:
Thanks @brunnerh! Amended to use `inert`.

metonym:
Released in [v0.92.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.92.0)

PR #2343, opened 2025-11-17
perf: use passive event listeners

metonym:
This adds passive event listeners to touch and mouse event handlers to improve scroll performance, particularly on mobile devices. Modern browsers must wait for event handlers to complete before scrolling in case they call `preventDefault()`. 

**Changes**

- breakpointObserver: media query change
- ContextMenuOption: mousemove
- Slider: mousemove, mouseup, touch

**References**

- [MDN: EventTarget.addEventListener()](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener)
- [Chrome DevTools: Use passive listeners to improve scrolling performance](https://developer.chrome.com/docs/lighthouse/best-practices/uses-passive-event-listeners)

### DIFF
--- a/src/Breakpoint/breakpointObserver.js
+++ b/src/Breakpoint/breakpointObserver.js
@@ -39,7 +39,7 @@ export function breakpointObserver() {
     }
 
     for (const [_size, queryList] of matchers) {
-      queryList.addEventListener("change", handleChange);
+      queryList.addEventListener("change", handleChange, { passive: true });
     }
 
     return () => {


brunnerh:
The legacy syntax would have a `passive` event modifier, so you probably could do things like `on:touchstart|passive={...}` instead of all the add/remove.

In Svelte 5, `addEventListener` generally should be migrated to use [`on`](https://svelte.dev/docs/svelte/svelte-events#on) which uses event delegation for some additional performance improvement, so the code might change again anyway. Also, when the syntax is switched to `ontouchstart`, the [handler will be passive by default](https://svelte.dev/docs/svelte/v5-migration-guide#Breaking-changes-in-runes-mode-Touch-and-wheel-events-are-passive).

> When using onwheel, onmousewheel, ontouchstart and ontouchmove event attributes, the handlers are [passive](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#using_passive_listeners) to align with browser defaults.

(`mousemove` appears to be missing on that list, though.)

metonym:
@brunnerh As always, thanks for the review and insight.

Pushed a commit that restores idiomatic usage of the passive modifier: [e117c19](https://github.com/carbon-design-system/carbon-components-svelte/pull/2343/commits/e117c197caee9b2e3d7eeb2794fe797e39399764)

PR #2349, opened 2025-11-21
fix(context-menu): close on any click instead of only outside clicks

### LINKED ISSUE #578: ContextMenu close logic unexpected
Currently it only closes if the context menu is in the sub-tree of the clicked element. I would expect it close when I click anywhere outside the menu. Most applications seem to handle it this way as well.

E.g. if a context menu is defined for each row in a table, opening the menu of one row then clicking on another will not close the menu, because those are separate trees.

metonym:
Fixes #578

Fixes a bug where the ContextMenu component was closing on any window click instead of only closing when clicking outside the menu. This caused issues in scenarios with multiple context menu triggers (like table rows) where clicking a different trigger would close the currently open menu before the new one could open.

The root cause was that the window click handler didn't distinguish between clicks inside and outside the menu element. The fix adds a `ref.contains(e.target)` check to ensure the menu only closes when clicking outside its DOM tree.

### DIFF
--- a/src/ContextMenu/ContextMenu.svelte
+++ b/src/ContextMenu/ContextMenu.svelte
@@ -166,7 +166,9 @@
   }}
   on:click={(e) => {
     if (!open) return;
-    close();
+    if (ref && !ref.contains(e.target)) {
+      close();
+    }
   }}
   on:keydown={(e) => {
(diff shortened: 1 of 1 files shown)


metonym:

PR #2351, opened 2025-11-21
fix(file-uploader): use `button` for accessibility

b-r-i-a-n-w-e-s-t:
Currently the file upload button is announced in VoiceOver as "[LabelText], group." This seems to occur because of the span with role="button" inside the label.

This replaces the <label> with a proper <button> element for correct semantics and screen reader announcements. Native keyboard handling (Enter/Space) now works without custom event handlers. VoiceOver now announces, "[LabelText], button."


metonym (approved):
Nice, thank you!

metonym:

PR #2354, opened 2025-11-22
fix(number-input)!: remove unused `iconDescription` prop

metonym:
The `iconDescription` prop was never properly functional as it only acted as a fallback when `translateWithId` returned empty strings.

Users should use `translateWithId` to customize button labels instead.


metonym:

PR #2362, opened 2025-11-23
feat(theme): export `themes` constant

### LINKED ISSUE #939: Allow exportable themes to add custom themes
https://github.com/carbon-design-system/carbon-components-svelte/blob/3cadf2c00132ef4dbcdd5e7eb8e6de5f02387991/src/Theme/Theme.svelte#L51

metonym:
Closes #939

Export the `themes` constant from the Theme component to provide programmatic access to all available Carbon theme keys and their human-readable labels. 

**Changes**

The `themes` constant is now exported as a module-level constant from `Theme.svelte`, making it available for import alongside the Theme component. This constant was previously only available internally within the Theme component.

```js
import { themes } from "carbon-components-svelte/src/Theme/Theme.svelte";

// Access all theme keys
const themeKeys = Object.keys(themes);

// Build custom UI with correct labels
Object.entries(themes).forEach(([key, label]) => {
  // Use key and label to build theme options
});
```

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -1,6 +1,49 @@
+import fs from "node:fs";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
 import { svelte, vitePreprocess } from "@sveltejs/vite-plugin-svelte";
 import { optimizeImports } from "carbon-preprocess-svelte";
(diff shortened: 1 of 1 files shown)


metonym:

PR #2375, opened 2025-11-29
fix(tile-group): improve context types for `SelectableTileGroup` and `TileGroup`

metonym:
Strongly type `groupName` and `groupRequired` context properties in `TileGroup` components.

**Changes**

- Added explicit JSDoc type annotations for `groupName` and `groupRequired` writable stores in both `TileGroup.svelte` and `SelectableTileGroup.svelte`
- Created typed readonly store variables (`groupNameReadonly`, `groupRequiredReadonly`) with explicit `Readable<string | undefined>` and `Readable<boolean | undefined>` types before passing to `setContext`
- Replaced inline `readonly(groupName)` calls with the typed readonly store variables to enable proper type inference by sveld
- Updated reactive statements to use `.set()` method calls instead of direct assignment for writable stores

### DIFF
--- a/docs/src/COMPONENT_API.json
+++ b/docs/src/COMPONENT_API.json
@@ -13746,7 +13746,8 @@
             },
             {
               "name": "groupName",
-              "type": "any",
+              "type": "import(\"svelte/store\").Readable<string | undefined>",
+              "description": "",
               "optional": false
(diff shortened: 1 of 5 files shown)


metonym:

Issue #1644, opened 2023-01-31
[Documentation] Show how to put a ToastNotification in a UiShell


uvulpos:
Hi there,
I'm having a hard time to figure out where to put a toast notification correctly, does it come in the content section or, ...? Could you add this to the documentation?

brunnerh:
The [design system docs](https://carbondesignsystem.com/components/notification/usage/#toast-notifications) show such notifications in the top right under the header.

This might be more than a documentation issue.
It probably would be helpful to have something like a `ToastNotificationPanel` for this which:
- Takes care of stacking and animating notifications correctly
- Spaces them below the header without being scrolled off-screen if the page is long

derekhearst:
I would really love to be able to call this programmatically, is there a recommended library for this?

metonym:
Added an example to the documentation: https://svelte.carbondesignsystem.com/components/UIShell#header-with-toast-notification

A proper `NotificationQueue` component is forthcoming: [#2283](https://github.com/carbon-design-system/carbon-components-svelte/issues/2283)

uvulpos:
Awesome! Will check out later :)

metonym:
`NotificationQueue` added in [v0.94.0.](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.94.0)

See [Documentation.](https://svelte.carbondesignsystem.com/components/NotificationQueue)

```svelte
<script>
  import { NotificationQueue } from "carbon-components-svelte";

  let queue;

  function showSuccess() {
    queue.add({
      kind: "success",
      title: "Saved",
      subtitle: "Your changes have been saved.",
      timeout: 3000
    });
  }

  function showError() {
    queue.add({
      id: "network-error",
      kind: "error",
      title: "Network Error",
      subtitle: "Please check your connection."
    });
  }
</script>

<NotificationQueue bind:this={queue} position="top-right" />
<button on:click={showSuccess}>Save</button>
<button on:click={showError}>Simulate Error</button>
```

PR #2377, opened 2025-11-29
feat(notification): add `NotificationQueue` component

### LINKED ISSUE #2283: Add `NotificationQueue` component
Another basic functionality for a component library: a component that manages and instruments a notification queue.

The motivation is to replace the `HeaderWithToast.svelte` example in UI Shell, and support this at the library level.

https://github.com/carbon-design-system/carbon-components-svelte/blob/master/docs/src/pages/components/UIShell.svx#L92-L96

Component API:

- It should use `ToastNotification` and allow consumers to imperatively `add` and `remove` notification elements.
- Each element can be keyed, so that duplicate notifications are not appended. Auto-dismiss/customizable timeouts should be forwarded to ToastNotification.
- It should support top right and bottom right placements. For top-right, it should sit below the shell. The consumer should be able to customize the margin offsets for either position).

Related #1644

metonym:
Closes #2283, related #1644

This adds a new `NotificationQueue` component that provides an imperative API for managing toast notifications. This addresses the common use case where applications need to programmatically show notifications without managing component state declaratively.

The queue automatically handles positioning, deduplication, and overflow management.

**Changes**

- Add `NotificationQueue` component with three imperative methods via component instance binding:
  - `add()`, `remove()`, and `clear()`
  - Notifications are deduplicated by id. If a notification with the same id already exists in the queue, subsequent `add()` calls are ignored. The queue enforces a maximum of 3 visible notifications, automatically removing the oldest when the limit is exceeded.

Positioning is configurable via props with support for top-right and bottom-right placement. The component uses fixed positioning with customizable offsets and a default z-index of 9000. Each notification in the queue is rendered using the existing `ToastNotification` component.

```svelte
<script>
  import { NotificationQueue } from "carbon-components-svelte";

  let queue;

  function showSuccess() {
    queue.add({
      kind: "success",
      title: "Saved",
      subtitle: "Your changes have been saved.",
      timeout: 3000
    });
  }

  function showError() {
    queue.add({
      id: "network-error",
      kind: "error",
      title: "Network Error",
      subtitle: "Please check your connection."
    });
  }
</script>

<NotificationQueue bind:this={queue} position="top-right" />
<button on:click={showSuccess}>Save</button>
<button on:click={showError}>Simulate Error</button>
```

### DIFF
--- a/docs/src/pages/framed/UIShell/HeaderWithToast.svelte
+++ b/docs/src/pages/framed/UIShell/HeaderWithToast.svelte
@@ -6,22 +6,23 @@
     Content,
     Grid,
     Header,
-    HeaderGlobalAction,
-    HeaderUtilities,
+    NotificationQueue,
     Row,
     SkipToContent,
     Stack,
-    ToastNotification,
   } from "carbon-components-svelte";
-  import Notification from "carbon-icons-svelte/lib/Notification.svelte";
-  import { fade } from "svelte/transition";
 
-  let showToast = false;
-  let toastPosition = "top";
+  let queue;
+  let queuePosition = "top-right";
 
-  const triggerToast = (position = "top") => {
-    toastPosition = position;
-    showToast = true;
+  const triggerToast = (position) => {
+    queuePosition = position;
+    queue.add({
+      kind: "success",
+      title: "Success",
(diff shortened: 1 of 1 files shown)


metonym:

Issue #2379, opened 2025-11-29
`InlineNotification` title/subtitle should be optional


metonym:
Related #2378

```svelte
      <div class:bx--inline-notification__text-wrapper={true}>
        <p class:bx--inline-notification__title={true}>
          <slot name="title">{title}</slot>
        </p>
        <div class:bx--inline-notification__subtitle={true}>
          <slot name="subtitle">{subtitle}</slot>
        </div>
        <slot />
      </div>
```

metonym:
Released in [v0.94.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.94.0)

PR #2381, opened 2025-11-29
fix(toast-notification): conditionally render title/subtitle/caption

### LINKED ISSUE #2378: `ToastNotification` title/subtitle/caption should all be optional
I should be able to omit the title/subtitle/caption props. Currently, an empty element is rendered, causing visual displacement in the toast.

```svelte
    <div class:bx--toast-notification__details={true}>
      <h3 class:bx--toast-notification__title={true}>
        <slot name="title">{title}</slot>
      </h3>
      <div class:bx--toast-notification__subtitle={true}>
        <slot name="subtitle">{subtitle}</slot>
      </div>
      <div class:bx--toast-notification__caption={true}>
        <slot name="caption">{caption}</slot>
      </div>
      <slot />
    </div>
```

metonym:
Fixes #2378, fixes [#2379](https://github.com/carbon-design-system/carbon-components-svelte/issues/2379)

Only render title, subtitle, and caption elements when the corresponding prop or slot is provided, preventing empty HTML elements from causing visual displacement in the toast notification.

Supports #2377

### DIFF
--- a/src/Notification/InlineNotification.svelte
+++ b/src/Notification/InlineNotification.svelte
@@ -94,12 +94,16 @@
         iconDescription={statusIconDescription}
       />
       <div class:bx--inline-notification__text-wrapper={true}>
-        <p class:bx--inline-notification__title={true}>
-          <slot name="title">{title}</slot>
-        </p>
-        <div class:bx--inline-notification__subtitle={true}>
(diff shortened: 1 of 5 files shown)


metonym:

PR #2382, opened 2025-11-29
fix(toast-notification): conditionally render subtitle/caption

metonym:
Revises 4f6646de from #2381

Amend an unreleased fix to require `title` for `ToastNotification`. Without a title, a margin is not rendered, causing visual imbalance if only `subtitle` or `caption` is provided.

This also adds examples to the docs.

Supports #2377


metonym:

Issue #310, opened 2020-10-07
Modal with Form on:submit method


albertms10:
When a Form is placed inside a Modal, the Modal `on:submit` method is called with a `target: null` property either clicking on the primary button or hitting enter (as `shouldSubmitOnEnter` is `true` by default). Shouldn’t it contain a reference to the `HTMLFormElement` instead? Or do I need to manually manage `on:click` primary button to submit the form?

```svelte
<Modal
  ...
  hasForm
  on:submit={(e) => {
    console.log('modal', e);
  }}
>
  <Form
    on:submit={(e) => {
      console.log('form', e); // Is never printed
    }}
  >
    <TextInput name="name" labelText="Name" required />
    <TextInput name="email" labelText="Email" required />
  </Form>
</Modal>
```

```console
modal CustomEvent {..., target: null}
```

metonym:
This is most definitely a bug. I'd expect `<Modal on:submit />` to return a reference to the form. Would you like to work on this?

albertms10:
I don’t know if I understand what should be going on here, but, for the Modal `on:submit` event to contain a reference to the Form as a target, it should be placed inside the Modal component definition, not as a child of it, right?

So, @metonym, what is the expected way of this to work in an _automatic_ way?—without having to manually manage the Form reference to the Modal outside of it.

metonym:
After looking at this again, I don't think the Form state should be managed by Modal. That being said, I feel that a prop like `hasForm` ideally should be set automatically if a Form is present. `hasForm` is purely cosmetic.

albertms10:
As Svelte cannot know which elements are contained in the component’s `slot`, `hasForm` could be there to set the appropriate style—or the CSS styles should take that into account.
So, as you stated, the Modal `on:submit` should not be considered to also submit the Form content. To make things easy, something like what is described [in this REPL](https://svelte.dev/repl/53b51d2080914e71bb1fbaf78d65d444?version=3.29.4)—by using stores—should make it work.
If this is the expected behaviour, we should add an example to the docs, as well.

brunnerh:
Both `hasForm` and `hasScrollingContent` seem lacking in terms of documentation.

sallaben:
`shouldSubmitOnEnter` doesn't seem to work at all!

metonym:
@sallaben Good catch. Tracking in a separate issue: https://github.com/carbon-design-system/carbon-components-svelte/issues/1005

[...]

PR #2383, opened 2025-11-30
feat(modal): add `formId` prop to enable native form submission

### LINKED ISSUE #1006: Get rid of Modal's `shouldSubmitOnEnter`?
To me the functionality seems more trouble than it is worth, especially since it is the default. For example I encountered these issues before:

- By default the close `X` button is focused, so the user might expect that pressing Enter just closes the dialog. Instead it will cause the primary action, which in case of an irreversible effect could be pretty bad.
- It interferes with some form mechanics. E.g. the dropdown of `<select>` elements can be opened via Enter on some browsers/OSs. With `shouldSubmitOnEnter` it submits the dialog instead.

I would suggest to focus the first input in the dialog content if there is one or if not, focus the primary button by default. If the button is focused it will submit by pressing Enter without having this additional flag.

Also, in the case of having inputs, they should just be used with a form which provides its own logic of submitting on Enter if applicable (see e.g. exception for `<select>` elements). The submission just needs to be forwarded to the Modal; maybe this process could be optimized.

metonym:
Closes [#310](https://github.com/carbon-design-system/carbon-components-svelte/issues/310), closes #1006

This adds a `formId` prop to `Modal to enable native form submission from the primary button. This allows the primary button to submit a form element even when the button is rendered *outside* the form's DOM hierarchy.

When `formId` is provided, clicking the primary button or pressing Enter triggers the form's submit event, while maintaining backward compatibility with existing Modal event dispatching.

This differs from the existing `shouldSubmitOnEnter` prop, as it is designed for native `form` usage. `shouldSubmitOnEnter` does not actually "click" the primary button; instead, the user must handle the primary button and submission/close.

**Usage**

```svelte
<Modal
  bind:open
  modalHeading="Create account"
  primaryButtonText="Submit"
  secondaryButtonText="Cancel"
  hasForm
  formId="account-form"
  on:click:button--secondary={() => (open = false)}
>
  <form id="account-form" on:submit|preventDefault={handleSubmit}>
    <Stack gap={5}>
      <TextInput bind:value={name} labelText="Name" placeholder="Enter name" />
      <TextInput
        bind:value={email}
        labelText="Email"
        type="email"
        placeholder="Enter email"
      />
    </Stack>
  </form>
</Modal>

```

### DIFF
--- a/docs/src/pages/components/Modal.svx
+++ b/docs/src/pages/components/Modal.svx
@@ -34,6 +34,12 @@ Create a non-interactive modal for displaying information. This variant lacks ac
 
 <FileSource src="/framed/Modal/PassiveModal" />
 
+## With form submission
+
+Connect the modal's primary button to a form using the `formId` prop. This enables native form submission even though the button is outside the form element. Clicking the primary button or pressing Enter will trigger the form's submit event.
+
+<FileSource src="/framed/Modal/ModalWithForm" />
+
 ## Has scrolling content
 
 Enable vertical scrolling for modals with lengthy content. This ensures all content remains accessible while maintaining a reasonable modal height.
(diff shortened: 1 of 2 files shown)


metonym:

Issue #835, opened 2021-10-02
Modals when the svelte part of a website isn't full-width


mrme44:
When adding a modal, the modal is created within a position fixed element with a width of 100vw. This can make the modal run off the screen if there are any parent components with position: relative that are not full-width, or any [transform animations](https://www.darrenlester.com/blog/why-fixed-position-element-not-relative-to-viewport) on a parent container that are not full-width.

One solution can be to make sure modal components are always placed at a top level in the application. That would work, except, Svelte can be embedded within a larger application. This is the problem I'm running into. There are 2 better solutions that would work:

Solution 1) Change the presentation element created by the modal to have a width and height of 100% instead 100vw/100vh. Since the modal can currently only be used in its intended form as a top-level element where 100% is the same as 100vw, I don't see how anyone can be negatively affected by this change.

Solution 2) Inject the modal element into either the body element or the root element of the svelte application. This can be done with the technique shown in [this example](https://svelte.dev/repl/dd6088388b564a70b710e43504f2c193?version=3.24.0).

Solution 1 would make the modal only show up over part of the screen, instead of full-width, when used in sub-components. Solution 2 would make sure the modal is always shown full-width. This is probably the better option.

mrme44:
I have addressed this problem in my project by creating the following component that I use instead of the modal component:

```
<script>
import {onMount} from 'svelte';
import { Modal } from "carbon-components-svelte";
let modal;
onMount(()=>{
    document.body.appendChild(modal)
})
</script>

<div bind:this={modal}>
    <Modal {...$$restProps}>
        <slot />
    </Modal>
</div>
```

brunnerh:
The logic of appending the content to the body can also be wrapped in a `use` action, that way you just need to add one simple wrapper and no additional components.

```svelte
<script>
  import { portal } from './portal.js';
  // ...
</script>

<div use:portal={document.body}>
  <Modal .../>
</div>
```

[REPL](https://svelte.dev/repl/8aa94bc793d64a978305f08b96c118db?version=3.43.1) (made for https://github.com/sveltejs/svelte/issues/1133)

metonym:
Related #309

metonym:
`Portal` component added in [v0.95.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.95.0).

[...]

PR #2388, opened 2025-11-30
feat(portal): add `Portal` component

### LINKED ISSUE #309: Modal position inside UIShell with a Header
When opening a Modal in a UIShell with a Header, it is positioned as if no Header was present, only taking into account a `max-height: 90%` of its parent. Therefore, the Modal footer is cut when the screen is not tall enough.

Svelte screen capture desktop

Moreover, when opening it in a mobile device, it is far taller and the footer is never shown, as the body is no scrollable due to the `bx--body--with-modal-open` class.

Svelte screen capture mobile

Should the Header component be hidden behind the modal, filling the whole viewport size?

metonym:
Related #2150

Closes #309, closes [#835](https://github.com/carbon-design-system/carbon-components-svelte/issues/835)

Cherry-picks only `Portal`-related changes, omitting `FloatingPortal` for now.

`Portal` would immediately unblock `Modal` usage by consumers. `FloatingPortal` is better suited for internal usage with tooltips/popovers etc..

This adds a new `Portal` component that renders its content into a fixed container appended to `document.body`. This enables components like modals, tooltips, and dropdowns to escape parent overflow constraints and z-index stacking contexts, which is essential for proper layering and positioning in complex UI layouts.

**Changes**

- Add `Portal` component that renders content into a fixed container in `document.body`
  - Implement a shared container approach with automatic cleanup when all instances unmount

**Usage**

```svelte
<script>
  import { Portal } from "carbon-components-svelte";
</script>

<div>
  <Portal>
    Parent is document.body, not div
  </Portal>
</div>
```

### DIFF
--- a/docs/src/pages/components/ComposedModal.svx
+++ b/docs/src/pages/components/ComposedModal.svx
@@ -18,6 +18,12 @@ Create a modal with a header, body, and footer. Each section can be customized i
 
 <FileSource src="/framed/Modal/ComposedModal" />
 
+## With Portal
+
+Wrap `ComposedModal` in a `Portal` to ensure it renders above all z-index stacking contexts and parent overflow constraints, preventing visual clipping and layering issues.
+
+<FileSource src="/framed/Portal/ComposedModalPortal" />
+
 ## Prevent default close behavior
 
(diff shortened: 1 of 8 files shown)


metonym:

Issue #2105, opened 2025-02-27
[Svelte 5] Cannot use `#snippet` with named slots


FabrizioFerrari-ILS:
When converting to Svelte 5, I think there is a problem due to the same name used for properties and slots. For example, this works:

```
<FileUploaderDropContainer bind:files>
	<svelte:fragment slot="labelText">
		<Icon icon="fluent:document-add-48-regular" />
		Add file
		<br />
		(1GB max)
	</svelte:fragment>
</FileUploaderDropContainer>

```

If I convert the slot to a snippet

```
<FileUploaderDropContainer bind:files>
	{#snippet labelText()}
		<Icon icon="fluent:document-add-48-regular" />
		Add file
		<br />
		(1GB max)
	{/snippet}
</FileUploaderDropContainer>

```
I get this error on "labelText"

`Type '() => any' is not assignable to type 'string'.ts(2322)`

probably because Svelte engine tries to assign the snippet to the property.
It would be a great effort to migrate all components to Svelte 5 snippets, but avoiding name conflicts would allow to use them with the new syntax, keeping slots internally.

metonym:
This issue has been resolved in [v0.96.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.96.0).

In Svelte 5, props and slots ([`#snippet`](https://svelte.dev/docs/svelte/snippet)) must have distinct names to avoid conflicts. Slots that previously shared names with props to use the `Children` suffix (e.g., `labelText` → `labelChildren`).

<details>
<summary><b>Migration guide</b></summary>

### Form Components: `labelText` → `labelChildren`

#### Svelte 3/4

```diff
<TextInput>
-  <span slot="labelText">Custom Label</span>
+  <span slot="labelChildren">Custom Label</span>
</TextInput>
```

#### Svelte 5

```svelte
<TextInput>
  {#snippet labelChildren()}
    <span>Custom Label</span>
  {/snippet}
</TextInput>
```

### Notification Components: `title`/`subtitle`/`caption` → `titleChildren`/`subtitleChildren`/`captionChildren`

#### Svelte 3/4

```diff
<InlineNotification>
-  <strong slot="title">Error: </strong>
-  <span slot="subtitle">An internal server error occurred.</span>
+  <strong slot="titleChildren">Error: </strong>
+  <span slot="subtitleChildren">An internal server error occurred.</span>
</InlineNotification>
```

#### Svelte 5

```svelte
[...]

PR #2408, opened 2025-12-05
fix!: use distinct names for label props and slots (svelte 5 fix)

spburtsev:
Fixes [#2411](https://github.com/carbon-design-system/carbon-components-svelte/issues/2411), supports [#2105](https://github.com/carbon-design-system/carbon-components-svelte/issues/2105)

### Breaking changes:
* `labelText` slot renamed to `labelChildren` to resolve Svelte 5 errors when passing snippets

In Svelte 5, having both a `labelText` prop and `labelText` slot causes the static analyzer to flag errors when using snippet syntax. Renaming the slot eliminates this error and creates a clearer distinction between text-only labels (prop) and rich content labels (slot/snippet).


metonym:

PR #2420, opened 2025-12-13
feat(inline-notification): add `open` prop

### LINKED ISSUE #630: Inline Notification - Ability to Reuse
Would it be possible to open the ability to reuse the Inline Notification control, please?
Right now, it appears there is no way to show a notification, hide it, and then show it again with either new or the same message.
It is fairly common to display error messages or success messages in the same place of the page, depending on user actions.

metonym:
Closes #630

Adds the ability to reuse InlineNotification by making the `open` prop controllable. Previously, once a notification was closed, it couldn't be reopened without recreating the component instance. This change allows parent components to control notification visibility by binding to the `open` prop.

**Changes**

- Made `open` an exported prop that defaults to `true` for backward compatibility.
  - When bound from a parent component, the notification can be shown, hidden, and shown again with the same or different messages.
  - The reactive timeout handling was updated to properly manage timeouts when `open` changes, ensuring timeouts are cleared when the notification is closed and restarted when reopened.
- Added tests and a documentation example.

**Usage**

```svelte
<script>
  import { InlineNotification, Button } from "carbon-components-svelte";
  
  let notificationOpen = false;
  let message = "An error occurred.";
  
  function showNotification() {
    message = "Operation failed.";
    notificationOpen = true;
  }
</script>

<Button on:click={showNotification}>Show Notification</Button>

<InlineNotification 
  bind:open={notificationOpen}
  title="Error:"
  subtitle={message}
/>
```

### DIFF
--- a/docs/src/pages/components/InlineNotification.svx
+++ b/docs/src/pages/components/InlineNotification.svx
@@ -27,6 +27,12 @@ The component is controlled, allowing you to prevent the default close behavior 
   // custom close logic (e.g., transitions)
 }} />
 
+## Reusable notification
+
+Control the notification visibility by binding to the `open` prop. This allows you to show, hide, and reuse the same notification component with different messages. The `open` prop defaults to `true`.
+
+<FileSource src="/framed/InlineNotification/InlineNotificationReusable" />
+
 ## Slottable title and subtitle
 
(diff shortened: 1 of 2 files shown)


metonym:

PR #2421, opened 2025-12-13
feat(toast-notification): add `open` prop

metonym:
Related #2420, #630

Also export the `open` state as a prop to allow controlled usage of `ToastNotification`.

This differs from a [NotificationQueue](https://svelte.carbondesignsystem.com/components/NotificationQueue) as this allowed standalone, controlled usage of `ToastNotification`.

**Changes**

- Made `open` an exported prop that defaults to `true` for backward compatibility in both InlineNotification and ToastNotification.

**Usage**

```svelte
<script>
  import { ToastNotification, Button } from "carbon-components-svelte";
  
  let open = false;
</script>

<Button on:click={() => (open = true)}}>Show Notifications</Button>

<ToastNotification 
  bind:open={open}
  title="Error"
  subtitle="An error occurred."
/>
```

### DIFF
--- a/docs/src/pages/components/ToastNotification.svx
+++ b/docs/src/pages/components/ToastNotification.svx
@@ -35,6 +35,12 @@ Prevent the default close behavior using `e.preventDefault()` in the `on:close` 
   // custom close logic (e.g., transitions)
 }} />
 
+## Reusable notification
+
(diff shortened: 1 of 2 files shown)


metonym:

PR #2424, opened 2025-12-13
fix!: rename `legend` prop to `legendText`

metonym:
Blocks #2422

Currently, there is a mix of "legend" and "legendText" props. Consistently align on `legendText`, which is the dominant usage and also more explicit.

- **SelectableTileGroup**: rename `legend` to `legendText`
- **TileGroup**: rename `legend` to `legendText`


metonym:

PR #2419, opened 2025-12-10
fix(toast-notification)!: change title/subtitle/caption slot names to support Svelte 5

metonym:
Fixes #2414, supports https://github.com/carbon-design-system/carbon-components-svelte/issues/2105, related https://github.com/carbon-design-system/carbon-components-svelte/pull/2408, related #2416

Renames slot names in notification components to avoid shadowing prop names, which is required for Svelte 5 compatibility. This follows the same pattern as #2416 where `company` prop was renamed to `companyName` and `text` slot was renamed to `textChildren`.

**Changes**

- **InlineNotification**: rename `title` slot to `titleChildren`, `subtitle` slot to `subtitleChildren`
- **ToastNotification**: rename `title` slot to `titleChildren`, `subtitle` slot to `subtitleChildren`, `caption` slot to `captionChildren`

PR #2416, opened 2025-12-08
fix(header)!: change `company` prop to `companyName`

metonym:
Fixes #2415, supports [#2105](https://github.com/carbon-design-system/carbon-components-svelte/issues/2105), related #2408

- **Header**: rename `company` prop to `companyName` to avoid shadowing the slot name. `<noun>Name` is consistent with `platformName`
- **HeaderAction**: rename `text` slot to `textChildren` (similar to #2408)

Issue #2412, opened 2025-12-07
[Svelte 5] Rename `legendText` slot to `legendChildren`


metonym:

metonym:
Fixed in [v0.96.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.96.0)

PR #2425, opened 2025-12-13
fix(tile)!: change legendText slot to legendChildren

metonym:
Closes #2422

Fixes [#2412](https://github.com/carbon-design-system/carbon-components-svelte/issues/2412), supports https://github.com/carbon-design-system/carbon-components-svelte/issues/2105, related https://github.com/carbon-design-system/carbon-components-svelte/pull/2408, related #2416, related #2419

Renames slot names in form group components to avoid shadowing prop names, which is required for Svelte 5 compatibility. This follows the same pattern as #2416 where `company` prop was renamed to `companyName` and `text` slot was renamed to `textChildren`, and #2419 where notification slot names were updated.

**Changes**

- **RadioButtonGroup**: rename `legendText` slot to `legendChildren` to avoid shadowing the `legendText` prop
- **TileGroup**: add `legendChildren` slot (previously implemented but missing from type definitions and documentation)
- **SelectableTileGroup**: add `legendChildren` slot (previously implemented but missing from type definitions and documentation)

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -4397,19 +4397,20 @@ export type CarbonTheme = "white" | "g10" | "g80" | "g90" | "g100";
 
 ### Props
 
-| Prop name  | Required | Kind             | Reactive | Type                            | Default value          | Description                                               |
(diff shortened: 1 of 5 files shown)


metonym:

PR #2437, opened 2025-12-20
fix(file-uploader-button): prevent crash when `files` is cleared

### LINKED ISSUE #2436: `FileUploaderButton` can throw type error when programmatically cleared
`FileUploaderButton` component throws a `TypeError: Cannot read properties of undefined (reading 'name')` when files are cleared (i.e., when the file input's `files` property becomes an empty array).

```svelte
on:change|stopPropagation={({ target }) => {
  files = multiple ? [...files, ...target.files] : [...target.files];

  if (files && !disableLabelChanges) {
    labelText = files.length > 1 ? `${files.length} files` : files[0].name;
  }

  dispatch("change", files);
}}
```

An empty array is still truthy, but the statement attempts to access the first item in the empty object.

metonym:
Fixes #2436

Empty arrays are truthy in JavaScript, so `files &&` check passed even when `files.length === 0`, causing `files[0].name` to throw. Added an explicit length check to prevent accessing an undefined array element.

### DIFF
--- a/src/FileUploader/FileUploaderButton.svelte
+++ b/src/FileUploader/FileUploaderButton.svelte
@@ -188,7 +188,7 @@
   on:change|stopPropagation={({ target }) => {
     files = multiple ? [...files, ...target.files] : [...target.files];
 
-    if (files && !disableLabelChanges) {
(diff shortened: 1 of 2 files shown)


metonym:

PR #2440, opened 2025-12-20
fix(tree-view-node): make `findParentTreeNode` and `computeTreeLeafDepth` stricter

metonym:
These utilities are used internally for styling the indentation of tree nodes. Logic is contained, but it makes runtime usage safer with unit tests.


metonym:

PR #2352, opened 2025-11-22
feat(stack): add `Stack` component

metonym:
Closes #2278, closes #1384, closes #1066

Backports the Stack component from v11 (#1963) to provide a utility for consistent spacing between stacked elements. 

This layout component supports both vertical and horizontal orientations, with gap scales 1-13 mapped to Carbon layout tokens, as well as custom string values via inline styles.

This is essential for managing basic spacing in layouts.

PR #2441, opened 2025-12-20
feat(stack): add `align` and `justify` props

metonym:
#2352 added a `Stack` component, which manages spacing between child elements.

It should also support basic alignment and justification.


metonym:
Released in [v0.97.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.97.0)

PR #2446, opened 2025-12-21
fix(slider): fix typo in `labelText` prop description

metonym:
The named slot was renamed to `labelChildren`.


metonym:

PR #2451, opened 2025-12-21
feat(file-uploader): add `maxFileSize` prop

metonym:
Ports over functionality from https://github.com/carbon-design-system/carbon/pull/21167

Adds a `maxFileSize` prop (expressed in bytes) for `FileUploader`. The error message itself is left for the user to instrument.


metonym:
Released in [v0.98.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.98.0)

PR #2452, opened 2025-12-21
fix(radio-tile): forward `aria-*` attributes to input element

metonym:
Ports fix from https://github.com/carbon-design-system/carbon/pull/21068

Accessibility attributes provided to `RadioTile` should go to the input, not the label.


metonym:
Fixed in [v0.98.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.98.0)

PR #2459, opened 2025-12-23
fix(toolbar-search): replace `JSON.stringify` with custom deep comparison for filtered rows

metonym:
Follow-up to #2458

This replaces `JSON.stringify`-based row comparison in `ToolbarSearch` with a custom deep equality function for improved performance and correctness.

The guard exists to avoid infinite updates in Svelte 5 (see #2143), where rows should only be updated if rows from context actually changed.

However, the existing `JSON.stringify` comparison is inaccurate and suboptimal from a performance standpoint (i.e., large rows).

The custom function should be faster, more accurate (can handle objects, Dates, regular expressions), and safe (avoid crashes by circular references). As the `rows` object grows, memory should also be reduced (i.e., avoiding `JSON.stringify` creating a massive string for comparison).

### DIFF
--- a/src/DataTable/ToolbarSearch.svelte
+++ b/src/DataTable/ToolbarSearch.svelte
@@ -48,6 +48,7 @@
 
   import { afterUpdate, getContext, onMount, tick } from "svelte";
   import Search from "../Search/Search.svelte";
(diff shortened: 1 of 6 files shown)


metonym:

PR #2461, opened 2025-12-23
fix(data-table): inferred `PropertyPath` depth limit supports 3 nested levels

metonym:
Follow-up to #2460

`DataTable` headers is generic. It uses the `PropertyPath` utility type to infer keys based on the object structure of the provided `rows`.

```ts
export type DataTableKey<Row = DataTableRow> = import("./DataTableTypes.d.ts").PropertyPath<Row>;
```

However, the current inferred depth is only two levels deep (e.g., `category.department`). A nesting like `category.department.name` would not work). Test cases added in #2462 illustrate this bug.

```ts
type PathDepth = [never, 0, 1, 2, ...0[]];

// With `D = 10` (default):
// - Level 1: `PathDepth[10] = 0` (from `...0[]` spread) → recurse with `D = 0`
// - Level 2: `PathDepth[0] = never` → **STOP**
```

The issue is that `PathDepth[10]` falls back to `never`, and stops recursing.

Updating it to `PathDepth[3]` correctly uses the `2` in the `PathDepth` mapping:

```ts
type PathDepth = [never, 0, 1, 2, ...0[]]; // PathDepth[3] --> 2
```

**Performance Considerations**

- [Per feedback](https://github.com/carbon-design-system/carbon-components-svelte/pull/2461#issuecomment-3685145457), restrict to 3 nested levels instead of the intended 10 for performance (guard against runaway growth).

---

## Before (inferred depth limit is 2)

Screenshot 2025-12-23 at 8 11 57 AM

## After (inferred depth limit is 3 nested levels)

Screenshot 2025-12-23 at 8 12 37 AM

## Explicit generics

The nesting cap of 3 levels deep also applies to explict generics (e.g., `DataTableHeader<ProductRow>`). This is a reasonable constraint; it's left to the user to map values at a shallower depth.

```ts
type ProductRow = {
  id: string;
  name: string;
  price: number;
  category: {
    name: string;
    department: {
      name: string;
      location: {
        // Inferred keys stops at "region"
        region: {
          city: string;
          country: string;
        };
      };
    };
  };
  inStock: boolean;
};
```

Screenshot 2025-12-23 at 8 29 38 AM

### DIFF
--- a/src/DataTable/DataTableTypes.d.ts
+++ b/src/DataTable/DataTableTypes.d.ts
@@ -6,8 +6,8 @@ type Join<K, P> = K extends string | number
     : never
   : never;
 
-// For performance, the maximum traversal depth is 10.
-export type PropertyPath<T, D extends number = 10> = [D] extends [never]
+// For performance, the maximum traversal depth is 3.
+export type PropertyPath<T, D extends number = 3> = [D] extends [never]
   ? never
   : T extends object
     ? {
--- a/tests/DataTable/DataTable.test.ts
+++ b/tests/DataTable/DataTable.test.ts
@@ -1593,19 +1593,7 @@ describe("DataTable", () => {
           level2: {
             level3: {
               level4: {
-                level5: {
-                  level6: {
-                    level7: {
-                      level8: {
-                        level9: {
-                          level10: {
-                            level11: string;
-                          };
-                        };
-                      };
-                    };
-                  };
-                };
+                level5: string;
               };
             };
           };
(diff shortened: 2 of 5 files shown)


brunnerh:
Note that higher depths can cause type performance issues.
I have a project that has such types for object relation paths so an ORM can join linked entities, at a maximum of 5 levels that generates over 40,000 paths in some cases. Cycles in the types contribute to this explosion.

A mitigating factor here might be that the data for the UI is often mapped to only contain the necessary parts and thus the objects tend to not be that deep or cyclical.

metonym:
Thanks @brunnerh for the insight, as always.

What about a middle ground, say a default max of 3 or 4 levels instead of 10? As you mention, objects are rarely deeper than that for UI.

Increasing the current inferred depth would balance DX with performance.

brunnerh:
A middle ground sounds good to me, that should prevent the most egregious cases.

metonym:

Issue #2463, opened 2025-12-23
Address Svelte 5 quirks identified via unit tests


metonym:
During analysis of `isSvelte5` conditionals in test files (#2397), six potential bugs were identified that require investigation to determine if they are legitimate framework behavior changes or component bugs that need fixing.

This parent issue track investigates and fixes these true positives, if possible.

### 1. ComboBox: Reactive Binding Timing Issue

✅ Fixed in [#2471](https://github.com/carbon-design-system/carbon-components-svelte/pull/2471)

**Issue:** In Svelte 5, when clearing a ComboBox selection via keyboard (Space), the input value binding may not update immediately after the clear event is dispatched. The event fires correctly, but the DOM update is delayed.

**Current Workaround:** Test waits for reactive update with `await tick()` and skips value assertion.

---

### 2. TextInput: Inconsistent Empty Value Handling

✅ Expected Svelte 5 behavior, addressed in #2480 

**Issue:** When clearing a number input in Svelte 5, the value may be represented as an empty string (`""`) instead of `null` in the displayed text content, even though the actual value is `null`.

**Current Workaround:** Test accepts either empty string or "null" as valid.

---

### 3. NumberInput: Inconsistent Value Representation (Preserve Null)

✅ Expected Svelte 5 behavior, addressed in #2480 

**Issue:** When clearing a NumberInput with `allowEmpty: true`, the displayed value text may show as empty string instead of "null" in Svelte 5, even though the input value is correctly `null`.

**Current Workaround:** Test accepts either empty string or "null" as valid for text content.

---

### 4. NumberInput: Inconsistent Value Representation (Parse Empty)

✅ Expected Svelte 5 behavior, addressed in #2480 

**Issue:** When parsing an empty string to null in NumberInput, the displayed value text may show as empty string instead of "null" in Svelte 5.

**Current Workaround:** Test accepts either empty string or "null" as valid. Same as issue 3 - verify if this is expected display behavior or needs normalization.

---

### 5. ContentSwitcher: Focus Management Difference

✅ Fixed in #2469 

**Issue:** Focus management in Svelte 5 may work differently for keyboard navigation. The test skips focus verification entirely in Svelte 5, only checking selection state changes.

**Current Workaround:** Test skips focus check in Svelte 5 and only verifies selection state. Investigate if this is expected Svelte 5 focus management behavior or if the component needs to explicitly manage focus in Svelte 5.

---

[...]

PR #2465, opened 2025-12-23
fix(checkbox): prevent duplicate "check" event dispatches in Svelte 5

metonym:
Follow-up to #2464

Fixes [#2467](https://github.com/carbon-design-system/carbon-components-svelte/issues/2467), supports #2463

This fixes a quirk where the "change" event can be dispatched multiple times in Svelte 5.

In Svelte 5, the Checkbox component was emitting the `check` event multiple times for a single user interaction. The test suite had a workaround that accepted `>= 2` calls in Svelte 5, but the expected behavior is exactly one `check` event per state change.

**Root Cause**

The component used a reactive statement to dispatch the `check` event whenever `checked` changed:

```svelte
$: dispatch("check", checked);
```

In Svelte 5, reactive statements can execute multiple times during a single update cycle due to changes in the reactivity system. This can cause the same event to be dispatched multiple times even when `checked` only changes once. Reference: https://github.com/sveltejs/svelte/issues/6732

**Solution**

This implements a guard mechanism that tracks the previous `checked` value and only dispatches the event when the value actually changes.

**Testing**

The testing matrix for Checkbox across Svelte 3/4/5 should all pass.

Additionally, more comprehensive tests around reactivity and dispatched events were added in #2466

### DIFF
--- a/src/Checkbox/Checkbox.svelte
+++ b/src/Checkbox/Checkbox.svelte
@@ -65,7 +65,17 @@
 
   $: useGroup = Array.isArray(group);
   $: if (useGroup) checked = group.includes(value);
-  $: dispatch("check", checked);
+
+  // Track previous checked value to avoid duplicate dispatches in Svelte 5
+  // The reactive statement will only dispatch when checked changes externally (e.g., via bind:checked)
+  let previousChecked = checked;
+  $: {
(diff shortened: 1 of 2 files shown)


metonym:

PR #2469, opened 2025-12-25
fix(content-switcher): fix focus management for Svelte 5 compatibility

### LINKED ISSUE #2468: [Svelte 5] `ContentSwitcher` does not focus via keyboard navigation
As identified in #2463, in Svelte 5, the switch is not receiving focus when using keyboard navigation (e.g., "Tab").

metonym:

Fixes #2468, supports https://github.com/carbon-design-system/carbon-components-svelte/issues/2463

This PR fixes focus management in ContentSwitcher to work consistently across all Svelte versions. The previous implementation used `afterUpdate()` in the Switch component to auto-focus whenever a tab became selected, which caused timing issues in Svelte 5.

This makes the internal `change` method async to await the next tick before updating.

### DIFF
--- a/docs/src/COMPONENT_API.json
+++ b/docs/src/COMPONENT_API.json
@@ -2734,7 +2734,7 @@
             },
             {
               "name": "change",
-              "type": "(direction: number) => void",
+              "type": "(direction: number) => Promise<void>",
               "description": "",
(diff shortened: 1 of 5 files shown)


metonym:

PR #2471, opened 2025-12-25
fix(combo-box): ensure `bind:value` syncs when clearing via keyboard in Svelte 5

### LINKED ISSUE #2470: [Svelte 5] ComboBox `clear()` does not update input value in Svelte 5
As identified in https://github.com/carbon-design-system/carbon-components-svelte/issues/2463, in Svelte 5, the `ComboBox` clear function does not sync the input value until the next tick.

metonym:

Fixes #2470, supports https://github.com/carbon-design-system/carbon-components-svelte/issues/2463

This PR fixes a reactive binding timing issue in ComboBox when clearing selections via keyboard navigation in Svelte 5. 

When pressing Space on the clear button, the `clear()` function would set `value = ""` synchronously, but the `bind:value` update to the parent component was delayed, causing the input DOM to not reflect the cleared state immediately. This fix ensures binding updates complete before the function returns, requiring `clear()` to be made asynchronous.


metonym:

Issue #2472, opened 2025-12-25
[Svelte 5] `ComboBox` dispatches select event on initial render


metonym:
Identified in https://github.com/carbon-design-system/carbon-components-svelte/issues/2463

metonym:
Fixed in [v0.98.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.98.1)

PR #2473, opened 2025-12-25
fix(combo-box): prevent dispatching initial "close" event in Svelte 5

metonym:

Fixes https://github.com/carbon-design-system/carbon-components-svelte/issues/2472, supports https://github.com/carbon-design-system/carbon-components-svelte/issues/2463

This PR fixes `ComboBox` emitting a select event on initial render in Svelte 5.

When a ComboBox is rendered with a `selectedId` prop, the reactive statement would dispatch a select event during mount because `prevSelectedId` starts as `null` and differs from the initial `selectedId`.


metonym:

PR #2476, opened 2025-12-25
fix(tooltip): prevent dispatching events on initial render in Svelte 5

metonym:
Follow-up to https://github.com/carbon-design-system/carbon-components-svelte/pull/2464

Fixes https://github.com/carbon-design-system/carbon-components-svelte/issues/2475, supports https://github.com/carbon-design-system/carbon-components-svelte/issues/2463

Related #2473

This PR fixes `Tooltip` emitting events on initial render in Svelte 5.

The reactive statement `$: dispatch(open ? "open" : "close")` was dispatching a "close" event during component mount because `open` defaults to `false`.

```svelte
<Tooltip on:close={() => {console.log('close')}}>
  <p>Resources are provisioned based on your account's organization.</p>
</Tooltip>
```

PR #2479, opened 2025-12-25
fix(structured-list): prevent dispatching initial "change" event in Svelte 5

metonym:

Fixes https://github.com/carbon-design-system/carbon-components-svelte/issues/2478, supports https://github.com/carbon-design-system/carbon-components-svelte/issues/2463

Related #2473, #2476

This PR fixes `StructuredList` emitting a change event on initial render in Svelte 5. The reactive statement `$: dispatch("change", $selectedValue)` was dispatching a change event during component mount when `selected` was `undefined`.

### DIFF
--- a/src/StructuredList/StructuredList.svelte
+++ b/src/StructuredList/StructuredList.svelte
@@ -27,6 +27,9 @@
    */
   const selectedValue = writable(selected);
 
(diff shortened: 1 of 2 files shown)


metonym:

PR #2481, opened 2025-12-25
perf(data-table): optimize `resolvePath` and extract table utilities

metonym:
This optimizes `resolvePath` by caching parsed path segments to avoid repeated regex splits when sorting or filtering large datasets. Paths are typically stable (defined in headers), so cache hit rates are high. 

The cache only stores parsed segments, not resolved values. Current implementation uses FIFO eviction with 1000 entry limit to prevent unbounded memory growth.


metonym:
Released in [v0.98.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.98.1)

PR #2483, opened 2025-12-25
perf(tree-view): optimize node traversal

metonym:
Optimizes `TreeView` for large datasets:

- memoize flattenedNodes/nodeIds to avoid O(n) recomputation
- optimized `expandNodes` and `collapseNodes` using a Set to store nodes
- only recreate TreeWalker instance when the DOM structure changes
- batch updates for active/selected IDs
- optimized `showNode` to use path expansion instead of full tree traversal

### DIFF
--- a/src/TreeView/TreeView.svelte
+++ b/src/TreeView/TreeView.svelte
@@ -192,17 +192,20 @@
     const { expand = true, select = true, focus = true } = options;
 
     for (const child of nodes) {
-      const nodes = findNodeById(child, id);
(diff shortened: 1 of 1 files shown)


metonym:
Released in [v0.98.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.98.1)

metonym:
Released in [v0.98.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.98.1)

PR #2516, opened 2026-01-17
fix(select)!: remove unused `SelectValue` type

metonym:
Remove the old `SelectValue` type.

`Select` is now generic, and the type signature is the same. This is a small breaking change but is easy to mitigate.


metonym:
Fixed in [v0.98.6](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.98.6)

PR #2517, opened 2026-01-17
fix(theme): pass `CarbonTheme` to generic `SelectProps`

metonym:
`SelectProps` is generic and requires a parameter.


metonym:
Fixed in [v0.98.6](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.98.6)

PR #2526, opened 2026-01-19
fix(multi-select): prevent dispatching initial "select" event in Svelte 5

### LINKED ISSUE #2525: [Svelte 5] `MultiSelect` with `selectedIds` fires `select` event on load
The event should only fire on user interaction but right now it always fires initially.

If I remember correctly there have been similar issues in other components when using Svelte 5.

metonym:
Fixes #2525

Prior art https://github.com/carbon-design-system/carbon-components-svelte/pull/2473

In Svelte 5, an initial `select` event is dispatched if `selectedIds` is specified. Similar to #2473, this adds a flag to avoid the initial dispatch.

Added regression tests.

### DIFF
--- a/src/MultiSelect/MultiSelect.svelte
+++ b/src/MultiSelect/MultiSelect.svelte
@@ -194,6 +194,7 @@
 
   let highlightedIndex = -1;
   let prevChecked = [];
+  let isInitialRender = true;
 
   /**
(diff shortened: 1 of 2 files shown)


metonym:
Fixed in [v0.98.7](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.98.7)

PR #2546, opened 2026-01-20
fix!: use camel case for named slots for Svelte 5 compatibility

### LINKED ISSUE #2545: [Svelte 5] Cannot use "skip-to-content" or "cell-header" slots as snippets
There are currently two[^1] slots that cannot be used with inline snippets because their name does not work as an identifier.

- `Header`: `skip-to-content`
- `DataTable`: `cell-header`

I would recommend renaming them to `camelCase` to better support the newest syntax.

This is a breaking change but the migration should be straightforward.

[^1]: Regex: `<slot name="[^"]*-[^"]*"`

metonym:
Fixes #2545 

Renames "skip-to-content" and "cell-header" to use camelcase to allow `#snippet` usage in Svelte 5.

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -1664,12 +1664,12 @@ None.
 
 ### Slots
 
-| Slot name       | Default | Props                               | Fallback                         |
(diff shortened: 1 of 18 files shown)


metonym:

PR #2548, opened 2026-01-20
fix(number-input): add styles for `type="text"`

metonym:
Carbon v10 styles only target `input[type='number']`, but `NumberInput` uses `type="text"` when `allowDecimal` is enabled.

This backports v11 styles to ensure that the `type="text"` case is covered.

---

## Before
Screenshot 2026-01-20 at 8 35 22 AM

## After
Screenshot 2026-01-20 at 8 35 15 AM


metonym:

PR #2550, opened 2026-01-20
fix(number-input): allow deleting all characters if `allowDecimal` is true

metonym:
`allowDecimal` uses `type="text"` for `NumberInput`. However, if true, the last digit cannot be deleted. This applies a similar approach for non-allowDecimal by not binding the input value in this case.

```svelte
<NumberInput
  allowDecimal
  value={1.0}
  helperText="Trailing zeros preserved"
  step={0.01}
  labelText="Amount"
/>
```


metonym:

PR #2556, opened 2026-01-25
feat(local-storage): auto-sync updates across tabs

metonym:
Use a window event listener to listen to storage updates to auto-sync updates across tabs.

A use case is toggling the theme locally: updating the theme in one tab should immediately propagate to other tabs (of the same browser).


metonym:

PR #2562, opened 2026-01-25
fix(overflow-menu): item supports `target` and `rel` props

metonym:
Currently, overflow menu items support native links. However, `target` cannot be added since rest props are spread to the top-level `li` element.

This explicitly adds support for `target` and `rel` props. If target is _blank, "nooopener" is automatically used for `rel` (but can be overridden if `rel` is specified). This follows the behavior of `Link`.


metonym:

PR #2564, opened 2026-01-25
feat(combo-box)!: `ComboBoxItemId` is generic, defaulting to `any`

metonym:
This makes `ComboBox` id generic, defaulting to `any` (existing type), allowing readonly values to infer the `id`. Now, both the `item.id` and `selectedId` props are generic, falling back to `any`.

This removes the auto-generated `ComboBoxId`, but this is a trivial breaking change.

---

## Destructured items

Screenshot 2026-01-25 at 1 26 16 PM

## `selectedId`

Screenshot 2026-01-25 at 1 40 00 PM

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -654,10 +654,8 @@ None.
 ### Types
 
 ```ts
-export type ComboBoxItemId = any;
-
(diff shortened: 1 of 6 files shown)


metonym:

PR #2566, opened 2026-01-25
fix(types): generic component props should extend default

metonym:
The generated generic props should extend the specified default to allow indexed access types to be used in the internal component props.

Unblocks #2564


metonym:

PR #2567, opened 2026-01-25
feat(multi-select)!: `MultiSelectItemId` is generic, defaulting to `any`

metonym:
This makes `MultiSelectItem` id generic, allowing readonly values to infer the id type. Now, both the `item.id` and `selectedIds` props are generic, falling back to `any`. The auto-generated `MultiSelectItemId` typedef has been removed (trivial breaking change).

This change mirrors the implementation in the ComboBox component (PR #2564) and includes comprehensive type tests.

---

Screenshot 2026-01-25 at 2 15 08 PM

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -2441,12 +2441,10 @@ None.
 ### Types
 
 ```ts
-export type MultiSelectItemId = any;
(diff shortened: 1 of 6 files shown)


metonym:

PR #2568, opened 2026-01-25
feat(dropdown)!: `DropdownItem` id generic, defaulting to `any`

metonym:
Makes `Dropdown` id generic, defaulting to `any` (existing type), allowing readonly values to infer the `id`. Now, both the `item.id` and `selectedId` props are generic, falling back to `any`.

This removes the auto-generated `DropdownItemId` typedef, matching the pattern from #2564.

**Notably**, unlike `ComboBox` or `MultiSelect` where an item may not be initially selected, `Dropdown` *does* require that an item ID is specified as the inital selected item.

---

Screenshot 2026-01-25 at 2 21 27 PM

### DIFF
--- a/COMPONENT_INDEX.md
+++ b/COMPONENT_INDEX.md
@@ -1209,13 +1209,9 @@ None.
 ### Types
 
 ```ts
-export type DropdownItemId = any;
-
-export type DropdownItemText = string;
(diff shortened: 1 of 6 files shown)


metonym:

PR #2570, opened 2026-01-25
feat(data-table)!: `DataTableRowId` is generic, defaulting to `any`

metonym:
Make `DataTableRowId` generic, using the indexed access type from props.

Additionally, this fixes the `ToolbarSearch` generic to preserve row ID types from the parent `DataTable`. Previously, `filteredRowIds` and `shouldFilterRows` were typed as `any`; now TypeScript infers the correct types automatically.

---

Screenshot 2026-01-25 at 2 57 32 PM

Screenshot 2026-01-25 at 2 57 49 PM

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -141,6 +141,12 @@
       }
     },
     {
+      "includes": ["types/**/*.d.ts"],
+      "formatter": {
+        "enabled": false
+      }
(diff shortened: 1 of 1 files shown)


metonym:

Issue #1192, opened 2022-03-21
DataTable supports virtualized rows


adamrpostjr:
The data table is amazing, one of the better data tables that I have used. The single problem that I have with the data table is that when I have 1000+ items in the list I start to overload the DOM, causing the page to be VERY slow. The only fix for this currently is to paginate.
Pagination works fine, but in the age of Svelte where everything is SO reactive it feels old, slow, and clunky to paginate.

Recently, I have moved away from using carbon components' data table to using a virtual list with headers. 

It would be a MAJOR benefit to have a virtual list in the data table natively!

brunnerh:
I second this. Unfortunately this probably is not a simple thing to add.
Ideally, it also should still support dynamic cell heights, allowing for line breaks according to the content, but I have yet to see a virtualized table implementation that can actually do that properly.

albert-schilling-enerparc:
> but I have yet to see a virtualized table implementation that can actually do that properly

This is actually no problem to implement with popular virtualization libraries like [react-window](https://github.com/bvaughn/react-window) and [tanstack/virtualiz](https://github.com/tanstack/virtual). 

We currently face the same limitation of Carbon. We have our own implementation of a virtualized highly flexibel table with flexibel row heights, sub-rows etc. built with `tanstack/table` and `react-window`. As long as there is no support for virtualization in the Carbon Table, our plan is to adjust our table's styling to CDS. 

But of course, we would prefer having an official CDS Data Table with virtualization.

metonym:
Released in [v0.101.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.101.0)

`DataTable` now supports virtualized rows for large, in-memory datasets.

Virtualization renders only the rows visible in the viewport, improving performance for tables with thousands of rows. Set `virtualize={true}` to enable. The header stays visible while scrolling.

See the [DataTable docs](https://svelte.carbondesignsystem.com/components/DataTable#virtualized-rows-large-datasets) for more.

https://github.com/user-attachments/assets/f2b04fa0-3d3f-4111-aab8-6cbffd25b42f

PR #2573, opened 2026-01-26
feat(data-table): support virtualized rows

metonym:
Closes [#1192](https://github.com/carbon-design-system/carbon-components-svelte/issues/1192)

This adds row virtualization to `DataTable` for large datasets. Only visible rows are rendered in the DOM, which improves performance when displaying thousands of rows. Virtualization is opt-in and works with existing features like sorting, filtering, expandable rows, and row selection.

Previously, virtualization support was added for `ComboBox`, `MultiSelect`, and `Dropdown`.

**Changes**

- Virtualized tables are intended for use with `stickyHeader={true}` so the header stays visible while scrolling.
- New `virtualize` prop: set to `true` for defaults, or pass a config object with `itemHeight`, `maxVisibleRows`, `containerHeight`, `overscan`, `threshold`, and `maxItems`. Row height defaults by size variant (48px medium, 24px compact, 32px short, 64px tall). Virtualization activates only when row count exceeds `threshold` (default 100).

**Example usage**

```svelte
<script>
  import { DataTable, Toolbar, ToolbarContent, ToolbarSearch } from "carbon-components-svelte";

  const rows = Array.from({ length: 10_000 }, (_, i) => ({
    id: i,
    name: "Load Balancer " + (i + 1),
    protocol: "HTTP",
    port: 3000 + i * 10,
    rule: i % 2 ? "Round robin" : "DNS delegation",
  }));

  let filteredRowIds = [];
</script>

<DataTable
  sortable
  stickyHeader
  virtualize={true}
  headers={[
    { key: "protocol", value: "Protocol" },
  ]}
  {rows}
>
  <Toolbar>
    <ToolbarContent>
      <ToolbarSearch persistent placeholder="Search 10,000 rows..." shouldFilterRows bind:filteredRowIds />
    </ToolbarContent>
  </Toolbar>
</DataTable>
```

---

https://github.com/user-attachments/assets/47f2ece7-2f8d-48de-b89e-ca8a542bcd1f

### DIFF
--- a/docs/src/pages/components/DataTable.svx
+++ b/docs/src/pages/components/DataTable.svx
@@ -8,7 +8,7 @@ components: ["DataTable", "Pagination","Toolbar", "ToolbarContent", "ToolbarSear
   import Preview from "../../components/Preview.svelte";
 </script>
 
-`DataTable` displays structured data in a tabular format. Use it to present large datasets with features like sorting, filtering, pagination, and row selection.
+`DataTable` displays structured data in a tabular format. Use it to present large datasets with features like sorting, filtering, pagination, virtualization, and row selection.
 
 <InlineNotification svx-ignore lowContrast title="Note:" kind="info" hideCloseButton>
   <div class="body-short-01">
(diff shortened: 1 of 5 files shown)


metonym:

Issue #2579, opened 2026-01-26
ComboBox - clearFilterOnOpen is not restoring the values as expected


SLAN14:
I'm using Svelte 5, where I'm using the ComboBox with clearFilterOnOpen attribute. The selectedId attribute persits the Id, but the value is not getting restored, if I close the ComboBox without any selection (as mentioned in the docs).

metonym:
Fixed in [v0.99.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.99.0)

PR #2580, opened 2026-01-26
fix(combo-box): `clearFilterOnOpen` restores value after typing

metonym:
Fixes [#2579](https://github.com/carbon-design-system/carbon-components-svelte/issues/2579)

Currently, `ComboBox` with `clearFilterOnOpen` uses `!ref.contains(document.activeElement)` to restore the value. This is too restrictive since the menu may be closed while the input retains focus.

This bug is reproducible by setting an initial `selectedId`:

```svelte
<ComboBox
  labelText="Contact"
  items={[
    { id: "0", text: "Slack" },
    { id: "1", text: "Email" },
    { id: "2", text: "Fax" },
  ]}
  {shouldFilterItem}
  clearFilterOnOpen
  selectedId="1"
/>
```

The value is cleared when focused, but typing is not allowed until re-focusing the input.

This fixes it by removing the focus check so that `valueBeforeOpen` is restored when the menu is closed without a new selection.

### DIFF
--- a/docs/src/pages/framed/ComboBox/ClearFilterOnOpen.svelte
+++ b/docs/src/pages/framed/ComboBox/ClearFilterOnOpen.svelte
@@ -17,5 +17,6 @@
   ]}
   {shouldFilterItem}
   clearFilterOnOpen
+  selectedId="1"
   helperText="Select an item, then reopen to see all items again"
 />


metonym:

PR #2581, opened 2026-01-26
fix(combo-box): keyboard navigation starts at selected item index

metonym:
Similar to https://github.com/carbon-design-system/carbon-components-svelte/pull/2577, keyboard navigation for `ComboBox` should start at the selected item index, not 0.


metonym:

PR #2583, opened 2026-01-27
fix(number-input): include correct styles for `allowDecimal` input

metonym:
`allowDecimal` turns the input into `type="text"`, which misses default v10 styles.

This ports over styles from the source .scss file so that the `allowDecimal` variant uses consistent styles.

---

## Before

Screenshot 2026-01-26 at 6 16 24 PM

## After

Screenshot 2026-01-26 at 6 16 16 PM


metonym:

Issue #2491, opened 2025-12-28
Feature request: keepExpanded property on TreeView


bhavers:
It would be very nice to have a `keepExpanded` property on TreeView.
The desired behaviour would be:
on true: leaves all opened nodes (branches) open
on false: closes other nodes when selecting a node in another branch

Example:
-folder 1
--item 1
--item 2
-folder 2
--item 3
--item 4

on true: selecting folder 1, followed by selecting folder 2 would show the whole tree (including item 1 to 4)
on false: selecting folder 1, followed by selecting folder 2 would first show folder 1 (item 1 and 2), and than show folder (item 3 and 4) but closing folder 1 (hiding item 1 and 2).

Alternative is to (custom) implement this outside the TreeView component by modifying expandedIds. Though i think it would be great if it is a property. Certainly not critical, a nice to have.

metonym:
Interesting. Something like this?

https://github.com/user-attachments/assets/c38cf319-bb01-4e3b-bbd7-0e6b2d621e8c

bhavers:
Exactly that! Great.

metonym:
Released in [v0.99.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.99.0)

Set `autoCollapse` to true on `TreeView` for siblings to be auto-collapsed.

https://svelte.carbondesignsystem.com/components/TreeView#auto-collapse-siblings

bhavers:
@metonym First of all: thanks for all the hard work and the many releases in the recents months!

I tested this new feature and it works great. There's one thing: 
My app contains a Tree component (containing the TreeView) and an Overview component (containing a DataTable). They both show the same reactive `items[]` object and a `selectedId` for the selectedI item. The `selectedId` is binded to the `activeId` prop in TreeView.

When i select an item from the Overview component it selects the item in TreeView, but doesn't take the autoCollapse setting into account (leaves all nodes open). I was thinking of solving this outside the TreeView component with and $effect reaction to `selectedId` that will do a .collapseAll() (when autoCollapse is true), and than setting the activeId. But that feels a bit clunky. Any thoughts? Would this be better solved in TreeView or outside?

Thanks

metonym:
@bhavers Mind providing a minimal repro?

bhavers:
[...]

PR #2585, opened 2026-01-28
fix(tree-view): `autoCollapse` respects `activeId`

metonym:
Fixes [#2491](https://github.com/carbon-design-system/carbon-components-svelte/issues/2491)

Follow-up to #2493, which added `autoCollapse` support to `TreeView`. The auto-collapse behavior should also be triggered when `activeId` changes. See the screen recording below.

---

https://github.com/user-attachments/assets/3a92cbdf-75b4-4bec-805a-d27635678cea


metonym:
Fixed in [v0.99.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.99.1)

PR #2586, opened 2026-01-28
feat(data-table): add `rowClass` prop

Marshallup:
### Summary
Added a new `rowClass` prop to the `DataTable` component. This allows passing a custom CSS class to <tr> elements, enabling per-row styling.

### Motivation
In some use cases, developers need to apply different styles to table rows (e.g., highlight specific rows based on data). Currently, `DataTable` does not provide a way to inject custom classes into <tr>.

### Implementation
- Added optional `rowClass?: string | ((row: DataTableRow) => string)` prop to DataTable.
- Applied `rowClass` to each <tr> in the table body.
- Updated TypeScript types and added a small example in documentation/tests.

### Checklist
- [x] Tests added / updated
- [x] Documentation updated
- [x] Code compiles correctly

### DIFF
--- a/src/DataTable/DataTable.svelte
+++ b/src/DataTable/DataTable.svelte
@@ -67,6 +67,10 @@
    * @property {DataTableCell<Row>} cell
    * @property {EventTarget} target
    * @property {EventTarget} currentTarget
+   * @typedef {{ row: Row, rowIndex: number, selected: boolean, expanded: boolean }} DataTableRowClassArgs<Row=DataTableRow>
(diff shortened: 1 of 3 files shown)


src/DataTable/DataTable.svelte
```diff
@@ -516,7 +539,7 @@
             ? 'bx--parent-row'
             : ''} {expandable && parentRowId === row.id
             ? 'bx--expandable-row--hover'
-            : ''}"
+            : ''} {rowClassValue || ''}"
```
brunnerh: Would generally advise against using this pattern.


```suggestion
            : ''} {rowClassValue ?? ''}"
```

(I think there are still quite a few `||` on non-boolean values elsewhere, but `??` is also being used.)

metonym:

PR #2589, opened 2026-01-29
perf(data-table): optimize row selection/expansion lookups using Sets

metonym:
Replace `Array.includes` with `Sets` for more efficient lookups. Improves performance for tables with many rows.


metonym:

PR #2592, opened 2026-01-29
perf(tree-view): improve lookups using Set, guard store updates

metonym:
Similar to `DataTable` (#2590), this adds performance optimizations for `TreeView`, especially when used with many nodes or many selected/expanded nodes. The same patterns (Set lookups, guarded store updates) improve both performance and consistency.

The motivation is that both of these components likely use a large number of items, necessitating performance.

- only update `activeNodeId` when `activeId` changes
- only update `selectedNodeIds` / `selectedIdsSetStore` when `selectedIds` value changes
- only update `expandedNodeIds` / `expandedIdsSetStore` when `expandedIds` value changes
- use Set lookups (`selectedIdsSetStore.has`, `expandedIdsSetStore.has`) in TreeViewNode and TreeViewNodeList instead of array `.includes()`

### DIFF
--- a/src/TreeView/TreeView.svelte
+++ b/src/TreeView/TreeView.svelte
@@ -333,6 +333,25 @@
   let expandedIdsSet = new Set(expandedIds);
   /** @type {ReadonlyArray<Node["id"]>} */
   let lastExpandedIdsRef = expandedIds;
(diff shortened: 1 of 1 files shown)


metonym:

PR #2601, opened 2026-01-31
fix(code-snippet): type definitions include `$$restProps`

metonym:
`CodeSnippet` has rest props being spread to `span` (inline) or the copy `button` (single/multi).

The type definitions are missing `$$restProps`, requiring an explicit JSDoc annotation.


metonym:

PR #2602, opened 2026-01-31
fix: use nullish coalescing (`??`) to respect falsy values

metonym:
Use nullish coalescing (`??`) instead of logical OR (`||`) for fallbacks so that intentionally falsy values (empty string, `0`) are respected instead of being replaced by defaults.

```js
0 || '' // --> ''
0 ?? '' // --> ''
```

With `||`, any falsy left-hand side triggers the fallback. So passing `aria-label=""` or `item.text=""` would still yield the default label or `item.id`. With `??`, only `null` and `undefined` trigger the fallback, so empty strings and zero are preserved. This aligns component behavior with caller intent and avoids surprising overrides for accessibility and data display.

f4ddac1c adds regression tests that fail without these fixes.

**Changes**

- **CodeSnippet**: `aria-label` now uses `??` so an explicit empty string is kept instead of `copyLabel` or `"code-snippet"`.
- **ComboBox**: `itemToString` uses `??` so `item.text === ""` is shown rather than `item.id`; `overscan` and `currentScrollTop` use `??` so `0` is not replaced; `aria-label` fallback uses `??`.
- **ComposedModal**: `aria-label` uses `??` so empty string is preserved over `$label`.
- **DataTable**: `header.key` uses `??` so `""` is not replaced by `key-${index}`; `hasCustomHeaderWidth` uses `??` so `header.width === 0` is not replaced by `minWidth`.
- **Dropdown**: Same `itemToString`, `overscan`, `currentScrollTop`, and `aria-label` fixes as ComboBox.
- **FormGroup**: `aria-labelledby` uses `??` so an explicit empty string is kept instead of `legendId`.
- **Modal**: `ariaLabel` chain (modalLabel, aria-label, modalAriaLabel, modalHeading) uses `??` throughout.
- **MultiSelect**: `itemToString` and `aria-label` use `??` with the same semantics as ComboBox/Dropdown.
- **OverflowMenu**, **Slider**, **Tabs**, **Toggle**, **ToggleSkeleton**, **UIShell Header**: `aria-label` (and Slider `minLabel`/`maxLabel`) use `??` so empty string and other intentional falsy values are not overridden by defaults.

### DIFF
--- a/src/Dropdown/Dropdown.svelte
+++ b/src/Dropdown/Dropdown.svelte
@@ -23,7 +23,7 @@
    * Override the display of a dropdown item.
    * @type {(item: Item) => string}
    */
-  export let itemToString = (item) => item.text || item.id;
+  export let itemToString = (item) => item.text ?? item.id;
 
   /**
    * Specify the selected item id.


metonym:

PR #2603, opened 2026-01-31
fix(data-table): default string sorting is case- and accent-insensitive

metonym:
`DataTable` column sorting now uses case- and accent-insensitive string comparison by default. This fixes surprising behavior when users sort text columns and expect alphabetical order regardless of case or accents.

## Current default `DataTable` string sorting

```js
String(itemA).localeCompare(String(itemB), "en", { numeric: true })
```

- **Case-sensitive:** "Apple" and "apple" were treated as different, so "Apple" could sort before "apple" (uppercase code points come before lowercase in many collations). Users often expect "A" and "a" to be treated the same when sorting names or labels.
- **Accent-sensitive:** "café" and "cafe" were distinct, so accented and unaccented forms did not group together.
- **Locale fixed to "en":** Collation always used English rules instead of the user's locale.

## New default `DataTable` string sorting

```js
String(itemA).localeCompare(
  String(itemB),
  // Use undefined to default to user's locale
  undefined,
  {
    // Enable numeric sorting for strings that look like numbers
    // E.g., "10" should come after "2"
    numeric: true,
    // Comparison is case- and accent-insensitive
    // E.g., "apple" == "Apple", "café" == "cafe"
    sensitivity: "base",
  },
)
```

- **Case-insensitive:** Matches common expectations for text columns (e.g., names, categories). "Berry", "apple", "Cherry" now sort in true alphabetical order: apple, Berry, Cherry, instead of Cherry, Berry, apple.
- **Accent-insensitive:** "café" and "cafe" compare equal for ordering, so variants group together. Implemented via `Intl.Collator`'s `sensitivity: "base"`, which ignores case and accent differences.
- **User locale:** Passing `undefined` as the locale uses the runtime's default (user/browser locale), so sort order respects the user's language and region instead of hardcoding "en". Numeric sorting (`"2"` before `"10"`) is unchanged via `numeric: true`.

Custom sort functions are unchanged: when a column provides a `sort` function, it is used as before, and this default does not apply.

Tests added in #2600 are unskipped, which illustrate the expected default sorting behavior.

### DIFF
--- a/src/DataTable/data-table-utils.js
+++ b/src/DataTable/data-table-utils.js
@@ -229,10 +229,19 @@ export function compareValues(itemA, itemB, ascending, customSort) {
     } else if (!itemB && itemB !== 0) {
       result = -1;
     } else {
-      // String comparison with locale-aware numeric sorting
-      result = String(itemA).localeCompare(String(itemB), "en", {
-        numeric: true,
-      });
+      result = String(itemA).localeCompare(
+        String(itemB),
+        // Use undefined to default to user's locale
+        undefined,
+        {
+          // Enable numeric sorting for strings that look like numbers
+          // E.g., "10" should come after "2"
+          numeric: true,
+          // Comparison is case- and accent-insensitive
+          // E.g., "apple" == "Apple", "café" == "cafe"
+          sensitivity: "base",
+        },
+      );
     }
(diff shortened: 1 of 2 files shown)


metonym:

PR #2604, opened 2026-01-31
fix(number-input): preserve value during invalid decimal input

metonym:
Currently, typing invalid input like "1.5." in `allowDecimal` mode would immediately set `value` to `null`, losing user data. 

This fixes the behavior so that typing an invalid value ("1.5.") will only re-normalize ("1.5") when blurring the input, and no longer clears the value completely.


metonym:

PR #2607, opened 2026-02-01
fix(number-input): support locale-specific decimal separators

### LINKED ISSUE #2547: NumberInput `allowDecimal` should allow more separators
When using the NumberInput component with the lang prop set to "de", pressing the period key (".") while typing clears the entire input field. I would expect pressing the period key (".") to behave like the comma (",") or be ignored altogether, rather than clearing the input field. It would make for a smoother user experience.

Interestingly, this issue doesn't seem to occur when the lang attribute is set to "en".

metonym:
Fixes #2547

Closes #2605, closes #2594

This is a hot fix for #2547. Currently, only "." separators are allowed when `allowDecimal` is true. This doesn't work for countries that use other separators for decimals, like ",".

- For `allowDecimal`, allow `,` and `٫` when typing
- When blurring, normalize the input

This normalizes input values:

- 3.14 (period) or 3,14 (comma) --> 3.14
- 1,000.5 or 1.000,5 --> 1000.5

---

As a whole, however, `NumberInput` needs a larger re-work to properly handle international, localization formats, currency etc.. (specifically, type="number" is too narrow).

### DIFF
--- a/src/NumberInput/NumberInput.svelte
+++ b/src/NumberInput/NumberInput.svelte
@@ -122,6 +122,11 @@
 
   const dispatch = createEventDispatcher();
 
+  /** Regex to match period/dot characters globally */
+  const RE_DOT = /\./g;
+  /** Regex to match comma and Arabic decimal separator (٫) globally */
+  const RE_COMMA = /[,\u066B]/g;
+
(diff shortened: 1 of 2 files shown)


metonym:

PR #2620, opened 2026-02-01
fix(number-input): dynamically set `aria-describedby` based on state

### LINKED ISSUE #2613: `NumberInput`: `aria-describedby` should be dynamic
React dynamically sets `aria-describedby` to point at the error, warning, or helper text element ID depending on state. The Svelte component always sets `aria-describedby` to a fixed `errorId` regardless of whether there's an error, warning, or helper text.

metonym:
Fixes #2613

Previously, `aria-describedby` always pointed to `errorId` regardless of component state. Now it correctly references the error, warning, or helper text element ID, matching the behavior of `TextInput`.


metonym:

PR #2621, opened 2026-02-01
fix(select): dynamically set `aria-describedby` based on state

metonym:
Related #2620

Currently, `aria-describedby` only references `errorId`, ignoring warning and helper text states. Now dynamically references the appropriate ID based on priority: error > warn > helper.


metonym:

Issue #2609, opened 2026-02-01
`NumberInput`: add `disableWheel` prop


metonym:
React has `disableWheel` to prevent scroll-wheel from changing the input value when focused. The Svelte component has no equivalent.

metonym:
Released in [v0.100.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.100.0)

PR #2622, opened 2026-02-01
feat(number-input): add `disableWheel` prop

metonym:
Closes [#2609](https://github.com/carbon-design-system/carbon-components-svelte/issues/2609)

Parity with the upstream React behavior.

`disableWheel` prevents the mouse wheel from incrementing/decrementing when hovering over the input value.


metonym:

PR #2624, opened 2026-02-03
feat(pagination): add `dynamicPageSizes` prop

metonym:
Set `dynamicPageSizes` to `true` to automatically filter out redundant page sizes based on `totalItems`. Page sizes larger than needed to display all items on a single page are hidden.

For example, with a total of 9 items and page sizes of `[5, 10, 15]`, the page sizes will be `[5, 10]`. This is because 10 is already larger than the total number of items.

```svelte
<Pagination totalItems={9} pageSizes={[5, 10, 15]} dynamicPageSizes />
```


metonym:

Issue #2625, opened 2026-02-03
labelChildren() snippet does not render for `inline` TextInput elements


featherbear:
Testing against v0.99.2

Following the migration guide in https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.96.0, I changed my references usage of labels to a labelChildren() snippet. However, it appears that the labelChildren() snippet is not used when the (i.e.) `TextBox` inline variant is used - causing no label to be shown

Is this intended behaviour (and I should rollback my changes), or is this a bug pending fix?

|Code|Render|
|:---:|:---:|
|✅ Using `labelChildren()` snippet <img width="459" height="127" alt="Image" src="https://github.com/user-attachments/assets/b2077674-653f-4fff-be2d-171f316d84eb" />|<img width="764" height="70" alt="Image" src="https://github.com/user-attachments/assets/aeb4e4d4-a572-434f-b415-ff7600ee8923" />|
|❌ Using `labelChildren()` snippet with `inline` attribute <img width="458" height="121" alt="Image" src="https://github.com/user-attachments/assets/8d319831-043c-4ccf-9258-1890d77b8f68" />|🚨 No label is shown <img width="750" height="43" alt="Image" src="https://github.com/user-attachments/assets/0162990f-80b4-4c38-9605-b2973150a458" />|
|✅ Using `labelText` prop <img width="442" height="38" alt="Image" src="https://github.com/user-attachments/assets/f4c32335-4df4-4a95-9794-8b67b8408eb2" />|<img width="757" height="69" alt="Image" src="https://github.com/user-attachments/assets/ae2da88f-a57c-4101-998e-4e5f000c9fa9" />|
|✅ Using `labelText` prop with `inline` attribute <img width="572" height="50" alt="Image" src="https://github.com/user-attachments/assets/18688f77-a18b-4699-8d1c-2bdaf52441be" />|<img width="782" height="49" alt="Image" src="https://github.com/user-attachments/assets/c0ee53a1-ad99-4c73-9368-771febe35ef3" />|

metonym:
Thanks for the repro – fixed in [v0.99.3](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.99.3)

Example: https://svelte.carbondesignsystem.com/components/TextInput#inline-variant-with-custom-label

PR #2626, opened 2026-02-03
fix(text-input): inline variant supports `labelChildren` slot

metonym:
Fixes [#2625](https://github.com/carbon-design-system/carbon-components-svelte/issues/2625)

The inline `TextInput` should support the `labelChildren` slot.


metonym:
Fixed in [v0.99.3](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.99.3)

Issue #2610, opened 2026-02-01
`NumberInput`: add `stepStartValue` prop


metonym:
React has `stepStartValue` which controls what value stepping begins at when the input is empty (defaults to `0`). The Svelte component has no equivalent: it uses `min ?? 0` as the starting point.

metonym:
Released in [v0.100.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.100.0)

PR #2628, opened 2026-02-08
feat(number-input): add `stepStartValue` prop

metonym:
Closes [#2610](https://github.com/carbon-design-system/carbon-components-svelte/issues/2610)

Ports the upstream React behavior of allowing a starting step value if the input is empty.


metonym:

PR #2630, opened 2026-02-10
feat(number-input)!: dispatch blur, `click:stepper` events with `value`

### LINKED ISSUE #2611: `NumberInput`: add `on:blur:stepper` event
React fires `onStepperBlur` when the stepper buttons lose focus. No equivalent in Svelte.

This should work in conjunction with #2618 (dispatched event with a matching signature).

### LINKED ISSUE #2617: `NumberInput`: on:blur includes current/parsed value
React's `onBlur` handler receives both the event and the current parsed value: `(event, value) => void`.

The Svelte component forwards the native `blur` event but doesn't provide the parsed value. Implementing this would require changing the forwarded event to be dispatched.

### LINKED ISSUE #2618: `NumberInput`: add `on:click:stepper` callback on stepper buttons
React's `onClick` fires on stepper clicks and provides `{ value, direction }`. The Svelte component dispatches a generic `click` event on the wrapper `div`, not on the stepper buttons with state info.

metonym:
Closes #2611, closes #2617, closes #2618

This aligns `NumberInput` dispatched events with the React `NumberInput` callback signatures. Consumers can now get structured payloads from stepper clicks, input blur, and stepper blur (instead of relying on raw native events or having no event at all).

**Changes**

`on:click:stepper` is a new event dispatched when a stepper button is clicked. Its `detail` carries `{ value, direction }` where `direction` is `"up"` or `"down"`. It does *not* fire on arrow-key stepping, only on actual button clicks.

```svelte
<NumberInput
  on:click:stepper={(e) => {
    console.log(e.detail.value, e.detail.direction);
  }}
/>
```

`on:blur` is now a dispatched event instead of a forwarded native `FocusEvent`. This is a **breaking change**: code that previously read the event as a `FocusEvent` directly will need to read `e.detail.event` instead.

```svelte
<NumberInput
  on:blur={(e) => {
    console.log(e.detail.value); // number | null
    console.log(e.detail.event); // FocusEvent
  }}
/>
```

`on:blur:stepper` is a new event dispatched when a stepper button loses focus. Its `detail` carries `{ event, value, direction }`, mirroring the shape of `on:click:stepper` plus the original `FocusEvent`.

```svelte
<NumberInput
  on:blur:stepper={(e) => {
    console.log(e.detail.value);     // number | null
    console.log(e.detail.direction); // "up" | "down"
    console.log(e.detail.event);     // FocusEvent
  }}
/>
```

### DIFF
--- a/docs/src/pages/components/NumberInput.svx
+++ b/docs/src/pages/components/NumberInput.svx
@@ -71,6 +71,42 @@ Use the `validate` prop to provide custom validation logic. The function receive
 
 <NumberInput validate={(raw) => { const num = Number(raw); return Number.isNaN(num) ? undefined : num % 2 === 0; }} invalidText="Must be an even number" labelText="Even number" value={0} />
 
+## Blur event
+
+Use `on:blur` to listen for when the input loses focus. The event detail includes the current parsed `value` and the original `FocusEvent`.
+
+<NumberInput
+  on:blur={(e) => {
+    console.log(e.detail.value); // number | null
+    console.log(e.detail.event); // FocusEvent
+  }}
+/>
+
+## Stepper click event
+
+Use `on:click:stepper` to listen for stepper button clicks. The event detail includes the updated `value` and the `direction` (`"up"` or `"down"`). This event only fires on button clicks, not arrow key presses.
+
+<NumberInput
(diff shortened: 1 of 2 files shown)


metonym:

Issue #2616, opened 2026-02-01
`NumberInput`: add `validate` prop


metonym:
Related #2615 

React supports a `validate` function `(value: string, locale: string) => boolean | undefined` that hooks into the built-in validation pipeline. The Svelte component has no custom validation callback. This is closely tied to the locale/`type="text"` work.

metonym:
Released in [v0.100.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.100.0)

Issue #2615, opened 2026-02-01
`NumberInput`: add `locale` prop with `Intl.NumberFormat` support


metonym:
React supports a `locale` prop (BCP47 language code) used with `type="text"` for locale-aware number parsing and formatting (e.g., `1.234,56` in German). The Svelte component has no locale support. This would also encompass the React `formatOptions` prop for `Intl.NumberFormat` options.

This would properly address #2547 instead of a manual, hardcoded workaround like #2607

metonym:
Released in [v0.100.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.100.0)

Issue #2614, opened 2026-02-01
`NumberInput`: auto min/max validation


metonym:
React automatically computes `invalid` state from `min`/`max` bounds via `getInputValidity()`: the input visually shows invalid if the value exceeds bounds even without `invalid={true}`. The Svelte component requires manually setting the `invalid` prop; it does not auto-validate against `min`/`max`.

metonym:
Released in [v0.100.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.100.0)

PR #2631, opened 2026-02-12
feat(number-input): add validation with auto min/max, locale formatting, `validate` prop

metonym:
Closes [#2614](https://github.com/carbon-design-system/carbon-components-svelte/issues/2614), closes [#2615](https://github.com/carbon-design-system/carbon-components-svelte/issues/2615), closes [#2616](https://github.com/carbon-design-system/carbon-components-svelte/issues/2616)

This adds a validation pipeline to `NumberInput`: auto min/max validation, locale-aware formatting via `Intl.NumberFormat`, and a custom `validate` callback. This brings the component closer to parity with React Carbon's `NumberInput` behavior.

**Changes**

- **Auto min/max validation**: values outside `min`/`max` bounds now automatically trigger the invalid visual state (red border, icon, `aria-invalid`). Previously, consumers had to manually set `invalid={true}` even when bounds were defined. The visual state fires without requiring `invalidText`; if `invalidText` is provided, the error message is also shown. Internally, `hasError` is replaced by `effectiveInvalid` (visual state) and `hasErrorMessage` (error text display).
- **Locale formatting**: new `locale` and `formatOptions` props. When `locale` is set, the input switches to `type="text"` with `inputmode="decimal"` and formats the display value on blur using `Intl.NumberFormat`. A `userInputActive` flag prevents the reactive sync from reformatting during typing; formatting only applies on blur, stepper clicks, and programmatic changes.

  ```svelte
  <NumberInput locale="de-DE" value={1234.5} labelText="Amount" />
  <NumberInput locale="en-US" formatOptions={{ minimumFractionDigits: 2 }} value={1234.5} labelText="Price" />
  ```

- **Custom validation**: new `validate` prop accepting `(value: string, locale: string | undefined) => boolean | undefined`. The priority chain is: explicit `invalid` prop > `validate()` returning `false` > `validate()` returning `true` (overrides auto bounds) > `autoInvalid`.

  ```svelte
  <NumberInput validate={(raw) => Number(raw) % 2 === 0} invalidText="Must be even" />
  ```

### DIFF
--- a/docs/src/pages/components/NumberInput.svx
+++ b/docs/src/pages/components/NumberInput.svx
@@ -29,7 +29,7 @@ Add descriptive text below the input using the `helperText` prop. This helps use
 
 ## Minimum and maximum values
 
-Constrain input values using `min` and `max` props. This example limits values between 4 and 20.
+Constrain input values using `min` and `max` props. When the value falls outside the specified range, the input automatically shows an invalid visual state. Provide `invalidText` to display an error message.
 
 <NumberInput min="{4}" max="{20}" value="{4}" invalidText="Number must be between 4 and 20." helperText="Clusters provisioned in your region" labelText="Clusters (4 min, 20 max)" />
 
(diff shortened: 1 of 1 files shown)


metonym:

PR #2633, opened 2026-02-12
feat(dropdown): add `labelChildren` slot

metonym:
`ComboBox` and `MultiSelect` have `labelChildren` slots (along with most/all input-like components).

This adds parity for `Dropdown`.


metonym:

PR #2638, opened 2026-02-14
fix(button): prevent icon-only button tooltip from triggering wrong action

### LINKED ISSUE #937: Button tooltip triggers the wrong button
Environment:
* carbon-components-svelte 0.50.2
* Chrome browser

Reproduction:
* Add buttons `<Button kind="ghost" tooltipPosition="right">` in a row: [Add], [Delete], [Save]. They invoke corresponding functions.
* Click [Add] a bunch of times.
* Go with the mouse cursor over the [Delete] button
* Move the mouse cursor to the right, to the [Save] button, and click
* (Use this exact mouse movement)

Actual result:
* The selected item is deleted.

Expected result:
* The selected item is saved.

Importance:
* Data loss (in this case)
* The wrong button is triggered

Implementation:
* This is a regression. Earlier versions of Carbon did not have this bug.
* The browser native tooltip does not have the bug. The bug is caused by Carbon trying to implement its own tooltip, and this tooltip implementation has this bug.
* There are 2 bugs here that in combination cause the problem:
  * Moving the mouse outside of the button should remove the tooltip. Even if the mouse is on the tooltip, but no longer over the actual button, the tooltip should disappear.
  * A click on the tooltip, if outside the button, should not trigger the action. Only a click on the button itself should trigger the action.

metonym:
Fixes #937

If multiple icon-only buttons are adjacent, their tooltips can overlap. Carbon CSS sets `pointer-events: all`, so clicking an adjacent tooltip can trigger the wrong button.

This fix overrides the Carbon style with inline `pointer-events: none` and adds a shared store to ensure only one icon-only button tooltip is visible at a time.

---

https://github.com/user-attachments/assets/98e6c9ee-5328-4c32-8d47-fb039f278206

### DIFF
--- a/docs/src/pages/components/Button.svx
+++ b/docs/src/pages/components/Button.svx
@@ -110,6 +110,12 @@ Set `isSelected` to `true` to enable the selected state for an icon-only, ghost 
 
 <FileSource src="/framed/Button/SelectedIconOnlyButton" />
 
+## Adjacent icon-only buttons
+
+When multiple icon-only buttons are placed next to each other, only one tooltip will be visible at a time. Hovering a new button automatically hides the previous tooltip, ensuring the correct button is always clicked.
+
+<FileSource src="/framed/Button/AdjacentIconOnlyButtons" />
+
 ## Link button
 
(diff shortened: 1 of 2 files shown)


metonym:

PR #2639, opened 2026-02-14
feat(overflow-menu): add `portalMenu` prop

metonym:
Fixes #796

Follow-up to #2636 

Also adds menu portaling support to `OverflowMenu`, which does not use `ListBoxMenu` for its implementation.

---

Screenshot 2026-02-14 at 10 33 46 AM

Issue #2632, opened 2026-02-12
ComboBox dropdown does not work well in small Modals


brunnerh:
The Modal always scrolls overflowing content which makes ComboBoxes not very usable unless there is enough space within the Modal to accommodate the dropdown.

<img width="1080" height="207" alt="Image" src="https://github.com/user-attachments/assets/10565807-2d10-47a1-b572-eee080cb2346" />

[Playground reproduction](https://svelte.dev/playground/21858d415b8644c4b13a4eff7a7e1bab?version=5.50.2) (Has option to override some styles to let the dropdown overflow the Modal)

<details><summary>Code</summary>

```svelte
<script>
	import { Header, Content, Modal, ComboBox, Toggle } from 'carbon-components-svelte@0.99.3';

	let forceAdjustStyles = $state(false);
</script>

<svelte:head><link rel="stylesheet" href="https://unpkg.com/carbon-components-svelte@0.99.3/css/g100.css" /></svelte:head>

<Header platformName="Carbon"/>

<Content>
	<div class:force-adjust-styles={forceAdjustStyles}>
		<Modal open passiveModal>
			<Toggle
				bind:toggled={forceAdjustStyles}
				labelText="Adjust styles"
				style="margin-bottom: 16px"
			/>
			<ComboBox
			  labelText="Contact"
			  placeholder="Select contact method"
			  items={[
			    { id: "0", text: "Slack" },
			    { id: "1", text: "Email" },
			    { id: "2", text: "Fax" },
			  ]}
			/>
		</Modal>
	</div>
</Content>

<style>
	.force-adjust-styles :global {
		.bx--modal-container {
			overflow: unset;
		}
		.bx--modal-content {
			overflow-y: unset;
		}
	}
</style>
```

</details>

Not sure what the best fix would be.
`DatePicker` has a similar issue.

metonym:
Related to [portal](https://github.com/carbon-design-system/carbon-components-svelte/issues?q=is%3Aissue%20state%3Aopen%20label%3Aportal), a common theme for menus (dropdown/tooltips) that need to render outside of its containing parent node.

metonym:
This will be supported in the next release.

<img width="869" height="442" alt="Image" src="https://github.com/user-attachments/assets/3691880f-d711-4824-9898-4857163af94c" />

brunnerh:
Nice!

Do we need a separate issue for `DatePicker`?

metonym:
[...]

PR #2640, opened 2026-02-14
feat(date-picker): add `portalMenu` prop

metonym:
Related #2632, related #2639

https://github.com/carbon-design-system/carbon-components-svelte/issues/2632#issuecomment-3902330693

When `portalMenu` is true, flatpickr renders the calendar at `document.body` instead of inside the date picker wrapper. This portaling work is a follow-up to support added for combo box, dropdown, multi-select, and overflow menu.

---

Screenshot 2026-02-14 at 10 52 14 AM


metonym:

PR #2646, opened 2026-02-14
fix(date-picker): sync "to" input after programmatic range update

### LINKED ISSUE #893: DatePicker changes values after update and inputs have changed from outside of DatePicker component
Hello, I didn't know what the topic of this issue should be but I hope I was able to come up with something valuable.

The problem I have with the `DatePicker` component is that I want to be able to change the `valueFrom` and `valueTo` variables outside from the component. I use the range as a filter so I have a `Select` that the user can use to quick select months to filter on. I can change and update the binded values from the select. And everything works fine.
Until I manually change any of the inputs in the `DatePicker`, when I do that. The value binded to the input not changed get "reset" to the value it was prior to filtering on month.

I created this REPL to show what I mean:
https://svelte.dev/repl/daa77dd9fe9740fe824157a59c3adf17?version=3.44.1
To reproduce:

1. Change the select to a month
2. Change any of the inputs to a date. Doesn't matter if the date is in the selected month or not. 
3. The other input gets changed to the value it was prior to selecting a month..

metonym:
Fixes #893

The `afterUpdate` hook synced `inputRef` (from) but not `inputRefTo` (to), causing the rangePlugin to retain stale DOM values that reverted on the next user interaction.

---

https://github.com/user-attachments/assets/a5dc6a3d-e173-41ac-939b-bc909b837b89

### DIFF
--- a/docs/src/pages/components/DatePicker.svx
+++ b/docs/src/pages/components/DatePicker.svx
@@ -27,6 +27,12 @@ Set `datePickerType` to `"range"` to enable date range selection.
 
 <FileSource src="/framed/DatePicker/DatePickerRange" />
 
+## Programmatic range update
+
+Use `bind:valueFrom` and `bind:valueTo` to programmatically update the date range from outside the component.
+
+<FileSource src="/framed/DatePicker/DatePickerProgrammatic" />
+
 ## Portal menu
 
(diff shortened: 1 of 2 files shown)


metonym:

PR #2647, opened 2026-02-14
fix(ui-shell): prevent `SideNav` flicker on page load

### LINKED ISSUE #1383: SideNav from ui-shell is flickering on page reload
Hi,

I tried to copy an example of UI Shell with a sidebar as is from the official docs, and use it in my project. (`Header with side navigation` from https://carbon-components-svelte.onrender.com/components/UIShell)

But for some reason `SideNav` is flickering on a page load so there's some delay before the component is rendered on a page. When the window is minimized, the hamburger button is not loaded instantly as well.

I'm using SvelteKit, my only assumption is that it somehow related to a hydration, but it's just a wild guess. 
Here's the minimal code to reproduce the issue: https://github.com/fbjorn/svelte-carbon-sidenav-issue

And the demo of what I'm talking about:

sidenav
sidenav-collapsed

metonym:
Fixes #1383

Gate the collapsed class, `visibility: hidden`, and `isSideNavCollapsed` store on `winWidth !== undefined` so CSS media queries handle the initial display during SSR instead of JS forcing a hidden state.

### DIFF
--- a/src/UIShell/SideNav.svelte
+++ b/src/UIShell/SideNav.svelte
@@ -48,7 +48,12 @@
     dispatch(isOpen ? "open" : "close");
     prevIsOpen = isOpen;
   }
-  $: $isSideNavCollapsed = !isOpen;
+  // Only update the collapsed store after hydration (winWidth is known).
+  // During SSR, defer to Carbon CSS media queries to handle visibility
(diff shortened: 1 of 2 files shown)


metonym:

Issue #746, opened 2021-07-14
SideNavMenus don't collapse when used with SideNav rail


dqzx:
SideNavMenus should collapse when using the rail variant of the SideNav, if a child item is active then the active state should be rolled up to the group. Currently they don't collapse which causes a few issues, for example take the following menu:

![image](https://user-images.githubusercontent.com/4139931/125641181-dae78cfa-dd3b-404b-89fb-0d97f7b0ed1d.png)

1) tabbing through the closed menu is possible and messes up the SideNavMenu positioning as the drop down is hidden but still a tab target. It'd probably be reasonable to expect the sidenav to expand on tab focus, example:

![image](https://user-images.githubusercontent.com/4139931/125641386-6a81d936-5716-43e3-881e-6e1eabaeb92e.png)

2) SideNavLinks can't be seen on the menu but can still be selected, example:

![image](https://user-images.githubusercontent.com/4139931/125641762-a8896058-df5c-4139-9694-e473312719c8.png)

While the current svelte implementation does match both the current react and vue components, these issues have been fixed in the watson IoT SideNav variant that can be seen at https://carbon-addons-iot-react.com/?path=/story/watson-iot-sidenav--side-nav-component.

metonym:
Fixed in [v0.101.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.101.0)

PR #2648, opened 2026-02-15
fix(ui-shell): collapse `SideNavMenu` items when rail is closed

metonym:
Fixes [#746](https://github.com/carbon-design-system/carbon-components-svelte/issues/746)

Collapse/make inert the `SideNavMenu` for the rail variant when the rail is closed, so expanded items don't interfere with tab order.


metonym:

Issue #1464, opened 2022-08-26
`ExpandableTile` doesn't reduce click target to chevron icon


Alexey1100:
There's an issue with `ExpandableTile` logic.
Clicking on an item inside the tile collapses or expands the tile, but according to the [reference](https://carbondesignsystem.com/components/tile/usage/#expandable):

> Expandable tiles can contain internal CTAs (like links to docs) if they are given their own click targets and the click target is reduced to only the chevron icon.

Could be reproduced with:
```svelte
<script>
  import { ExpandableTile } from "carbon-components-svelte";
</script>

<ExpandableTile tileExpandedLabel="View less" tileCollapsedLabel="View more">
  <a slot="above" href="#" style="height: 10rem">Above</a>
  <a slot="below" href="#" style="height: 10rem">Below</a>
</ExpandableTile>
```

metonym:
Thanks for flagging this. This indeed is an unsupported feature.

The "interactive" Expandable Tile is supported in v11; the corresponding `--tile-expandable-interactive` classes do not exist in `carbon-components@10`.

As a workaround, if you intend for interactive content not to trigger the expand behavior, use the `stopPropagation` modifier. In the following example, clicking the link will not expand the tile and will log "Hello world" in the browser console.

[REPL](https://svelte.dev/repl/1c5339a26fc14e28a012822bffac5813?version=3.49.0)

```svelte
<script>
  import { ExpandableTile } from "carbon-components-svelte";
</script>

<ExpandableTile>
  <div slot="above" style="height: 10rem">
    Above the fold content here This is a
    <a
      href="/"
      on:click|preventDefault|stopPropagation={() => console.log("Hello world")}
    >
      Link
    </a>
  </div>
  <div slot="below" style="height: 10rem">Below the fold content here</div>
</ExpandableTile>
```

PR #2653, opened 2026-02-15
fix(expandable-tile): reduce click target to chevron when tile contains interactive content

metonym:
Fixes [#1464](https://github.com/carbon-design-system/carbon-components-svelte/issues/1464)

`ExpandableTile` is a button by default. Per the Design System guidance, nested interactive content should be allowed. To avoid invalid markup, the component should automatically detect interactive content so that the proper semantic element is used, and avoid needing to manually stop event propagation on nested, interactive elements.


brunnerh:
Maybe there could also be a property to force it into either state; having to re-create the tile with different elements at runtime is not ideal. (And JS might not be running for various reasons.)

metonym:
@brunnerh Fair enough, opened #2658

Issue #1362, opened 2022-06-22
DatePicker broken validation if dateFormat="Y-m-d"


rkakrik:
When set dateFormat in DatePicker to "Y-m-d" validation show error:
https://svelte.dev/repl/4180232a624e4a9abea7d5051c8a8001

jimjam-slam:
This validation error also prevents the form from submitting, so it's not just a visual problem. I'm also seeing a similar problem with TimePicker, even when I'm not specifying any format (just using the default HH:mm). Both popup with "Match the requested format" when I submit the form they're in. The DatePicker can be ameliorated by switching to `dateFormat="d/m/Y"`, but I don't have a workaround for the TimePicker.

metonym:
Fixed in [v0.101.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.101.0)

Example: https://svelte.carbondesignsystem.com/components/DatePicker#programmatic-range-update

PR #2654, opened 2026-02-15
fix(date-picker): derive input pattern from `dateFormat` prop

metonym:
Fixes [#1362](https://github.com/carbon-design-system/carbon-components-svelte/issues/1362)

Currently, a custom date input pattern does not work (not passed to input), which can cause a native input HTML validation error.

This plumbs the format through from `DatePicker` to the individual element.

```svelte
<DatePicker datePickerType="single" dateFormat="Y-m-d">
  <DatePickerInput labelText="Date of birth" placeholder="yyyy-mm-dd" />
</DatePicker>

```


metonym:

PR #2656, opened 2026-02-15
fix(overflow-menu): `OverflowMenuItem` has explicit `type="button"` to prevent form submission

metonym:
Identified via linting in #2655 

The overflow menu item uses a button for non-links. It should have an explicit `type="button"` to avoid accidental form submission (implicitly `"submit"` when used inside a form).


metonym:

PR #2657, opened 2026-02-15
fix(expandable-tile): prevent height animation on initial load

metonym:
On initial load, `ExpandableTile` has an unexpected height animation.

The issue is that `max-height` starts at 0 before DOM measurement, which triggers the Carbon CSS transition when the height is actually measured. The animation should not apply if the max height is not measured yet.


metonym:

PR #2658, opened 2026-02-15
feat(expandable-tile): add `hasInteractiveContent` prop

metonym:
Fixes https://github.com/carbon-design-system/carbon-components-svelte/issues/1464

Refactors the approach from #2653 to better support nested interactive content in `ExpandableTile`.

Instead of auto-detection at runtime, expose an explicit prop to render the tile as a div instead of button. There is some prior art: `Tag` has an interactive prop to enable interactive semantics.


metonym:

PR #2659, opened 2026-02-15
feat(stack): add `inline` prop for `inline-flex` display

metonym:
Adds an `inline` prop to `Stack` to enable inline-flex usage.

---

Screenshot 2026-02-15 at 10 57 39 AM


metonym:

PR #2662, opened 2026-02-16
feat: `RadioButton` and `RadioButtonGroup` selected value is generic

metonym:
Improves TypeScript DX by making the `selected` value generic, defaulting to `string | number`.


metonym:

PR #2664, opened 2026-02-16
feat(structured-list): `selected` value is generic

metonym:
The `selected` value should be generic.


metonym:

PR #2666, opened 2026-02-16
feat(radio-tile): `value` is generic

metonym:
Make `value` generic.


metonym:

PR #2669, opened 2026-02-16
fix(portal): transfer focus to active element

metonym:
The portal should restore the focus to its portaled element. For example, overflow menu should focus the first menu item, and return focus back to its trigger.


metonym:

PR #2673, opened 2026-02-16
fix(modal): primary button should receive focus by default

### LINKED ISSUE #2671: Initial focus in Modal
Currently, the close button is focused, but according to [the spec](https://carbondesignsystem.com/components/modal/usage/#focus), the first input or the primary button should be focused instead.

> **Focus**\
> Once the modal is open, set the initial focus to the first location that accepts user input. For example, if the modal contains a form then the focus should automatically be set to the first field when opened. If it is a transactional modal without form inputs in the body section then the first focus should be on the primary button.

metonym:
Fixes #2671 

Unless `selectorPrimaryFocus` is specified, the primary button is focused by default. For passive modal (where buttons do not exist), the close button should be the fallback.

### DIFF
--- a/docs/src/pages/components/Modal.svx
+++ b/docs/src/pages/components/Modal.svx
@@ -1,4 +1,5 @@
 <script>
+  import { OrderedList, ListItem } from "carbon-components-svelte";
   import Preview from "../../components/Preview.svelte";
 </script>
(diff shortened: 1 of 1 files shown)


metonym:

PR #2674, opened 2026-02-16
fix(modal): return focus to trigger

metonym:
This fixes an accessibility issue.

When a modal / composed modal is closed, it should return focus to the trigger element, if valid. This allows unbroken keyboard navigation.

---

https://github.com/user-attachments/assets/ab79d710-6d2c-4705-b849-dc738dad591e


metonym:

PR #2675, opened 2026-02-16
fix(modal): auto-focus first input for transactional dialogs

metonym:
Closes [#2671](https://github.com/carbon-design-system/carbon-components-svelte/issues/2671), follow-up to #2673

This adds automatic focus on the first input when a modal contains form fields, per [Carbon's accessibility spec](https://carbondesignsystem.com/components/modal/accessibility) for transactional dialogs.

Currently, you must explicitly set `selectorPrimaryFocus` or add `data-modal-primary-focus` to achieve this. Now the first focusable input/textarea/select is tried before falling back to the primary or close button, improving the default DX.

**Focus order (updated)**

1. Element matching `selectorPrimaryFocus` (default: `[data-modal-primary-focus]`)
2. First input, textarea, or select
3. Primary button if transactional, secondary button (cancel) if danger.
4. Close button

### DIFF
--- a/docs/src/pages/components/ComposedModal.svx
+++ b/docs/src/pages/components/ComposedModal.svx
@@ -29,6 +29,14 @@ The modal dispatches a cancelable `close` event, allowing you to prevent the mod
 
 <FileSource src="/framed/Modal/ComposedModalPreventClose" />
 
+## Danger modal
+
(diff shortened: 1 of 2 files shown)


brunnerh:
One more thing, the part where the inputs are mentioned is not in the accessibility section, but I just saw something in there which was missing from the usage section:

> For destructive interactions, the “cancel” button takes focus, not the red danger/delete button.

metonym:

Issue #2687, opened 2026-02-19
Tooltip should show on hover


metonym:
### Discussed in https://github.com/carbon-design-system/carbon-components-svelte/discussions/1802

<div type='discussions-op-text'>

<sup>Originally posted by **geoidesic** September  1, 2023</sup>
https://carbon-components-svelte.onrender.com/components/Tooltip

It seems ironic that the tooltips on the docs page (like for the Copy Code icon) work the way tooltips are expected to work on a desktop browser (i.e. display on hover) but the Tooltip component that the docs describe doesn't seem to offer that functionality :)

Currently I'm not sure how this component is different from a button?

I can appreciate that on click is useful and necessary for mobile devices but it's not useful for desktop, where the idea is that a tooltip provides information about the element you are hovering over.

Am I doing it wrong? Is there a way to get it to show the text on hover? 

The one on the IBM live demo works as expected: https://carbondesignsystem.com/components/tooltip/usage/#live-demo</div>

brunnerh:
This is a v10/v11 difference as I mentioned here before:
- #1805

(Not sure how you want to handle the version stuff.)

metonym:
Thanks for the context! I think this should be a fix. I was astonished that hover for Tooltip was not already implemented, as it's a very basic expectation. And as you mentioned, Toggltip is explicitly a "click" / button behavior.

metonym:
Fixed in [v0.101.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.101.1)

PR #2688, opened 2026-02-20
fix(tooltip): show tooltip on hover

metonym:
Fixes [#2687](https://github.com/carbon-design-system/carbon-components-svelte/issues/2687)

Per the design spec, tooltips should show on hover, not just click/focus. A toggle tip only supports the click interaction.

---

https://github.com/user-attachments/assets/9c769ee1-edd2-401f-aba0-1551a57a8582


metonym:

Issue #2689, opened 2026-02-20
DatePicker derives invalid patterns for given dateFormat


spburtsev:
`DatePicker` component derives an invalid pattern for custom `dateFormat` property, which prevents form submission.
The docs give an example `dateFormat="Y-m-d"` which does not seem to work with the derived pattern.

Svelte version: 5

Reproduction sandbox: https://stackblitz.com/edit/vitejs-vite-rwyfqevg?file=src%2FApp.svelte

<img width="353" height="494" alt="Image" src="https://github.com/user-attachments/assets/c4a44d90-c82d-4c71-a61f-db6fb6552089" />

metonym:
Fixed in [v0.101.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.101.1)

Updated example: https://svelte.carbondesignsystem.com/components/DatePicker#custom-date-format

PR #2690, opened 2026-02-20
fix(date-picker): don't escape hyphens in derived input pattern

metonym:
Fixes [#2689](https://github.com/carbon-design-system/carbon-components-svelte/issues/2689)

Follow-up to #2654; don't escape hyphens in the derived input pattern.


metonym:

PR #2700, opened 2026-02-23
fix(dropdown): menu should close if `portalMenu` is true

### LINKED ISSUE #2699: Dropdown menu does not close after item selection in Portal mode
I noticed this issue after upgrading to v0.101.0, which caused several of my Playwright tests to fail (Dropdown inside modal).

You can reproduce the bug on the documentation page: [Dropdown / Portal menu](https://carbon-components-svelte.onrender.com/components/Dropdown#portal-menu)

metonym:
Fixes #2699

`Dropdown` with `portalMenu` does not close the dropdown after selection.

### DIFF
--- a/src/Dropdown/Dropdown.svelte
+++ b/src/Dropdown/Dropdown.svelte
@@ -583,6 +583,7 @@
                     }
                     selectedId = item.id;
                     dispatchSelect();
+                    open = false;
                     ref.focus();
                   }}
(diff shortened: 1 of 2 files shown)


metonym:
Fixed in [v0.101.2](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.101.2)

Issue #2701, opened 2026-02-25
UIShell Content margin problem with fixed SideNav


szegheo:
I'm using the `UIShell` component and my `SideNav` is created with the `fixed` prop, since I need it always to be visible, no matter how small the window is.

Currently, when the window size goes under the breakpoint (1056px):

* the "Close menu" button appears, but does nothing
  (I think it should remain hidden when the SideNav is fixed)
* the main content receives a `style="margin-left: 0px;"` attribute ([code](https://github.com/carbon-design-system/carbon-components-svelte/blob/v0.101.2/src/UIShell/Content.svelte#L26))
   (I think it's wrong, because the SideNav is `fixed` and now the content slips under)

I made a minimal Svelte Repl to showcase the current behaviour:
https://svelte.dev/playground/2f89c91e14b442029d5564da43daaf5c?version=latest

<img width="938" height="622" alt="Image" src="https://github.com/user-attachments/assets/430d14d3-f2d8-442f-8c02-a77eb1cddb8e" />

___

For a quick workaround, I added the `expansionBreakpoint={0}` prop to my SideNav and made some custom CSS to always hide the close/open menu.

metonym:
Thank you for reporting. Fixed in [v0.101.3](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.101.3)

A fixed `SideNav` will not unset the left margin, nor will the hamburger menu be rendered.

Example: https://svelte.carbondesignsystem.com/components/UIShell#header-with-fixed-side-navigation

PR #2702, opened 2026-02-25
fix(ui-shell): fixed `SideNav` should not render hamburger menu or unset left margin

metonym:
Fixes [#2701](https://github.com/carbon-design-system/carbon-components-svelte/issues/2701)

The fixed side nav should not unset the left margin, nor should the hamburger menu be rendered in the header.

Screenshot 2026-02-25 at 11 16 22 AM


metonym:
Fixed in [v0.101.3](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.101.3)

PR #2704, opened 2026-02-26
fix(ui-shell): remove incorrect role="menu" from `SideNavMenu`

b-r-i-a-n-w-e-s-t:
aria `menu` role on SideNavMenu was getting flagged in automated accessibility testing because it's descendants do not have a compatible role (e.g. `menuitem`). This list serves as navigation, is wrapped in `<nav>`, and does not need a menu role or enhanced aria markup at all. 

https://web.dev/articles/website-navigation#navigations_versus_menus

### DIFF
--- a/src/UIShell/SideNavMenu.svelte
+++ b/src/UIShell/SideNavMenu.svelte
@@ -53,9 +53,7 @@
       <ChevronDown />
     </div>
   </button>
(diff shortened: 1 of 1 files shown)


metonym:
Fixed in [v0.101.4](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.101.4)

PR #2705, opened 2026-02-26
perf: hoist regexes

metonym:
Hoist regexes used in `DatePickerInput` and `DataTable`, and add missing unit tests.


metonym:
Fixed in [v0.101.4](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.101.4)

PR #2709, opened 2026-02-28
feat(data-table): allow custom expand icon

metonym:
Closes [#2696](https://github.com/carbon-design-system/carbon-components-svelte/issues/2696)

Allow the expandable row icon to be customized via a slot/snippet.

---

https://github.com/user-attachments/assets/c65ade11-590f-4824-9991-133187665348

PR #2710, opened 2026-03-01
feat(data-table)!: rename `expanded-row` slot to `expandedRow` for Svelte 5 snippet support

metonym:
Stacked on #2709

The `expanded-row` slot should be renamed so that it can be used using Svelte 5 snippet syntax.


metonym:
Released in [v0.102.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.102.0)

Issue #2694, opened 2026-02-22
Tag component should support clicking + close event


metonym:
### Discussed in https://github.com/carbon-design-system/carbon-components-svelte/discussions/1953

<div type='discussions-op-text'>

<sup>Originally posted by **xiatian12345** April 17, 2024</sup>
When add close event,the click event is auto disabled,how should i do to retain the click and close event?</div>

metonym:
Fixed in [v0.102.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.102.0)

PR #2711, opened 2026-03-01
fix(tag)!: filterable tag forwards `click` event to tag body, not close button

metonym:
Fixes [#2694](https://github.com/carbon-design-system/carbon-components-svelte/issues/2694)

An interactive (filterable) tag has a close button. Click events should still be valid (clicking on the body of the tag), since the close button is separate.

This is technically a breaking change but a necessary behavioral fix, where the click event would fire only on the close button. Instead, this adds a distinct close event, while the click should apply to the tag body.


metonym:
Fixed in [v0.102.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.102.0)

Issue #2697, opened 2026-02-22
DataTable allows disabling the "none" sort direction


metonym:
### Discussed in https://github.com/carbon-design-system/carbon-components-svelte/discussions/1746

<div type='discussions-op-text'>

<sup>Originally posted by **ibilux** June 10, 2023</sup>
Is there a way to disable the null `sortKey` and none `sortDirection` in a sortable DataTable, and use only `descending` and `ascending` values ?</div>

metonym:
Released in [v0.102.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.102.0)

PR #2712, opened 2026-03-01
feat(data-table): support `sortAlways`

metonym:
Closes [#2697](https://github.com/carbon-design-system/carbon-components-svelte/issues/2697)

Sortable DataTable columns currently cycle through three states on click: `none` (unsorted) → `ascending` → `descending` → `none`. A third click on the same column unsorts and restores the original row order. For tables where a sort should always be active, this behavior is undesirable: users can accidentally unsort and lose the intended view.

The new `sortAlways` prop changes the direction map so `descending` cycles back to `ascending` instead of `none`. The first click still goes to ascending; subsequent clicks alternate between ascending and descending. The column never unsorts.

**Usage**

Table-level:

```svelte
<DataTable sortable sortAlways {headers} {rows} />
```

Per-column override (header-level `sortAlways` overrides the table default):

```svelte
<DataTable
  sortable title="Recent transactions"
  headers="{[
    { key: 'date', value: 'Date', sortAlways: true },
    { key: 'description', value: 'Description' },
    { key: 'amount', value: 'Amount' },
    { key: 'status', value: 'Status', sortAlways: false }
  ]}"
/>
```

In the example above, Date always stays sorted (chronological order for audit trails); Status can be unsorted.

### DIFF
--- a/docs/src/pages/components/DataTable.svx
+++ b/docs/src/pages/components/DataTable.svx
@@ -946,6 +946,87 @@ Set `sortable` to `true` to enable column sorting. Disable sorting on specific c
   ]}"
 />
 
+## Sortable (sort always)
+
+Sortable columns support three directions: `none` (unsorted, original row order), `ascending`, and `descending`. Clicking a column header cycles through them in order.
+
+By default, a third click on a sorted column unsorts it (`none` restores the original row order).
+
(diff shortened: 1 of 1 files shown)


metonym:
Released in [v0.102.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.102.0)

PR #2713, opened 2026-03-07
fix(tree-view): use correct leaf depth to restore active node style

metonym:
This fixes a visual regression where the active node's left outline is cut off.

The issue was that `computeTreeLeafDepth(refLabel)` counts the node's own `<li>` wrapper as a depth level, so root nodes returned depth=1 instead of 0. This made every node's indentation 1rem too large (e.g., root: 3.5rem instead of 2.5rem).

The fix subtracts 1 from the depth in both `TreeViewNode.svelte` and `TreeViewNodeList.svelte` offset calculations, so:
- **Root leaf node**: 0 + 2.5 = **2.5rem** (was 3.5)
- **Root parent node**: 0 + 1 = **1rem** (was 2)
- Each nesting level still adds 1rem as expected

---

## Before
Screenshot 2026-03-07 at 9 54 53 AM

## After
Screenshot 2026-03-07 at 9 50 04 AM

### DIFF
--- a/src/TreeView/TreeViewNode.svelte
+++ b/src/TreeView/TreeViewNode.svelte
@@ -87,7 +87,7 @@
     focusNode,
   } = getContext("carbon:TreeView");
   const offset = () =>
-    computeTreeLeafDepth(refLabel) + (leaf && icon ? 2 : 2.5);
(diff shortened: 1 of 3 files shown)


metonym:
Fixed in [v0.102.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.102.0)

PR #2718, opened 2026-03-07
chore: use kebab-case for utilities

metonym:


metonym:
This will likely break `carbon-preprocess-svelte` usage. Defer for now.

PR #2750, opened 2026-03-13
fix(inline-notification): title uses `strong` element for emphasis

b-r-i-a-n-w-e-s-t:
Wrapping the title slot with a `<strong>` so that bolded text is conveyed semantically. 

This should help with [WCAG 1.3.1 "Info and Relationships (Level A)"](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships) by using [technique H49 "Using semantic markup to mark emphasized or special text" ](https://www.w3.org/WAI/WCAG21/Techniques/html/H49)


metonym:
Thank you.

Issue #1603, opened 2022-12-30
PasswordInput causes overflow-x in Modal


M4TY:
Hey, I have ran into an issue when using the PasswordInput component.  When using it in a Modal, the component causes some weird overflow on x. It is not caused by normal TextInput, thus, I believe there are some margins set in the PasswordInput.

PasswordInputs:
![image](https://user-images.githubusercontent.com/35110087/210052225-cf218b1d-b64d-4160-8b2e-b57bdc73498a.png)

TextInputs:
![image](https://user-images.githubusercontent.com/35110087/210052267-757b73ec-ed50-4e5c-a28d-d6a2140c988f.png)

Is there any way to fix this? Or is there something I am doing wrong? This is the code used:

```
<script>
    import {Modal, PasswordInput} from "carbon-components-svelte";
    import {set} from "../stores/locales/set.ts";

    export let open = false;
</script>

<Modal
    bind:open
    modalHeading={set.changePasswordTitle}
    primaryButtonText={set.changePasswordConfirm}
    secondaryButtonText={set.changePasswordCancel}
    selectorPrimaryFocus="#old-pass"
    on:click:button--secondary={() => (open = false)}
    on:open
    on:close
    on:submit
>
    <br>
    <PasswordInput
        id="old-pass"
        inline
        labelText={set.changePasswordOldPassword}
        placeholder={set.changePasswordOldPasswordPlaceholder}
    />
    <br>
    <PasswordInput
        id="new-pass"
        inline
        labelText={set.changePasswordOldPassword}
        placeholder={set.changePasswordOldPasswordPlaceholder}
    />
    <br>
    <PasswordInput
        id="new-pass-repeat"
        inline
        labelText={set.changePasswordOldPassword}
        placeholder={set.changePasswordOldPasswordPlaceholder}
    />
</Modal>
```

metonym:
Can you try adding the `hasForm` prop on the `Modal`?

M4TY:
Hasn't changed anything.

brunnerh:
This is caused by tooltips, which can happen just about everywhere, e.g. action buttons at the end of table rows.

The usual fix is to use a tooltip alignment of `"end"`.

```svelte
<PasswordInput tooltipAlignment="end"/>
```

![image](https://user-images.githubusercontent.com/834235/216752696-ffdbc69f-0cc4-4207-aeb7-36fb226df546.png)

metonym:
[...]

Issue #545, opened 2021-02-28
Tooltip is part of layout calculation


brunnerh:
For example in password fields there is this button on the far right which has a tooltip that can break the layout by causing the document content width to be larger than it is supposed to be:

![image](https://user-images.githubusercontent.com/834235/109414242-35e0da00-79b2-11eb-9162-08ef16b86610.png)

Tooltips should not affect the layout of the page like this; the page becomes horizontally scrollable just because of this tooltip. If there is not enough space to one side because of the document boundaries, the tooltips should be shifted to the side. This applies to both the left and right edge of the document. Of course the little arrow would still have to be centered on the object the tooltip is applied to.

The React implementation has the same issue, but I doubt that this could be fixed by simply adjusting the underlying base styles.

brunnerh:
Carbon appears to support hard alignments. In the case of the show/hide button in the password input the problem could be fixed by adding the `bx--tooltip--align-end` class to the button. Not particularly elegant in my opinion, but it works.

metonym:
You can use the `tooltipAlignment` and `tooltipPosition` props to customize the orientation of the visibility toggle:

```svelte
<FluidForm>
  <TextInput
    labelText="User name"
    placeholder="Enter user name..."
    required />
  <PasswordInput 
    tooltipAlignment="start"
    tooltipPosition="left"
    required
    type="password"
    labelText="Password"
    placeholder="Enter password..."
  />
</FluidForm>
```

brunnerh:
@metonym As noted, i do not think tooltips should affect layout and normally they should simply be bounded to the current dimensions of the page.

Having the option to manually set alignments to get an even nicer result is great, but the library should take take care of the basic task of not accidentally breaking the layout.

brunnerh:
My suggestion would be to constrain the tooltip to the governing scrolling/clipping container, or the document body, if no such container is found. This could be done by first calculating the desired layout and then offsetting the tooltip, if any part is out of bounds.

[REPL sketch](https://svelte.dev/repl/1d007a522a9449769244887be946351a?version=3.47.0)

![image](https://user-images.githubusercontent.com/834235/164083126-c32945a3-d2df-485c-8863-d5083f6431ff.png)

One possible optimization would be to flip the tooltip to above the target element, when it exceeds the lower bound. In any case, it would be nice if something could be done about this, as this is a recurring problem. Unfortunately there seems to be no easy, pure CSS solution for this (e.g. using `position: sticky` somehow).

[...]

Issue #386, opened 2020-11-03
Possible CSS Glitch with CodeSnippet


ispyinternet:
I would say there is a possible CSS glitch here when mutliple inline `<CodeSnippet>` are used, maybe the issue is in the parent CSS library thought?

<img width="186" alt="Screenshot 2020-11-04 at 18 03 41" src="https://user-images.githubusercontent.com/1741823/98023708-3cde4e00-1dff-11eb-9b0d-de883b7fd3f8.png">

metonym:
The Carbon React implementation uses JS to manipulate the position of the tooltip in the DOM so that it is as high as possible. Tooltips are implemented differently in this library; they are contained by their parent element, which may cause overlay issues.

The solution would be to elevate the z-index of the element containing the code snippet (or tooltip).

PR #2751, opened 2026-03-14
fix(password-input): portal tooltip inside modal context

metonym:
Fixes [#545](https://github.com/carbon-design-system/carbon-components-svelte/issues/545), fixes [#1603](https://github.com/carbon-design-system/carbon-components-svelte/issues/1603)

`FloatingPortal` was added to portal menus outside of their parent context to avoid overflow constraints.

This enhances the floating portal to support similar portaling for tooltips, starting with `PasswordInput`. Similar to other dropdown-like components, the tooltip is automatically portaled into a modal context, with the option to opt in or out. Additionally, some custom style overrides have to be added to support the portalled tooltip behavior.

This should also be applied to the copy button (issue #386, previous attempt #2652)

---

Screenshot 2026-03-14 at 11 46 12 AM


metonym:

PR #2755, opened 2026-03-16
feat(combo-box): add `selectTextOnFocus` prop

metonym:
Add the ability to select the input text on focus.

---

Screenshot 2026-03-16 at 12 01 20 PM


metonym:

PR #2756, opened 2026-03-16
feat(multi-select): support `isSelectAll` item option

metonym:
Support a select all option for `MultiSelect`.

If `item.isSelectAll` is true, the option will be sorted first. It has distinct styles, and will select all non-disabled options. The select all option is sorted first in the menu.

---

Screenshot 2026-03-16 at 12 49 48 PM


metonym:

PR #2766, opened 2026-03-21
feat(content-switcher): add `selectionMode` prop

metonym:
For parity with the upstream implementation, add a prop that allows a manual selection mode for content switcher.


metonym:

PR #2768, opened 2026-03-21
feat(context-menu): `e.preventDefault()` prevents menu from closing

metonym:
Similar to `OverflowMenu`, support the ability to prevent the menu from closing using `e.preventDefault()`. Real-world scenario is a copy option that provides feedback when clicking, instead of immediately closing.

Screenshot 2026-03-21 at 12 27 23 PM


metonym:

PR #2771, opened 2026-03-21
fix(context-menu): auto-indent option with icon

metonym:
`indented` should be derived from whether an icon is passed in. Without this, `indented` needs to be manually specified.

---

Screenshot 2026-03-21 at 1 19 58 PM


metonym:

PR #2774, opened 2026-03-21
feat(file-uploader): improvements

metonym:
This PR tightens up `FileUploader` behavior around validation, events, list rendering, and accessible labels. The main themes are: emit the same lifecycle when the list is cleared whether you call `clearFiles()`, set `files` to `[]`, or remove the last row; surface oversize and duplicate picks via a dedicated `rejected` event; and make row-level icon labels configurable per file (including optional function props) with sensible defaults in `Filename`.

**Changes**

- `rejected` event with `{ file, reason: "size" | "duplicate" }` when files fail `maxFileSize` or duplicate checks.
- `preventDuplicate` prop to block new `File` objects that match an existing row (name + size + `lastModified`), without rejecting the browser’s re-listed originals.
- `change` + `clear` fired when the list becomes empty from `clearFiles()`, `files = []`, or removing the last row (aligned with programmatic clear).
- File list `{#each}` uses stable keys (`getFileId` includes index) and removes rows by `File` reference instead of index (duplicate filenames, middle-row delete).
- `iconDescription` optional with `Filename` defaults (`uploading` / `Remove file` / `Upload complete`); supports `string | (ctx) => string | undefined` with `{ file?, fileName, status, invalid }`.
- `FileUploader` / `FileUploaderItem` pass `fileName` (and `file` from `FileUploader` only) into `Filename` for the callback context.

### DIFF
--- a/docs/src/pages/components/FileUploader.svx
+++ b/docs/src/pages/components/FileUploader.svx
@@ -83,7 +83,17 @@ This example accepts multiple JPEG files and displays them in a completed state.
 
 Set `iconDescription` to a function for accessible labels that depend on each row. The callback receives `{ file, fileName, status, invalid }`. On `FileUploader`, `file` is that row’s `File` object; on `FileUploaderItem`, `file` is always `undefined` and `fileName` matches the `name` prop.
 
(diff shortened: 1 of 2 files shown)


metonym:

PR #2777, opened 2026-03-22
fix(tooltip-icon): only one tooltip is visible at a time

metonym:
Related #2638

Only one tooltip should be visible at a time. This avoids the papercut where mousing over a right-aligned tooltip unintentionally keeps it open.

---

## Before

https://github.com/user-attachments/assets/8abb96b5-d986-4d32-a70c-57d18bb838d4

## After

https://github.com/user-attachments/assets/db5d6cbb-ab6e-4af1-b2d4-77443b0b90e5


metonym:

PR #2778, opened 2026-03-22
feat(skeleton-placeholder): add size/width/height props

metonym:
Make `SkeletonPlaceholder` more ergonomic by making it easier to set size/width/height. Currently, you need to pass a raw, inline style string, which is difficult to read and requires more effort by the user.


metonym:

PR #2779, opened 2026-03-22
feat(skeleton-icon): add `SkeletonIcon` component

metonym:
The React implementation has `IconSkeleton`: https://react.carbondesignsystem.com/?path=/docs/components-skeleton-skeletonicon--overview

This adds a similar utility component, which is just a wrapper around `SkeletonPlaceholder`, now that it supports a `size` prop


metonym:

PR #2753, opened 2026-03-14
feat(tooltip): portal tooltip inside modal context

metonym:
Related #2751, #2752

Apply the same tooltip portaling behavior to the `Tooltip` component.

FloatingPortal is extended to support it: `direction` now accepts `"left"` and `"right"` with flip behavior, and new gap/alignment props for fine-grained layout control (`gapTop`, `gapBottom`, `horizontalGapLeft`, `horizontalGapRight`, `verticalAlignOffsetLeft`, `verticalAlignOffsetRight`).

---

Screenshot 2026-03-14 at 1 24 08 PM

PR #2780, opened 2026-03-22
feat(tooltip): portal `TooltipDefinition` and `TooltipIcon` inside modals

metonym:
Similar to #2753, also portal `TooltipDefinition` and `TooltipIcon` inside modals, and add a `portalTooltip` prop to allow opt-in or opt-out behavior.

---

Screenshot 2026-03-22 at 4 18 45 PM


metonym:

PR #2783, opened 2026-03-23
feat(tooltip): add enter/leave delay for `TooltipIcon` and `TooltipDefinition`

metonym:
Similar to `Tooltip` (#2688), implement a similar hover-delay trigger to avoid flashing the tooltip for `TooltipIcon` and `TooltipDefinition`. For `TooltipIcon`, rapidly moving between tooltips (if one is already open), should skip the delay.


metonym:

PR #2786, opened 2026-03-23
feat(theme): set `color-scheme` on root element

metonym:
`Theme` should also set `color-scheme` (impacting browser default scrollbar colors) based on theme classification.


metonym:

PR #2787, opened 2026-03-23
fix(ui-shell): header action button style fixes

metonym:
Ensure consistency with the spec, and enough specificity.


metonym:

PR #2796, opened 2026-03-26
 fix(side-nav-menu): `button` should not nest `div` elements

b-r-i-a-n-w-e-s-t:
div is flow content and not valid inside button elements. Need to use [phrasing content](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content) per the HTML spec.


metonym:
Thank you.

PR #2797, opened 2026-03-27
fix(data-table): custom sort accounts for sort direction

GaboUCR:
The sorting function for carbon components datatable is not working for the descending order. I tested this in the documentation example and it doesn't work as well:

https://svelte.carbondesignsystem.com/components/DataTable#sortable-with-custom-display-and-sort-methods

image

❗ This error only happend for custom sorting methods


metonym:
Thank you.

PR #2798, opened 2026-03-28
fix(toolbar-search): remove redundant `<any>` from Row generics

metonym:
`DataTableRow` ID already defaults to `any`.


metonym:

PR #2808, opened 2026-04-01
fix(modal): use mouse events for outside click

### LINKED ISSUE #2805: Modal closing on mouse button release (Chromium based browsers)
The modal closes on a mouse button release (if the mouse click action started on a modal and was then released behind its border).
It makes the workflow pretty annoying, especially when dealing with other elements inside of a modal (like a text area/field, slider, etc. which require some mouse dragging interaction (you start selecting a text in an input inside of a modal, and when the click is released outside of a modal, it closes)
There is a way to disable closing on outside click, but it's not what we want.
The issue is reproducible in Chrome. Firefox works well

A fix proposal:
Replace the on:click events with on:mouseup/down in a Modal.svelte events that handle the didClickInnerModal

diff --git a/src/Modal/Modal.svelte b/src/Modal/Modal.svelte
index 0a80d7e2..588c7ab9 100644
--- a/src/Modal/Modal.svelte
+++ b/src/Modal/Modal.svelte
@@ -263,7 +263,7 @@
     }
   }}
   on:click
-  on:click={() => {
+  on:mouseup={() => {
     if (!didClickInnerModal && !preventCloseOnClickOutside) {
       close("outside-click");
     }
@@ -293,7 +293,7 @@
     class:bx--modal-container--xs={size === "xs"}
     class:bx--modal-container--sm={size === "sm"}
     class:bx--modal-container--lg={size === "lg"}
-    on:click={() => {
+    on:mousedown={() => {
       didClickInnerModal = true;
     }}
   >

metonym:
Fixes #2805 

Fixes a UI glitch where click-dragging in an interactive element like an input can trigger the click-outside modal detection logic, causing the modal to unexpectedly close.

This fixes it to use mouseup/mousedown events.

### DIFF
--- a/e2e/composed-modal.test.ts
+++ b/e2e/composed-modal.test.ts
@@ -44,6 +44,47 @@ test.describe("ComposedModal", () => {
     await expect(primaryFocus).toBeFocused();
   });
 
+  test("does not close when mousedown inside modal and mouseup outside", async ({
+    page,
+  }) => {
+    await page.getByTestId("open-modal").click();
+    const modal = page.locator(".bx--modal");
+    await expect(modal).toHaveClass(/is-visible/);
+
+    const container = page.locator(".bx--modal-container");
+    const containerBox = await container.boundingBox();
+    const modalBox = await modal.boundingBox();
+    if (containerBox == null || modalBox == null) {
+      throw new Error("expected modal container and modal bounding boxes");
+    }
+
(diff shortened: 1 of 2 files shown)


metonym:
Fixed in [v0.105.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.105.1)

PR #2813, opened 2026-04-04
fix(dropdown): selected item has checkmark icon

metonym:
Not sure how this was missed for so long.

The selected `Dropdown` item should have a checkmark icon. This greatly helps visually; before, a slight background color difference was the only way to differentiate the selected item.

...

Screenshot 2026-04-04 at 10 56 22 AM


metonym:

PR #2814, opened 2026-04-04
feat(dropdown): `selectedId` is optional

metonym:
Previously, `Dropdown` required a `selectedId` value (one of `items["id"]`). However, as reflected in the upstream React implementation, a `Dropdown` can have no initial selection state.

If the dropdown has no selected item, the `label` prop text is displayed.

---

Screenshot 2026-04-04 at 11 01 58 AM


metonym:

PR #2815, opened 2026-04-04
feat(theme): support rendering as a `Dropdown`

metonym:
Support rendering a `Dropdown`, and also use the styled dropdown instead of a native select in docs to better match the other component-level options.


metonym:

PR #2818, opened 2026-04-04
feat(tag): support large size

metonym:
`Tag` supports a large size.

---
Screenshot 2026-04-04 at 1 07 45 PM


metonym:

PR #2820, opened 2026-04-04
perf(data-table): remove unused `thKeys` computation

metonym:
`thKeys` does not need derivation, and can be removed.


metonym:

PR #2821, opened 2026-04-04
perf(data-table): rebuild `tableCellsByRowId` for changed headers only

metonym:
`tableCellsByRowId` rebuilds the entire cell map whenever rows or headers change, even if only one row was added/modified. For large tables (e.g., 10k rows × 10 columns), this means 100k object allocations on every update.

**Approach**

Use a Map keyed by row ID to track the source row reference alongside cached cells. On update:

1. If headers changed: full rebuild (cell structure depends on headers)
2. If only rows changed: reuse cached cells for rows whose object reference is unchanged

### DIFF
--- a/src/DataTable/DataTable.svelte
+++ b/src/DataTable/DataTable.svelte
@@ -446,17 +446,30 @@
   let tableCellsByRowId = {};
   let prevRows;
   let prevHeaders;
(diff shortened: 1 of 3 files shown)


metonym:

PR #2822, opened 2026-04-04
perf(data-table): avoid virtualized scroll set-up after any reactive update

metonym:
The afterUpdate callback in `DataTable` sets up a scroll event listener for the sticky-header + virtualization case.

Because afterUpdate fires after every reactive update, this cleanup/re-attach cycle runs even on unrelated state changes (row selection, sorting, etc.).

This is wasteful, as the listener only needs to change when virtualConfig, stickyHeader, tableRef, or calculatedContainerHeight change.


metonym:

PR #2823, opened 2026-04-04
perf(tree-view): remove unused tree walker reactive initialization

metonym:
Removes dead code: `treeWalker` is already initialized in on mount, and does not require re-creating.


metonym:

PR #2824, opened 2026-04-04
fix: prevent infinite loop when all items are disabled

metonym:
## Summary

- Fix edge case infinite `while` loop in keyboard navigation when all items are disabled in **ComboBox**, **Dropdown**, and **MultiSelect**
- Fix ComboBox `change()` using the unfiltered `items` array instead of the filtered `_items` array, causing wrong disabled-item checks during active filtering

## Strategy

Added an `attempts` counter to bound the `while (disabled)` loop to at most `N` iterations (where `N` is the item count). If every item is disabled, the loop exits cleanly and `highlightedIndex` is left unchanged rather than landing on a disabled item.

## Additional `ComboBox` fix

The `change()` function declared `_items` as the filtered list but then checked `items[index].disabled` — the unfiltered array. When filtering narrows the list, these arrays have different lengths and orderings, so the disabled check could read the wrong item (or even out-of-bounds). Fixed all references in the loop to use `_items` consistently.

### DIFF
--- a/src/MultiSelect/MultiSelect.svelte
+++ b/src/MultiSelect/MultiSelect.svelte
@@ -275,8 +275,9 @@
     }
 
     let disabled = itemsToUse[index].disabled;
+    let attempts = 0;
 
-    while (disabled) {
+    while (disabled && attempts < length) {
       index = index + direction;
 
       if (index < 0) {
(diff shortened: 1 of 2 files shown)


metonym:

PR #2826, opened 2026-04-04
perf(data-table): guard rows assignment with reference check

metonym:
Closes #2819, alternative to #2819

Use a cheap reference check for syncing the original rows from filtered rows.


metonym:

PR #2828, opened 2026-04-05
perf(multi-select): optimize `isSelectAll` logic

metonym:
Optimizes the new `isSelectAll` option mapping/filtering in `MultiSelect`.

- **Single-item toggle**: Replace two full `.map()` passes with direct index assignment
- **Select-all toggle**: Skip allocating new objects for items already in the target state (disabled or already checked/unchecked)


metonym:

PR #2829, opened 2026-04-05
feat(data-table): dispatch `sort` event

metonym:
`DataTable` now dispatches a **`sort`** event when a sortable column header would change the active sort.

The event is **cancelable** (same pattern as `ToastNotification` / `OverflowMenu`): callers can `preventDefault()` to skip internal `sortKey` / `sortDirection` updates and **skip client-side reordering** of `rows`, which is the usual setup for **server-driven** tables.

Additional TypeScript fix: row-specific header keys are tightened via **`DataTableKey<Row>`** so generic row types (e.g. `ProductRow`) get accurate paths for headers, cells, and **`sort` event `detail`**.

## Motivation

A dedicated sort event helps implement the following scenarios:

- **Server-side sorting**: Large datasets are sorted in the database or API. The client should send sort column and order, then replace `rows` with the response. `preventDefault()` avoids the table applying a second, possibly inconsistent client sort on the same payload.
- **URL / query sync**: Persist `sort` + `order` in the query string so views are shareable and restorable.
- **Analytics**: Observe which columns users sort and how often.

---

Screenshot 2026-04-04 at 9 58 42 PM

### DIFF
--- a/docs/src/pages/components/DataTable.svx
+++ b/docs/src/pages/components/DataTable.svx
@@ -893,6 +893,8 @@ Set `size="compact"` for minimal row height.
 
 Set `sortable` to `true` to enable column sorting. Disable sorting on specific columns by setting `sort: false` in the header object.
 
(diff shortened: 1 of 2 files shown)


metonym:

PR #2830, opened 2026-04-05
fix(multi-select): `sortItem` comparator returns a number, not `Item`

metonym:
The TypeScript type for `MultiSelect` sortItem is incorrect, in the case the user returns the comparator.

It should be a number, not the `Item`.


metonym:

PR #2831, opened 2026-04-05
fix(multi-select): invalid/warn, filterable variant should only render one icon

metonym:
Fixes a minor UI bug.

For the filterable `MultiSelect`, if using invalid/warn states, the same icon would be rendered twice (overlaying each other in the DOM).


metonym:

PR #2832, opened 2026-04-05
perf(multi-select): avoid double re-sort for `selectionFeedback="top"`

metonym:
When selectionFeedback === "top", every item toggle causes two full reactive cycles:

1. `selectItem()` toggles checked flags → derivatives recompute → afterUpdate writes selectedIds
2. The $: if (selectedIds ...) reactive block sees the new selectedIds, calls sort(), reassigns sortedItems → derivatives recompute again → afterUpdate fires again (no-op)

This fix is to guard the reactive block so it only calls sort() for external selectedIds changes. Have selectItem() call sort() directly when selectionFeedback === "top" to get proper ordering in a single cycle.


metonym:

PR #2833, opened 2026-04-05
perf(number-input): cache `Intl.NumberFormat` locale usage

metonym:
Currently, `Intl.NumberFormat` is called on every key stroke if a custom locale is used. This caches these calls.


metonym:

PR #2835, opened 2026-04-05
feat(data-table): add top-level `sort` prop

metonym:
Today, custom sorting means attaching a `sort` function to each header object. That is repetitive when many columns share the same rules (for example, one function that switches on `key` for dates, numbers, and strings). A table-level `sort` prop is a convenience: define the comparator once, and only override per column when a header supplies its own `sort` (which still takes precedence).

This adds an optional `sort` prop on `DataTable` for a default comparator across sortable columns. Per-column `sort` on headers wins when present.

The comparator is called as `(a, b, context)` with:

- `a` / `b`: resolved cell values for the active sort column (typed as `DataTableSortValue<Row>` when using a row generic).
- `context`: `{ key, ascending, row_a, row_b }` (column key, direction, and full rows) so logic can branch on the column or use other fields.
- If a header defines `sort` (and it is not `false`), that function is used (unchanged).
  - Otherwise, if `sort` is set on the table, it runs and its result is applied with the same ascending/descending rules as today (`ascending ? result : -result`).
  - Otherwise, sorting falls back to the built-in `compareValues` behavior.

```svelte
<DataTable
  sortable
  title="Load balancers"
  sort={(a, b, { key }) => {
    switch (key) {
      case "expireDate":
        return new Date(a) - new Date(b);
      case "port":
        return a - b;
      default:
        return String(a).localeCompare(String(b));
    }
  }}
  headers={[
    { key: "protocol", value: "Protocol" },
    {
      key: "expireDate",
      value: "Expire date",
      display: (date) => new Date(date).toLocaleDateString(),
    },
  ]}
  rows={[
    {
      id: "a",
      name: "Load Balancer 3",
      protocol: "HTTP",
      port: 3000,
      expireDate: "2020-10-21",
    }
  ]}
/>

```

### DIFF
--- a/docs/src/pages/components/DataTable.svx
+++ b/docs/src/pages/components/DataTable.svx
@@ -1115,6 +1115,12 @@ Use `display` and `sort` functions in header objects to customize cell rendering
   ]}"
 />
 
+## Sortable with top-level sort
+
+Use the `sort` prop to define a default comparator for all sortable columns. Per-header `sort` functions take precedence. The comparator receives `(a, b, { key, ascending, row_a, row_b })`, enabling key-aware and row-level sorting logic.
+
+<FileSource src="/framed/DataTable/DataTableSortProp" />
+
 ## Sortable with nested object values
 
 Access nested object properties using dot notation in the header key.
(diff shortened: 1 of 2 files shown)


metonym:

PR #2837, opened 2026-04-05
fix(multi-select): Backspace/Delete should clear selection

metonym:
Per the spec ([React implementation as reference](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx)), the Backspace/Delete keys should clear the selection for a filterable multi-select.

The nuance with Delete is that is should only clear the selection if the menu is closed and input focused (e.g., focus the input, press "Escape" to close the menu).


metonym:

PR #2838, opened 2026-04-05
fix(dropdown): do not spread `translateWithId` to raw `button` element

metonym:
`translateWithId` is already passed to `ListBoxMenuIcon`, and should not be spread to the raw `button` element in `Dropdown`.


metonym:

PR #2840, opened 2026-04-05
fix(dropdown): button uses `combobox` role, apply missing ARIA attributes

metonym:
Aligns with the [upstream implementation](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Dropdown/Dropdown.tsx), applying the correct semantics (button has `combobox` role) and missing ARIA attributes.

Also incorporate another fix, now that `Dropdown` may not have an initially selected item (see #2716), where the `labelText` value should be used as the `aria-label` fallback value.


metonym:

PR #2841, opened 2026-04-05
fix(number-input): text mode should also apply min/max/step attributes to `input`

metonym:
The text-mode input should also forward the min/max/step props to `input`. Currently, only `$$restProps` are passed through, which omits the exported min/max/step props.


metonym:

PR #2842, opened 2026-04-05
fix(number-input): remove `pattern` for number type input

metonym:
Removes dead code: `pattern` is not respected if `type="number"` ([docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/number#pattern_validation)).


metonym:

PR #2843, opened 2026-04-05
fix(tree-view): escape node ID when using `querySelector` for focusing

metonym:
`TreeView` uses the user-supplied node ID for focusing in `showNode`.

However, interpolating the id directly breaks when the id contains characters that are special in CSS selectors or that terminate the attribute value. The id is now passed through `CSS.escape()` so the selector is valid and matches the real DOM `id` attribute.

Examples:

- `node-with-"-char`
- `"leading-quote`
- `trailing-quote"`

### DIFF
--- a/src/TreeView/TreeView.svelte
+++ b/src/TreeView/TreeView.svelte
@@ -293,7 +293,7 @@
 
         if (focus) {
           tick().then(() => {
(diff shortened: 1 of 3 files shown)


metonym:

PR #2844, opened 2026-04-05
fix(list-box): add missing "lg" size mapping

metonym:
The lg/xl sizes are the same in Carbon v10.

This simply adds the missing "lg" mapping for DX. It updates the ComboBox size mappings to allow this size (for consistency with Dropdown/MultiSelect), even though docs only advertise default/small/xl sizes.


metonym:

PR #2845, opened 2026-04-05
fix(multi-select): fix typo in `size` prop description

metonym:


metonym:

PR #2846, opened 2026-04-05
fix(combo-box): do not trap focus when clicking outside element

metonym:
When the `ComboBox` menu is open, tabbing away from the combobox triggers ref.focus() for any relatedTarget that isn’t an input-like element. This means if a keyboard user tabs to a plain link or any other focusable element outside the combobox, focus remains trapped in the combo box input.

The current allowlist approach is too restrictive; instead, use a more general `relatedTarget` check.


metonym:

PR #2847, opened 2026-04-05
fix(combo-box): do not auto-select partial match when pressing "Enter"

metonym:
A bug occurs where, if using a filterable `ComboBox`, pressing Enter will select the first partial match.

This may need auditing first to align better with the upstream implementation. The upstream implementation has a different approach: for a partial match, it first focuses the first match in the menu, allowing the user to then press Enter to select it.


metonym:

PR #2848, opened 2026-04-05
fix(multi-select): narrow `itemToString` return type

metonym:
Functionally no change, just for DX/documentation: instead of an explicity `any`, narrow to string or `Item["id"]` (generic, but defaults to `any`).


metonym:

PR #2849, opened 2026-04-05
fix(tree-view): `expandNodes` should expand all by default

metonym:
> Filter function should return `true` for nodes to expand. If not provided, expands all nodes.

`expandNodes` supports expanding a subset of nodes. However, it should be the converse of `collapseNodes`, and expand all nodes if no filter is provided (collapseNodes collapses all if no arguments are provided).

Added regression tests, and also update the docs examples to showcase this behavior.


metonym:

PR #2850, opened 2026-04-05
feat(ui-shell): support classic theme

### LINKED ISSUE #2836: Mixed-Theme Support
Is there guidance on how to achieve the older visual look post-v0.105.0?

> Previously, the UI Shell only supported one theme: a mixed Gray 100 header and White side navigation.

I find the high contrast between the header and the rest of the content to be a nice approach. I'm happy to see the theme component being introduced.

Being able to bring in visual styles like what is displayed [here](https://carbondesignsystem.com/components/UI-shell-header/usage/#anatomy) would be really useful.

Thanks in advance!

metonym:
Closes #2836

Support the classic (mixed) UI Shell theme. To minimize the generated CSS bundle size, only support the classic theme via `all.css`, which uses CSS variables for dynamic theming. With this approach, theming for `Header` and `SideNav` are reused via CSS variables.

---

Screenshot 2026-04-05 at 12 37 29 PM


metonym:

PR #2867, opened 2026-04-11
fix(types): `tabindex` type allows number or string

metonym:
#1524

The official Svelte types have `tabindex` as `number | undefined`. Historically, this library has left the inferred type to be string. This isn't incorrect, but the type is too narrow. This prop should explicitly allow number or string (I myself gravitate towards using numbers, because of React).

### DIFF
--- a/src/Button/Button.svelte
+++ b/src/Button/Button.svelte
@@ -91,7 +91,10 @@
    */
   export let href = undefined;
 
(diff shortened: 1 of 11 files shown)


metonym:
Fixed in [v0.106.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.106.1)

PR #2870, opened 2026-04-16
fix(data-table-skeleton): do not require `DataTableHeader` key/empty/value properties

### LINKED ISSUE #2869: DataTableSkeletonProps is intersected with DataTableHeader, making `key`/`empty`/`value` required at the component level
## Version
`carbon-components-svelte@0.106.0`

## Problem

`DataTableSkeleton.svelte.d.ts` declares props as:

```ts
export type DataTableSkeletonProps = Omit<
  $RestProps,
  keyof ($Props & DataTableHeader)
> & $Props & DataTableHeader;
The trailing & DataTableHeader pulls properties of a single header row (key, empty, value) onto the component itself. Every valid usage now fails type checking:


<DataTableSkeleton size="tall" showHeader={false} showToolbar={false} rows={10} />

Type '{ size: "tall"; ... }' is missing the following properties
from type 'DataTableEmptyHeader<DataTableRow<any>>': key, empty
```

## Fix
Drop the & DataTableHeader from the type:

export type DataTableSkeletonProps = Omit<$RestProps, keyof $Props> & $Props;
Header rows are already covered by the headers prop.

metonym:
Fixes #2869

### DIFF
--- a/src/DataTable/DataTableSkeleton.svelte
+++ b/src/DataTable/DataTableSkeleton.svelte
@@ -1,6 +1,4 @@
 <script>
-  /** @extends {"./DataTable.svelte"} DataTableHeader */
-
   /**
    * Specify the number of columns.
    * Superseded by `headers` if `headers` is a non-empty array.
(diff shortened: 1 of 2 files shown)


metonym:
Fixed in [v0.106.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.106.1)

PR #2876, opened 2026-04-25
feat(floating-portal): auto-detect native `dialog` as target

### LINKED ISSUE #2873: OverflowMenu portal menu mounts outside <dialog> element and is blocked by backdrop
When an OverflowMenu has `portalMenu` set to true and is placed inside a native `<dialog>` element (shown with showModal()), the menu is inserted at the document body and ends up behind the dialog backdrop. The control opens but is visually hidden under the backdrop so it cannot be interacted with.

#### Why this happens:
showModal() creates a modal/backdrop that sits above document contents. The OverflowMenu portal currently mounts to the document root (e.g. document.body), so the portal node is below the dialog backdrop and not visible/interactive.

#### Expected behavior:
With portalMenu={true}, the OverflowMenu should be visible and accessible when used inside a `<dialog>`. Either:
The portal is mounted into the dialog itself (so it’s above the backdrop), or the portal node is mounted to a container that is above the backdrop.

metonym:
Fixes [#2873](https://github.com/carbon-design-system/carbon-components-svelte/issues/2873)

If portallable-components (like `OverflowMenu`) are used inside a [native dialog](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog), portalling doesn't work as expected, as the portal is rendered to `document.body`, not inside the `dialog` layer.

The solution is two-fold:

- Allow a custom `target` override for `Portal` and `FloatingPortal`
- Auto-detect `dialog` so that a component like `OverflowMenu` will automatically target the `dialog` if `portalMenu` is true.

**Repro**

```svelte
<script>
  import {
    Button,
    OverflowMenu,
    OverflowMenuItem,
  } from "carbon-components-svelte";

  let dialog = null;
</script>

<button
  type="button"
  on:click={() => dialog?.showModal()}
>
  Open dialog
</button>

<dialog bind:this={dialog}>
  <OverflowMenu portalMenu>
    <OverflowMenuItem text="Edit" />
    <OverflowMenuItem text="Duplicate" />
    <OverflowMenuItem danger text="Delete" />
  </OverflowMenu>
  <Button kind="secondary" on:click={() => dialog?.close()}>Close</Button>
</dialog>

```

Screenshot 2026-04-24 at 8 58 36 PM

### DIFF
--- a/docs/src/pages/components/FloatingPortal.svx
+++ b/docs/src/pages/components/FloatingPortal.svx
@@ -69,3 +69,7 @@ By default, the floating content is rendered below the anchor element. However, 
 Use the `direction` slot prop to read the actual rendered direction after auto-flipping for conditional logic, like styles.
 
 <FileSource src="/framed/FloatingPortal/FloatingPortalSlotDirection" />
+
+## Custom target
+
+Use the `target` prop to mount the floating content into a specific DOM element instead of `document.body`. Positioning is always driven by the `anchor`, so `target` only changes which DOM subtree the floating content lives in. See the [Portal custom target example](/components/Portal#custom-target) for typical use cases.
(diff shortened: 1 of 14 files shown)


metonym:
Released in [v0.107.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.107.0)

PR #2947, opened 2026-05-09
fix(button): prevent literal `true` from leaking into `class`

metonym:
Operator precedence (`&&` over `||`) made the `size === "field"` branch evaluate to `true`, which survived `.filter(Boolean)` and stringified into the rendered `class`.

`bx--btn--field` is added by the next entry, so drop the redundant half.


metonym:
Superseded by #2949

PR #2981, opened 2026-05-12
Adding read only to components that dont have it

kevinepena:


metonym:
Thank you so much for this. I've opened #3019 which breaks down this PR by commit. I've assigned co-authorship to you.

#3019 also adds unit tests, and incorporates a small Toggle readonly modification (cursor should not be pointer).

metonym:
I'll open a separate PR for the other misc fixes, like NumberInput/Select.

PR #2996, opened 2026-05-14
fix(notification): track reactive `kind` for default `statusIconDescription`

metonym:
The default `${kind} icon` was evaluated once at script init, so toggling `kind` after mount kept announcing the original value. Default the prop to `undefined` and fall back to a reactive derived label.

### DIFF
--- a/src/Notification/InlineNotification.svelte
+++ b/src/Notification/InlineNotification.svelte
@@ -32,7 +32,7 @@
    * Specify the ARIA label for the status icon.
    * @type {string}
    * */
-  export let statusIconDescription = `${kind} icon`;
+  export let statusIconDescription = undefined;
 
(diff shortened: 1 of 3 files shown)


brunnerh:
Personally, I don't think the icon should have any description. I always have to go out of my way to remove it manually. There is no value in it, decorative elements don't need accessible text.

If anything, screen readers could have invisibly prefixed the title/message with the `kind`, but no-one needs to hear about there being an "icon".

brunnerh:
Though, invisible _not localized_ text is a pitfall in general.
So maybe it should just be dropped entirely...

metonym:
Makes sense to me, thanks for the feedback!

metonym:
https://github.com/carbon-design-system/carbon-components-svelte/pull/3000

PR #3000, opened 2026-05-15
fix(notification)!: remove decorative `statusIconDescription` prop

metonym:
Closes #2996

The icon displayed in notification components should not have a dedicated description. Instead, it should be treated as a decorative element so screen readers don't read out the icon description.


metonym:
CC @brunnerh

PR #3105, opened 2026-05-29
feat(outbound-link): add `assistiveText` prop for screen readers

b-r-i-a-n-w-e-s-t:
OutboundLink opens in a new tab but only signaled this visually (via the aria-hidden launch icon). This adds visually-hidden text with a new assistiveText prop defaulting to "(opens in a new tab)" so screen readers announce the behavior. It can be customized or set to "" to opt out.

### DIFF
--- a/docs/src/pages/components/Link.svx
+++ b/docs/src/pages/components/Link.svx
@@ -27,9 +27,20 @@ For external links that open in a new tab, the component automatically adds secu
 
 ## Outbound link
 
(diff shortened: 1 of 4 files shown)


metonym (approved):
Seems reasonable to me. Thanks!

metonym:
Released in [v0.108.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.108.0)

PR #3161, opened 2026-06-06
fix(accordion-item): a11y parities

metonym:
Aligns with the upstream implementation:

- Remove `iconDescription` on decorative chevron icon
- Breaking: rename it to `ariaLabel` and provide it to the heading button (useful for a11y if there is no visible title)


edge500:
(ROM)

PR #3235, opened 2026-06-09
feat(css)!: drop older browser support (match Svelte 5 as new baseline)

metonym:
This is a breaking change that modernizes the pre-bundles CSS stylesheets (~7% size decrease).

Swaps PostCSS/autoprefixer for Lightning CSS in `build-css` and narrows vendor-prefix targets to Svelte 5's supported browsers. The old query (`last 1 version`, `ie >= 11`, `Firefox ESR`) was a holdover from Carbon's IE era; Svelte 5 doesn't run in IE anyway, and Lightning CSS was emitting excess `-ms-` prefixes trying to satisfy that mix.

This is primarily a build performance and output-size win, with browser targets that actually match the runtime.

**Browser targets**

The new floors match [Svelte 5 browser support](https://svelte.dev/docs/svelte/browser-support):

```ts
["Chrome >= 87", "Firefox >= 83", "Safari >= 14", "Edge >= 87"]
```

Those minimums date to late 2020 (Chrome/Edge 87 and Firefox 83 shipped November 2020; Safari 14 shipped September 2020). Browserslist resolves every version from each floor through current: today, that means Chrome/Edge 87–149, Firefox 83–151, Safari 14–26.4. No IE, no `last 1 version` mobile churn.

The old query resolved to a different shape: bleeding-edge mobile (`and_chr 148`, `samsung 30`), current desktop (`chrome 149`, `firefox 151`), and **IE 11**. Prefix output was tuned for that set, not for a Svelte 5 app.

**Output size** (`master` autoprefixer + IE targets → this branch)

| File | Before | After | Δ |
|------|--------|-------|---|
| `css/all.css` | 853 KB | 793 KB | −7.0% |
| `css/white.css` | 674 KB | 621 KB | −7.8% |
| `css/g10.css` | 674 KB | 621 KB | −7.8% |
| `css/g80.css` | 674 KB | 621 KB | −7.8% |
| `css/g90.css` | 674 KB | 621 KB | −7.8% |
| `css/g100.css` | 674 KB | 621 KB | −7.8% |
| **Total (6 files)** | **4.22 MB** | **3.90 MB** | **−7.6%** |

`white.css` prefix counts: `-ms-` 245 → 153, `-webkit-` 652 → 77.

**Build speed** (cold cache, `rm -rf .cache/build-css && bun build:css`)

| Step | `master` (sass + autoprefixer) | This branch (sass-embedded + Lightning CSS) |
|------|-------------------------------|---------------------------------------------|
| Post-Sass pass (per file, parallel) | ~50–84 ms | 9–23 ms |
| Total `[build-css]` | ~2.5 s | 2.21 s |
| Cached rebuild | n/a (no cache) | ~23 ms |


metonym:
@brunnerh Curious to hear your take on this. I think shrinking the compiled CSS stylesheets is a major win, especially given that Svelte 5 has more modern browser requirements. And at this point, Svelte 5 has been stable for nearly two years.

brunnerh:
This sounds reasonable, but I can't claim to have much experience in this area.

There still seems to be a non-trivial amount of [usage of prior Svelte versions](https://majors.nullvoxpopuli.com/q/h?dateCutoff=2025-06-10&packages=svelte&showTotal=true), but I doubt that many of those are on Svelte 3/4 due to browser compatibility. The main concern is usually Safari because of the insanity that is the coupling between the application version and the OS.

Would still go ahead with this. Users of the library could probably re-introduce backward compatibility in their own build process where necessary.

metonym:
> This sounds reasonable, but I can't claim to have much experience in this area.
> 
> There still seems to a non-trivial amount of [usage of prior Svelte versions](https://majors.nullvoxpopuli.com/q/h?dateCutoff=2025-06-10&packages=svelte&showTotal=true), but I doubt that many of those are on Svelte 3/4 due to browser compatibility. The main concern is usually Safari because of the insanity that is the coupling between the application version and the OS.
> 
> Would still go ahead with this. Users of the library could probably re-introduce backward compatibility in their own build process where necessary.

Thank you for your advice. That's a very useful link.

Agreed, I think the CSS modernization is worth it here. This library still supports Svelte 3/4 (hence the legacy test harnesses), but it's long overdue for me to focus on Svelte 5 and beyond.

PR #3262, opened 2026-06-10
fix(combo-box): remove invalid listbox role from the root element

b-r-i-a-n-w-e-s-t:
`ComboBox` was getting wrapped by a `role="listbox"` element which was getting flagged as invalid. Followed the pattern from `Dropdown` and `MultiSelect` to set the role on `Listbox` to `undefined`

Now the `role="listbox"` is only applied on a sibling of `role="combobox"`

Tried to update the tests as well.


metonym (approved):

metonym:
Fixed in [v0.109.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.109.0)

PR #3279, opened 2026-06-11
fix(modal): prevent Enter on Cancel from also firing the primary action

wickning1:
It actually fires both primary and secondary events. This is a critical issue as users think they are canceling but we go ahead and carry out the action. I included a test that fails without the fix.


metonym (approved):

metonym:
Nice catch, thank you.

metonym:
Fixed in [v0.109.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.109.0)

PR #3352, opened 2026-06-21
perf(data-table): use Set for O(1) lookups for expanded rows

GaboUCR:
Optimized datatable variable `expandedRowIdsSet` to use a Set instead of a plain object. Increasing performance from linear O(n) to constant O(1) lookup time. This change enhances readability and cleaness for the use case needed for this variable, since a set is more appropiate than a plain object.

Included a new E2E test to verify this new functionality.

## Verification

1. Expand all rows

image

2. Contract them

image

3. Expand a few rows

image

4. Contract them

image

### DIFF
--- a/e2e/data-table.test.ts
+++ b/e2e/data-table.test.ts
@@ -54,6 +54,22 @@ test.describe("DataTable", () => {
     );
   });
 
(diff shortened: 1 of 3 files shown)


metonym (approved):
Thanks!

metonym:

PR #3359, opened 2026-06-24
fix(multi-select): make read-only reviewable, drop disabled semantics

b-r-i-a-n-w-e-s-t:
**Make read-only MultiSelect reviewable instead of disabled**

Read-only MultiSelect can't behave like a disabled field: the menu won't open providing no way to see what was selected, and screen readers announced it as disabled/unavailable.

This edit makes read-only reviewable:

- The menu opens and you can navigate the options to review what's selected, but you can't change or
clear the selection.
- Dropped aria-disabled in favor of aria-readonly on the combobox and listbox (both valid per ARIA
1.2).
- All selection changes funnel through one guard in selectItem, so every path (click, Enter, label
click, select-all, clear) is blocked consistently in read-only.
- AT doesn't reliably surface aria-readonly, so the state is also exposed as visually-hidden text
referenced by the field. The number of selected items is folded into that text ("Read-only, 3
selected") since the count badge is hidden from assistive tech in read-only.
- The read-only text is overridable via a new readonlyText prop.

### DIFF
--- a/docs/src/pages/components/MultiSelect.svx
+++ b/docs/src/pages/components/MultiSelect.svx
@@ -434,7 +434,7 @@ Disable specific items using the disabled property in the items array. This exam
 
 ### Read-only
 
-Set `readonly` to `true` to display non-editable values. Selected items remain visible but cannot be cleared.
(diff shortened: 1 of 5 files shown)


metonym (approved):
Makes sense to me. Thanks for digging into this.

metonym:

PR #3362, opened 2026-06-26
fix: preserve displayed selection across async items changes

b-r-i-a-n-w-e-s-t:
I ran into an issue where `ComboBox`, `Select`, and `MultiSelect` would lose their **displayed** selection whenever `items` changes asynchronously, for example when an async fetch fills an empty list, or when a list gets cleared and then repopulated. The retained value/id stays correct the whole time, and it's only the rendered selection that blanks out. `Dropdown` isn't affected here, since it derives its display purely from the items.

## Fix

- **ComboBox**: stopped it from nulling a set `selectedId` while the matching item is simply unresolved (the async items just haven't loaded yet). It now also re-resolves `selectedItem` on every items change, instead of only when the id changes. 
- **Select**: re-assert the native `<select>` value after the DOM updates. The `<option selected>` attribute only maps to the live selection at insert or reset time, so any options added later leave the element sitting at `selectedIndex = -1`.
- **MultiSelect**: update the `prevChecked` snapshot whenever items changes, so an items-driven count change isn't mistaken for the user deselecting something, which was resetting `selectedIds` to `[]`.

### DIFF
--- a/e2e/fixtures/ItemsResyncFixture.svelte
+++ b/e2e/fixtures/ItemsResyncFixture.svelte
@@ -1,5 +1,11 @@
 <script>
-  import { ComboBox, Select, SelectItem } from "carbon-components-svelte";
+  import {
+    ComboBox,
+    Dropdown,
+    MultiSelect,
+    Select,
+    SelectItem,
+  } from "carbon-components-svelte";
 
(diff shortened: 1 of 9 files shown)


metonym (approved):
Makes sense, thank you.

Cleaned up code comments and split into individual fix commits.

PR #3392, opened 2026-06-29
fix(select): re-assert native value when an option unmounts

b-r-i-a-n-w-e-s-t:
_Removing_ an option from a native `<select>` can reset selectedIndex to -1, so SelectItem now re-asserts the native value on unmount the same way 9e60780 does on mount. `ComboBox` and `MultiSelect` track their selection in JS so they don't need this.


metonym (approved):
LGTM, thank you!

metonym:

PR #3400, opened 2026-07-03
fix(select)!: isolate chevron position from user-supplied margin

### LINKED ISSUE #3136: Margins on `Select` affect chevron
The margin on the `<select>` affects the layout within the container and thus makes the chevron misaligned.

[Playground](https://svelte.dev/playground/69f52a72757d4ff3954993a14570cc36?version=latest)

Image

<details>
<summary>Code</summary>

```svelte
<script>
	import { Header, Content, Select } from 'carbon-components-svelte@0.107.1';
</script>

<svelte:head><link rel="stylesheet" href="https://unpkg.com/carbon-components-svelte@0.107.1/css/g100.css" /></svelte:head>

<Header platformName="Carbon"/>

<Content>
	<Select>
		<option>A</option>
	</Select>
	<Select class="mb32">
		<option>B</option>
	</Select>
	<Select style="margin-bottom: 32px">
		<option>C</option>
	</Select>
</Content>

<style>
	:global(.mb32) {
		margin-bottom: 32px;
	}
</style>
```

</details>

Not sure how to this could/should be fixed. Remember there was some discussion about having different properties to forward attributes to specific internal elements, e.g. for layout like this something like `containerProps` could help, though not pretty.

User space workaround: Wrap `<Select>` in element and apply layout classes/styles there.

metonym:
Fixes #3136

- A margin applied to `<select>` (via `class` or inline `style`) grew the auto-height flex wrapper the chevron icon sizes against (`height: 100%`), shifting the chevron off the visible field.
- `$$restProps` now spreads onto the outer `.bx--form-item` wrapper instead of the `<select>` element, so consumer-supplied layout props no longer participate in the chevron's positioning context.
- `aria-label`/`aria-labelledby` are forwarded explicitly to the `<select>` element(s) so the `noLabel` + `aria-label` accessible-naming pattern keeps working now that `$$restProps` moved.

### Breaking change

`$$restProps` attributes other than `class`/`style` (for example `data-testid`, `autofocus`, `form`) now land on the wrapper `<div>` instead of `<select>`. There was no documented `@restProps` contract prior to this change; `@restProps {div}` is now documented.

### DIFF
--- a/src/Select/Select.svelte
+++ b/src/Select/Select.svelte
@@ -2,6 +2,7 @@
   /**
    * @template {string | number} [Value=string | number]
    * @event {Value} update The selected value.
+   * @restProps {div}
    */
 
   /**
@@ -191,7 +192,11 @@
   $: isFluid = !inline && (fluid || !!formContext?.isFluid);
 </script>
 
-<div class:bx--form-item={true} class:bx--select--fluid={isFluid}>
+<div
+  class:bx--form-item={true}
+  class:bx--select--fluid={isFluid}
+  {...$$restProps}
+>
   <div
     class:bx--select={true}
     class:bx--select--inline={inline}
@@ -229,6 +234,8 @@
                   : undefined}
             aria-invalid={showInvalid || undefined}
             aria-readonly={readonly || undefined}
+            aria-label={$$props["aria-label"]}
+            aria-labelledby={$$props["aria-labelledby"]}
             disabled={disabled || undefined}
             required={required || undefined}
             {id}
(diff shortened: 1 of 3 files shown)


brunnerh:
I don't think this is good.

For a `Select` component, people expect the `$$restProps` to be applied to the underlying `<select>`, not the wrapper. Generally, layout should be the _rare exception_ where you might want to interact with the wrapper at all.

Especially with Svelte 5, where all event handlers (`on...`) and [attachments](https://svelte.dev/docs/svelte/@attach) can automatically be forwarded via rest props, having them on the correct element is even more important.

metonym:
Thank you for the feedback, agreed. Was experimenting with this, on a whim. But it's a prevalent problem affecting many other components as well...

And this would be a painful breaking change.

More than ever, I want to:

- Get to a comfortable v1 baseline (Carbon React v11 + IBM Products parity)
- Rewrite to Runes

PR #3437, opened 2026-07-17
fix(select): keep the placeholder when options change with nothing selected

b-r-i-a-n-w-e-s-t:
A `Select` with **no selection** loses its placeholder when the option list changes on the Svelte 5 runtime. Render a `Select` whose first option is an empty-valued placeholder, select nothing, then mount/unmount any option: the input goes blank and `select.selectedIndex` sticks at `-1`. 

Svelte 4 consumers are unaffected.

Also filed an issue with [svelte](https://github.com/sveltejs/svelte/issues/18557) as this isn't just a carbon issue.

### DIFF
--- a/e2e/fixtures/ItemsResyncFixture.svelte
+++ b/e2e/fixtures/ItemsResyncFixture.svelte
@@ -58,6 +58,8 @@
   let selRemoveItems = make();
   let selRemoveSelected = "orwell";
 
+  let selNoSelectionExtra = false;
+
   let msSwapItems = make();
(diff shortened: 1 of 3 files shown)


metonym (approved):
Nice find and fix. Agree with the e2e test only b/c of the mutation observer dependency.

metonym:
Fixed in [v0.110.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.110.1)

PR #3438, opened 2026-07-17
fix(multi-select): prevent page scroll on Space in filterable read-only mode

b-r-i-a-n-w-e-s-t:
When MultiSelect is in `filterable` mode and is also set to `readonly`, if someone uses SPACE to open the menu, the event falls through and scrolls the page.

This adds a readonly check and corresponding `preventDefault`


metonym:

PR #3445, opened 2026-07-29
fix(date-picker): keep calendar portalled after reactive re-run

### LINKED ISSUE #3444: DatePicker: portalMenu reverted to static:true on reactive re-run inside Modal
When DatePicker is inside a ComposedModal (or any container with overflow: hidden) and portalMenu is true, the calendar is correctly appended to body on first creation. However, on any subsequent re-evaluation of the reactive statement calling initCalendar, the flatpickrProps loop applies calendar.set("static", true) (from the default { static: true }), which moves the calendar back inside the date picker container where it gets clipped.
 
 Root cause: In initCalendar, when calendar already exists:
 
  ```js
    for (const [option, value] of Object.entries(flatpickrProps)) {
      applyOptionIfChanged(option, value);
    }
  ```
 
  This applies static: true from the default flatpickrProps, overriding the static: false that was set during
  initial creation when portalMenu was true.
 
  Workaround: Explicitly set flatpickrProps={{ static: false }} alongside portalMenu={true}.
 
  Expected fix: The flatpickrProps loop should skip static when effectivePortalMenu is true, or the initial
  static: false should be tracked in lastAppliedOptions so it isn't overwritten.

metonym:
Fixes #3444 

The initCalendar reactive block reapplied flatpickrProps' default
static: true on every re-run, clobbering the static: false that
`portalMenu` sets at creation, so the calendar drifted out of the
portal (and could get clipped inside a Modal) after any unrelated
prop update.

### DIFF
--- a/src/DatePicker/DatePicker.svelte
+++ b/src/DatePicker/DatePicker.svelte
@@ -420,6 +420,10 @@
       applyOptionIfChanged("locale", locale, resolveLocale(locale));
       applyOptionIfChanged("dateFormat", dateFormat);
       for (const [option, value] of Object.entries(flatpickrProps)) {
+        // `static` is decided by `effectivePortalMenu` at creation time
+        // (see below); re-applying the default `flatpickrProps.static`
+        // here would clobber that on every reactive re-run.
+        if (option === "static" && effectivePortalMenu) continue;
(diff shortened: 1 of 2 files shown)


metonym:

PR #3447, opened 2026-07-31
fix(date-picker): resync flatpickr `allowInput` on readonly toggle

metonym:
`allowInput` was only ever set from `!readonlyAny` at flatpickr
creation time, unlike the sibling `clickOpens` option, which is
kept in sync on every readonly change. Toggling `readonly` after
the calendar mounted left `allowInput` stale, contradicting the
option's own stated purpose.

The reactive block that already resyncs `clickOpens` now also
resyncs `allowInput`. Added a regression test and a docs example
showing readonly toggled at runtime.


metonym:

PR #3540, opened 2026-08-04
fix(tag)!: remove forwarded `on:click` event from non-interactive tags

anishrajpandey:
Currently, screen readers like NVDA announces non-clickable tags as "Clickable" because of click event present in the else block. Removing the event listener should fix it for ADA compliance.


metonym (approved):
Nice find. This fix makes sense, especially since there is a dedicated interactive `Tag` variant. We'll want to audit the library more broadly for similar violations, e.g., the component skeletons that have click events.

metonym:

PR #3546, opened 2026-08-05
fix(multi-select): keep disabled options in keyboard navigation

wickning1:
Arrow keys no longer skip disabled options. Disabled items stay in the navigation sequence with aria-disabled="true" and refuse selection, per APG guidance for composite widgets — otherwise a screen reader user driving the listbox through aria-activedescendant cannot discover that a disabled option exists. Hover highlighting of disabled options remains off.

Options disabled by maxSelectedItems now reference a visually-hidden "Maximum items selected" description via aria-describedby, so the reason an option refuses selection is announced, not just the dimmed state. The text is exposed as a new maxSelectedText prop for i18n.

Note that this is a departure from the React project. They skip disabled items. But that behavior is not correct according to https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols.

### DIFF
--- a/src/MultiSelect/MultiSelect.svelte
+++ b/src/MultiSelect/MultiSelect.svelte
@@ -191,6 +191,13 @@
   export let readonlyText = "Read-only";
 
   /**
+   * Specify the assistive text describing options disabled by `maxSelectedItems`.
+   * Announced so screen reader users learn why an option refuses selection,
+   * not just that it is disabled.
+   */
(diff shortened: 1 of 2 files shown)


metonym (approved):
Thanks, agreed. `ComboBox` and `Dropdown` should also mirror this behavior, right?

wickning1:
Yes, they're a little harder to change so I'll PR them separately.

PR #3547, opened 2026-08-06
fix(dropdown,combobox): keep disabled options in keyboard navigation

wickning1:
Arrow keys now step through disabled options instead of skipping them, so screen reader users driving the menu via aria-activedescendant can discover they exist (APG "Focusability of disabled controls"). Enter on a highlighted disabled option matches click: no selection, and the menu stays open. Typeahead, first-match auto-highlight, and hover highlighting still target enabled items only.


metonym:

PR #3548, opened 2026-08-06
fix(multi-select)!: name the non-filterable trigger by its field label

wickning1:
The trigger's aria-label ("Open menu"/"Close menu") overrode the `<label for>` association, so screen readers never heard the field's label or selection count. The open/close text stays on the chevron icon, the selection count is now a visually-hidden description, and ListBoxField's unused translateWithId prop is removed.


metonym (approved):
Appreciate all of these a11y fixes 🙏

metonym:

PR #3590, opened 2026-08-07
fix(multi-select): make Space toggle the highlighted option, add Home/End

wickning1:
In the non-filterable variant, Space while the menu is open closed the menu instead of toggling the highlighted option which goes against the APG multi-select pattern. Space now toggles the highlighted option and still opens a closed menu.

Home/End now jump the highlight to the first/last option while the menu is open. However, in the filterable input they are ignored so the native text-caret behavior is preserved.


metonym (approved):
Nice work, thank you.

metonym:

PR #3640, opened 2026-08-11
fix(multi-select): resolve dangling `aria-describedby` for invalid and warn text

wickning1:
A non-fluid MultiSelect with invalid/invalidText set points the field's aria-describedby at an id that is never rendered: the id'd error element only exists in the fluid branch, while non-fluid error text comes from ListBox with no id. Screen readers silently drop dangling describedby references, so the error message is never announced to the users who most need it. Same for warnText.

Give ListBox's form-requirement divs optional invalidId/warnId props and wire them up from MultiSelect so the existing describedby actually resolves. The props default to undefined, so Dropdown, ComboBox, and other ListBox consumers are unchanged and do not appear to need this.

### DIFF
--- a/src/ListBox/ListBox.svelte
+++ b/src/ListBox/ListBox.svelte
@@ -26,11 +26,23 @@
   /** Specify the invalid state text */
   export let invalidText = "";
 
+  /**
+   * Specify an id for the invalid state text
+   * @type {string}
+   */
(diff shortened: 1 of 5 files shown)


metonym (approved):

metonym:

PR #3641, opened 2026-08-11
fix(multi-select): prevent page scroll on arrow keys in filterable read-only mode

wickning1:


metonym (approved):

metonym:

PR #3642, opened 2026-08-11
fix(multi-select): add keyboard clearing to the non-filterable variant, announce cleared selection

wickning1:
Delete/Backspace now clear the selection in the non-filterable field (matching the filterable variant and Carbon React), the shortcut is advertised in the field's visually-hidden description, and clearing by keyboard or the clear button announces "All items cleared" through a role="status" live region. Adds `clearSelectionText` and `selectionClearedText` props for i18n.


metonym (approved):

metonym:

PR #3644, opened 2026-08-12
fix(date-picker): use layout viewport width for portalMenu right-align, not window.innerWidth

b-r-i-a-n-w-e-s-t:
Fixes a portalMenu calendar issue I was running into when the page has a persistent scrollbar or `scrollbar-gutter: stable`.

positionFlatpickrCalendarFixed() decided whether to right-align the calendar by comparing against window.innerWidth, but the input's bounds come from getBoundingClientRect(), which is relative to the layout viewport and excludes any reserved scrollbar gutter. When a gutter is present the two disagree by its width (~15px), so the rightMost check could come out false when it should be true and the calendar would left-align and overhang past the container's real right edge instead of tucking in against the input.

Swapped window.innerWidth for document.documentElement.clientWidth.


metonym (approved):

metonym:

PR #3645, opened 2026-08-14
feat(multi-select): announce filter result counts in the filterable variant

wickning1:
When typing in the filterable MultiSelect, the option list narrows silently — a screen reader user can't tell whether their query matched 40 items, 5, or none without arrowing into the list. Announce the count ("5 results available", "No results") through the status live region, debounced 800ms so it doesn't chatter on every keystroke, skipped when the count hasn't changed, and cleared when the menu closes. The message is customizable via the new `filterResultsText` prop for i18n.


metonym (approved):

metonym:

PR #3647, opened 2026-08-14
fix(multi-select): stop the option-click focus bounce, drop aria-owns, set aria list attributes in virtual

wickning1:
These are the last of the fixes I identified in my accessibility review of the multi-select. Thanks for all the merges!

Clicking an option let focus land on the option and then bounced it back to the field with a programmatic focus() call, making many screen readers re-announce the entire field on every click. Prevent default on the option's mousedown instead so focus never leaves the trigger.

Also remove aria-owns from both field variants: ARIA 1.2 comboboxes reference the popup via aria-controls only, and aria-owns can cause double-reading.

Finally, set aria-posinset and aria-setsize in the virtualized menu to ensure correct counts for screen readers.

### DIFF
--- a/src/ListBox/ListBoxField.svelte
+++ b/src/ListBox/ListBoxField.svelte
@@ -42,7 +42,6 @@
   {id}
   {role}
   aria-expanded={ariaExpanded}
-  aria-owns={(ariaExpanded && menuId) || undefined}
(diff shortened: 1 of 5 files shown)


metonym (approved):
Thanks for these fixes!

PR #3699, opened 2026-08-19
fix(text-input): position the fluid maxCount counter opposite the label

b-r-i-a-n-w-e-s-t:
`.bx--text-input--fluid .bx--label` positions every label in the wrapper absolutely at `inset-inline-start: $spacing-05`. The counter is a second `.bx--label` in that wrapper, so it lands on the label's coordinates. Fluid TextArea already splits the pair apart; TextInput has no equivalent.

Adds a higher-specificity rule pinning the counter to the trailing edge.

### DIFF
--- a/css/_fluid-text-input.scss
+++ b/css/_fluid-text-input.scss
@@ -33,9 +33,15 @@
     inset-inline-start: $spacing-05;
   }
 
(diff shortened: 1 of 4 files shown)


metonym (approved):

metonym:
Fixed in [v0.111.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.111.1)

PR #3700, opened 2026-08-19
fix(multi-select): forward rest props to the non-filterable trigger

b-r-i-a-n-w-e-s-t:
The filterable variant spreads `$$restProps` onto its `<input>`, but the non-filterable trigger dropped them entirely, so `aria-invalid`, `aria-required`, `data-*`, etc had nowhere to land on the default MultiSelect.


metonym (approved):

metonym:
Fixed in [v0.111.1](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.111.1)

PR #3737, opened 2026-08-27
feat(file-uploader-button): parity for size and duplicate rejection

b-r-i-a-n-w-e-s-t:
`FileUploaderButton` is the only file input that missed the 0.111.0 size and duplicate work. `maxFileSize` was not a prop at all, so setting it was silently accepted as an unknown attribute, and `preventDuplicate`'s skips were unreported since there was no rejected event.

Routes the change handler through the existing `filterIncomingFiles` util that FileUploader and `FileUploaderDropContainer` already use, so the button reports both reasons through the same rejected payload shape.

### DIFF
--- a/docs/src/pages/components/FileUploader.svx
+++ b/docs/src/pages/components/FileUploader.svx
@@ -24,12 +24,25 @@ Enable multiple file selection by setting `multiple` to `true`.
 
 ### Prevent duplicate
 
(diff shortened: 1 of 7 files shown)


metonym (approved):
Nice catch, thanks so much!

metonym:
Released in [v0.112.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.112.0)

PR #3738, opened 2026-08-27
feat(list-box): add wrapOptions for WCAG 1.4.10 Reflow (Dropdown, ComboBox, MultiSelect)

### LINKED ISSUE #3709: [a11y]: ListBox menu options (ComboBox / Dropdown / MultiSelect) truncate at 400% zoom, WCAG 1.4.10 Reflow
@metonym I'm wondering if you have any thoughts on doing something to handle this here since it doesn't seem to have gotten much traction from the carbon team?: https://github.com/carbon-design-system/carbon/issues/22462

`.bx--list-box__menu-item__option` is fixed at 2.5rem with `white-space: nowrap` / `text-overflow: ellipsis`. At 1280×1024 and 400% zoom, any label longer than the (now very narrow) field is cut off. 

Wrapping with css when they aren't virtualized isn't a problem, but once the virtualizer kicks in with lists over 100 it assumes a fixed `itemHeight` (`offsetY = startIndex * itemHeight, totalHeight = length * itemHeight`).

Since wrapping deviates from Carbon's design, I'm thinking an opt-in prop on Dropdown / ComboBox / MultiSelect (e.g. wrapItems) that:

1. Applies wrap styles to the option (`height: auto; white-space: normal; overflow-wrap: anywhere`), including the checkbox label in MultiSelect.
2. Switches virtualization to measured row heights: something like a per-index height cache with a running-average estimate for unmeasured rows, prefix sums for offsets, binary search in getVisibleRange, and scrollTop correction when a row above the viewport is measured. The existing fixed-height path stays the default, so nothing changes for current consumers.

I could work on a PR if that approach sounds workable or maybe you've got other ideas?

b-r-i-a-n-w-e-s-t:
Closes #3709

Option labels are one fixed-height line with `text-overflow: ellipsis`.  A stylesheet override doesn't work either: any list past the virtualization threshold derives every offset from a single `itemHeight`.

## The prop

```svelte
<Dropdown wrapOptions {items} />
```

On `Dropdown`, `ComboBox`, `MultiSelect`, and `ListBoxMenu`. Labels break between words. That's `overflow-wrap: break-word` rather than `anywhere`.

Nothing else moves: accessible names, `aria-setsize`/`aria-posinset`, the `scroll` payload, and `itemHeight`, which becomes the estimate for rows not yet measured.

## Measured virtualization

Wrapped rows aren't equal height, so offsets have to come from measurement: a per-option height cache. `wrapOptions` is the only thing that turns it on.

## Default

Off. Turning it on shifts rendered heights for every existing consumer and puts every long list on the measured path. The downside is real: people stay non-conformant until they find the prop, but perhaps CDS will revisit the design as a whole at some point. 

Broke it into four commits to try and make it a bit more digestible, but it's larger than I had hoped. Hopefully it's useful in some way or another.

### DIFF
--- a/src/ComboBox/ComboBox.svelte
+++ b/src/ComboBox/ComboBox.svelte
@@ -177,7 +177,10 @@
    * - `"hide"`: keep all options mounted and hide non-matching ones with the
    *   `hidden` attribute.
    *
-   * `"hide"` falls back to `"remove"` when virtualization is enabled.
+   * `"hide"` falls back to `"remove"` when virtualization is enabled, which
+   * also keeps `wrapOptions` correct on a windowed list: a hidden option has no
+   * rendered height, so measuring one would put a zero into offsets it does not
+   * occupy.
    * @type {"remove" | "hide"}
    */
   export let filterMode = "remove";
@@ -228,7 +231,7 @@
    * Set `virtualize={true}` to explicitly enable virtualization with default settings.
    *
    * Provide an object to customize virtualization behavior:
-   * - `itemHeight` (default: size-based, or 64px for fluid unless `condensed`): Height of each item in pixels. Override when custom slots change row height.
+   * - `itemHeight` (default: size-based, or 64px for fluid unless `condensed`): Height of each item in pixels. Override when custom slots change row height. Under `wrapOptions`, heights are measured from the rendered options and this serves as the starting estimate for ones not yet measured.
    * - `containerHeight` (default: 300): The maximum height in pixels of the dropdown container.
    * - `overscan` (default: 3): The number of extra items to render above and below the viewport for smoother scrolling. Higher values may cause more flickering during very fast scrolling.
    * - `threshold` (default: 100): The minimum number of items required before virtualization activates. Lists with fewer items will render all items normally without virtualization.
@@ -238,6 +241,13 @@
   export let virtualize = undefined;
 
   /**
+   * Set to `true` to let an option's label wrap onto as many lines as it needs
+   * instead of being truncated with an ellipsis.
+   * @type {boolean}
+   */
+  export let wrapOptions = false;
+
+  /**
    * Set to `true` to render the dropdown menu in a portal,
    * allowing it to escape containers with `overflow: hidden`.
    * When inside a Modal, defaults to `true` unless explicitly set to `false`.
@@ -245,7 +255,13 @@
    */
   export let portalMenu = undefined;
 
-  import { afterUpdate, createEventDispatcher, getContext, tick } from "svelte";
+  import {
+    afterUpdate,
+    createEventDispatcher,
+    getContext,
+    onMount,
+    tick,
+  } from "svelte";
   import Checkmark from "../icons/Checkmark.svelte";
   import WarningAltFilled from "../icons/WarningAltFilled.svelte";
   import WarningFilled from "../icons/WarningFilled.svelte";
@@ -255,21 +271,14 @@
   import ListBoxMenuIcon from "../ListBox/ListBoxMenuIcon.svelte";
   import ListBoxMenuItem from "../ListBox/ListBoxMenuItem.svelte";
   import ListBoxSelection from "../ListBox/ListBoxSelection.svelte";
-  import {
-    getMenuItemHeight,
-    getMenuMaxHeight,
-  } from "../ListBox/list-box-utils.js";
+  import { shouldVirtualizeMenu } from "../ListBox/list-box-utils.js";
+  import { createMenuWindow } from "../ListBox/menuWindow.js";
   import { dismiss } from "../utils/dismiss.js";
   import { isOutsideClick } from "../utils/isOutsideClick.js";
   import { createScrollEndTracker } from "../utils/isScrollNearEnd.js";
   import { moveIndex } from "../utils/moveIndex.js";
   import { uniqueId } from "../utils/uniqueId.js";
-  import {
-    resetVirtualScrollOnClose,
-    scrollHighlightedIntoView,
-    scrollSelectedIntoView,
-    virtualListState,
-  } from "../utils/virtualize.js";
+  import { resetVirtualScrollOnClose } from "../utils/virtualize.js";
 
   const dispatch = createEventDispatcher();
   const scrollEndTracker = createScrollEndTracker();
@@ -292,6 +301,25 @@
 
   /** @type {null | HTMLDivElement} */
   let fieldRef = null;
+
+  /** @type {import("../ListBox/menuWindow.js").MenuWindowState} */
+  let menuState;
+
+  const menuWindow = createMenuWindow({
+    getContainer: () => listRef,
+    onScrollTop: (scrollTop) => {
+      listScrollTop = scrollTop;
+    },
+    onState: (state) => {
+      menuState = state;
+    },
+  });
+
+  onMount(() => {
+    return () => {
+      menuWindow.destroy();
+    };
+  });
 
   /**
    * @returns {boolean}
@@ -319,6 +347,7 @@
   function handleMenuScroll(event) {
     const target = /** @type {HTMLElement} */ (event.target);
     listScrollTop = target.scrollTop;
+    menuWindow.noteScroll(target.scrollTop);
     const detail = scrollEndTracker.observe({
       scrollTop: target.scrollTop,
       scrollHeight: target.scrollHeight,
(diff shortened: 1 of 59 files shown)


src/ListBox/menuWindow.js
```diff
@@ -0,0 +1,507 @@
...
+      }
+    }
+
+    const option = container.querySelector(
+      `[${VIRTUAL_INDEX_ATTRIBUTE}="${pending.index}"]`,
+    );
+    if (!(option instanceof HTMLElement)) return;
+
+    scrollIntoViewWithinMenu(option);
+    scrollTo(container.scrollTop);
```
metonym, line 295: `settle()` never clears `pending` after satisfying it, unlike every other exit path (`reset()`, `destroy()`, the `index < 0` branch in `requestMeasured`, the DOM fallback in `scrollIntoView`). A later, unrelated ResizeObserver batch (font load, lazy image, zoom) can re-enter `handleMeasured` -> `settle()` and re-run `scrollIntoViewWithinMenu` against this stale index, snapping the menu back to an old highlight. Should this clear `pending` here?

src/ComboBox/ComboBox.svelte
```diff
@@ -395,27 +424,14 @@
...
-            overscan: virtualConfig.overscan ?? 3,
-            maxItems: virtualConfig.maxItems,
-          });
-          if (nextScrollTop !== null) {
-            listScrollTop = nextScrollTop;
-            listRef.scrollTop = nextScrollTop;
-          }
-        }
+        if (!listRef || highlightedIndex < 0) return;
+        if (highlightOrigin === "pointer") return;
```
metonym: This guard sits inside the block gated only by `shouldVirtualize`, not `wrapOptions`, so it changes hover-scroll behavior on the existing virtualized path too, not just the new measured path. Is that intentional? If so, worth noting in the PR description since it currently reads as unchanged. If not, should this be scoped to `wrapOptions` only?

src/Dropdown/Dropdown.svelte
```diff
@@ -377,27 +392,14 @@
...
-            overscan: virtualConfig.overscan ?? 3,
-            maxItems: virtualConfig.maxItems,
-          });
-          if (nextScrollTop !== null) {
-            listScrollTop = nextScrollTop;
-            listRef.scrollTop = nextScrollTop;
-          }
-        }
+        if (!listRef || highlightedIndex < 0) return;
+        if (highlightOrigin === "pointer") return;
```
metonym: This guard sits inside the block gated only by `shouldVirtualize`, not `wrapOptions`, so it changes hover-scroll behavior on the existing virtualized path too, not just the new measured path. Is that intentional? If so, worth noting in the PR description since it currently reads as unchanged. If not, should this be scoped to `wrapOptions` only?

src/MultiSelect/MultiSelect.svelte
```diff
@@ -613,27 +630,14 @@
...
-            overscan: virtualConfig.overscan ?? 3,
-            maxItems: virtualConfig.maxItems,
-          });
-          if (nextScrollTop !== null) {
-            listScrollTop = nextScrollTop;
-            listRef.scrollTop = nextScrollTop;
-          }
-        }
+        if (!listRef || highlightedIndex < 0) return;
+        if (highlightOrigin === "pointer") return;
```
metonym: This guard sits inside the block gated only by `shouldVirtualize`, not `wrapOptions`, so it changes hover-scroll behavior on the existing virtualized path too, not just the new measured path. Is that intentional? If so, worth noting in the PR description since it currently reads as unchanged. If not, should this be scoped to `wrapOptions` only?

src/ComboBox/ComboBox.svelte
```diff
@@ -541,20 +549,25 @@
       ? undefined
       : `${id}-${filteredItems[highlightedIndex].id}`;
 
-  $: menuMaxHeight = getMenuMaxHeight(size);
-
-  $: virtualState = virtualListState({
+  $: menuState = menuWindow.update({
```
metonym, line 559: MultiSelect passes `getKey: (item) => item.id` to `menuWindow.update()` (line 873), but this doesn't. Without it, change detection falls back to comparing `items` by reference, so a consumer that rebuilds `items` reactively will look changed on every scroll tick and keep wiping the measured-height cache. Should this pass `getKey` too?

src/Dropdown/Dropdown.svelte
```diff
@@ -346,29 +359,31 @@
...
-
-  $: menuMaxHeight = getMenuMaxHeight(size);
+  $: shouldVirtualize = shouldVirtualizeMenu({ items, virtualize });
 
   // Fluid (non-condensed) menu items are 64px tall (see css/_fluid-list-box.scss).
   // Portaled menus render outside the fluid wrapper, so they keep default heights.
   $: hasFluidMenuItems = isFluid && !condensed && !effectivePortalMenu;
 
-  $: virtualState = virtualListState({
+  $: menuState = menuWindow.update({
```
metonym, line 368: MultiSelect passes `getKey: (item) => item.id` to `menuWindow.update()` (line 873), but this doesn't. Without it, change detection falls back to comparing `items` by reference, so a consumer that rebuilds `items` reactively will look changed on every scroll tick and keep wiping the measured-height cache. Should this pass `getKey` too?

src/ListBox/list-box-utils.js
```diff
@@ -49,3 +49,31 @@ export function getMenuItemHeight(size = "md", { fluid = false } = {}) {
   if (fluid) return FLUID_MENU_ITEM_HEIGHT;
   return MENU_ITEM_HEIGHT[size] ?? MENU_ITEM_HEIGHT.md;
 }
+
+/**
+ * Item count past which a menu windows itself without being asked. Matches
+ * `virtualize.js`'s own `threshold` default.
+ */
+const VIRTUALIZE_ITEM_COUNT_THRESHOLD = 100;
```
metonym: This duplicates virtualize.js's own default `threshold`, kept in sync only by the comment above. Could this import the actual default instead, so the two can't drift apart?

src/ListBox/menuWindow.js
b-r-i-a-n-w-e-s-t: looks like it should. I'll fix.

src/ComboBox/ComboBox.svelte
b-r-i-a-n-w-e-s-t: probably cleaner to scope this. I'm thinking handle it with `isMeasured` so the reason is visible. That path checks whether the option is fully visible, not just whether it's rendered, so hovering one that's clipped by the menu's bottom edge scrolls the list under the cursor.

```diff
       tick().then(() => {
         if (!listRef || highlightedIndex < 0) return;
-        if (highlightOrigin === "pointer") return;
+        if (isMeasured && highlightOrigin === "pointer") return;
         menuWindow.scrollIntoView(highlightedIndex, "nearest");
       });
```

src/Dropdown/Dropdown.svelte
b-r-i-a-n-w-e-s-t: ditto combobox

src/MultiSelect/MultiSelect.svelte
b-r-i-a-n-w-e-s-t: ditto combobox

src/ComboBox/ComboBox.svelte
b-r-i-a-n-w-e-s-t: yes, I'll add this to bring it inline with MultiSelect

src/Dropdown/Dropdown.svelte
b-r-i-a-n-w-e-s-t: will add here as well.

src/ListBox/list-box-utils.js
b-r-i-a-n-w-e-s-t: that makes sense to me.

src/ListBox/menuWindow.js
b-r-i-a-n-w-e-s-t: Tried that first, and it broke three tests that predate this branch. A measured scroll request needs to survive its first placement, because later measurements change row heights and can clip the still-highlighted option again. So it cancels instead when the reader takes over: on a scroll, or on the pointer landing on another option, which means later measurements cannot return the menu to an option the reader has left.

src/ComboBox/ComboBox.svelte
b-r-i-a-n-w-e-s-t: Shipped in 372d4d18, with one addition to what I posted: the guard also withdraws the outstanding request, since the pointer moving the highlight on ends what the highlight it replaced asked for.

b-r-i-a-n-w-e-s-t:
@metonym refactored a bit here to address the initial questions.

metonym (approved):
Nice work, thank you so much!

metonym:
Released in [v0.112.0](https://github.com/carbon-design-system/carbon-components-svelte/releases/tag/v0.112.0)

PR #3820, opened 2026-09-15
fix(data-table): keep table layout for sticky headers

metonym:
Four independent DataTable fixes bundled together: sticky header
layout, header width sizing, radio selection clearing, and expanded
row styling.

| Change | Fix |
| --- | --- |
| Sticky header | `stickyHeader` switched the table to `display: flex`, desyncing header/body columns and ignoring `width`/`minWidth`. Headers now stay `display: table` and pin with `position: sticky` instead. |
| Header width | Setting `width`/`minWidth` on one header forced `table-layout: fixed` on the whole table, so every other column lost content sizing. Widths now apply through a `colgroup` under auto layout. |
| Radio deselect | Radio selection had no way to clear itself from a click; only writing to `selectedRowIds` worked. |
| Expanded row content | Content in the `expandedRow` slot rendered flush against the parent row with the same background, so a nested table read as more parent rows. |

Radio selection can now clear on click:

```svelte
<DataTable radio allowDeselect bind:selectedRowIds {headers} {rows} />
```

Screenshot 2026-09-15 at 6 04 15 AM

---

A sized column no longer drags every other column into fixed layout.
Set `fixedLayout` to opt back into the old strict-sizing behavior:

```svelte
<DataTable
  fixedLayout
  headers={[{ key: "name", value: "Name", width: "200px" }, ...]}
  {rows}
/>
```


metonym:
Split into standalone PRs against master:

- https://github.com/carbon-design-system/carbon-components-svelte/pull/3824
- https://github.com/carbon-design-system/carbon-components-svelte/pull/3825
- https://github.com/carbon-design-system/carbon-components-svelte/pull/3826
- https://github.com/carbon-design-system/carbon-components-svelte/pull/3827

metonym:
Superseded by the four standalone PRs above.

